PR #26, opened 2013-04-12
added handling of image messages with opencv encodings to calibrator.py

stwirth:


vrabaud:
OpenCV encodings should not be handled as  they do not have color encodings in them: is C1 grayscale ? Or the V in HSV ? or Y in YUV ? Nobody knows. I'd rather not include  that patch. Plus you're doing a scaling which seems arbitrary. What is your  use case ?

stwirth:
My intention was to make the calibrator behave like image_view. I can display depth images with image_view but cannot run them through the calibrator. The scaling stretches all values to be between 0 and 255.
I agree that HSV and YUV have channels that are difficult to visualize but I think as long as the input image has only one channel ( I should check for "C1" instead of just "C") the calibrator could work with it.

ahendrix:
Calibrating a depth image with the checkerboard calibrator and a flat checkboard doesn't make much sense.

stwirth:
@ahendrix Who says that I am trying to do that?
The proper usage of the proposed feature should be left to the end user. I changed the scaling to be similar to the scaling in image_view.
Feel free to close the pull request if you think it's not useful for the community.

vrabaud:
ok, I'm just trying to understand your use case: so you have an image (that somehow is a float or something else) but only one channel but you are scaling it so that corners are seen better, right ?
Basically, you are proposing a way to enhance images for calibration ? The idea is good and could be done in several ways but yours is simple. Maybe there could simply be an --enhance flag that would have that behavior ? what do you think ?

stwirth:
I think we are talking about two different improvements:
1. Accepting floating point images as input, just the same way as image_view does
2. Enhancing the input image for better pattern detection

This patch is meant for 1. The scaling is rather simple, and analog to what image_view does. There exist conventions for floating point images to have values between 0.0 and 1.0, so a scaling by 255 to make it "normal" gray scale would be sufficient. In my use case I am using depth images which have values in meters so the scaling has to be somewhere around 25.5 to visualize values between 0.0 and 10.0. image_view handles this by looking for the maximum value present and adapting the scaling to make this value become 255, just as my proposed solution here.

Regarding 2. the current flags for checkerboard detection already trigger an enhancement (flags CV_CALIB_CB_ADAPTIVE_THRESH and CV_CALIB_CB_NORMALIZE_IMAGE) and the underlying blob detector of the circle pattern detector tries out a series of thresholds for binarization. So I am not sure if any other image enhancement could help. I would rather like to see more parameters for the blob detector exposed (which the OpenCV python api does not support unfortunately). It does not work well in my situation as the image is quite noisy. I suppose the [convexity check](https://github.com/Itseez/opencv/blob/master/modules/features2d/src/blobdetector.cpp#L88) fails. Regarding the checkerboard detector I never had problems using it and the currently used parameters seem to be fine.

vrabaud:
ok, make sense, sorry for the delay

PR #28, opened 2013-04-25
fixed Boost error on OSX

walchko:
I just installed Groovy on OSX and there were compile errors with boost::lock_guard. These changes allowed the packages to compile so I could finish the install.


vrabaud:
That seems very weird that we need it in side the code: that means somebody does not export those. Can you please copy/paste your error ? Thx

PR #29, opened 2013-05-07
Fix bc905fcd, include boost/version.hpp before using BOOST_VERSION macro

kartikmohta:
The old commit uses the BOOST_VERSION macro before it is defined, hence the boost/thread/lock_guard.hpp never gets included, leading to the same compile errors which the commit is supposed to fix.

Also, some files in depth_image_proc needed the same fix.


vrabaud:
duuh, thx. Weird it compiled for me, not sure where it got that macro from.

Issue #31, opened 2013-06-03
cameracheck.py error with kinect/ir topic


bassofil:
When running

```
rosrun camera_calibration cameracheck.py --size 7x5 monocular:=/camera/ir image:=image_raw
```

I get the following error

```
OpenCV Error: Assertion failed (src.depth() == dst.depth()) in cvCvtColor, file /tmp/buildd/ros-groovy-opencv2-2.4.5-2precise-20130519-1747/modules/imgproc/src/color.cpp, line 3921
Exception in thread Thread-4:
Traceback (most recent call last):
  File "/usr/lib/python2.7/threading.py", line 551, in __bootstrap_inner
    self.run()
  File "/opt/ros/groovy/lib/camera_calibration/cameracheck.py", line 84, in run
    self.function(m)
  File "/opt/ros/groovy/lib/camera_calibration/cameracheck.py", line 158, in handle_monocular
    C = self.image_corners(rgb)
  File "/opt/ros/groovy/lib/camera_calibration/cameracheck.py", line 145, in image_corners
    (ok, corners, b) = self.mc.get_corners(im)
  File "/opt/ros/groovy/lib/python2.7/dist-packages/camera_calibration/calibrator.py", line 378, in get_corners
    (ok, corners) = _get_corners(img, b, refine)
  File "/opt/ros/groovy/lib/python2.7/dist-packages/camera_calibration/calibrator.py", line 153, in _get_corners
    cv.CvtColor(img, mono, cv.CV_BGR2GRAY)
error: src.depth() == dst.depth()
```

Obviously, the following command works.

```
rosrun camera_calibration cameracheck.py --size 7x5 monocular:=/camera/rgb image:=image
```

I'm using Ubuntu 12.04 with ROS Groovy.

chadrockey:
I think @stwirth can help with this.

stwirth:
This looks like a bug in cv_bridge to me, @vrabaud might have a look. I submitted a pull request for a workaround.

PR #32, opened 2013-06-03
use mono calibrators mkgray in cameracheck.py, some variable renaming, 

stwirth:
fixes ros-perception/image_pipeline#31

Somehow the conversion of mono16 to bgr8 did not work, the [two-stage approach of cameracalibrator.py](https://github.com/ros-perception/image_pipeline/blob/groovy-devel/camera_calibration/src/camera_calibration/calibrator.py#L252-L271) does work though.


vrabaud:
Yop definitely. Thx. This file is way too big. Could this mkgray function belong to an independent class ? (it needs a cvbridge object so it cannot be a pure function; or we create a global variable but that's terrible)

PR #33, opened 2013-06-12
depth image can use different transport hint

jihoonl:
I figured out that compression for depth image should use compressedDepth instead of compressed.

This patch is to allow to accept different params to set transport hints for depth and rgb.


vrabaud:
Makes sense, thx !

PR #48, opened 2013-11-12
Improvements in calibration from tarfile

jbosch:
Improved the script to calibrate from a tarfile:
1- Added possibility to set different calibration flags: number of coefficients, fix ratio, set zero tangent coeffs.
  This would help users to try different calibration settings with the same images, so he can get the bests results without spending as much time as before (bagging, replaying bag, changing parameters)
2- Rectified images and corners detected displayed after calibration.
  This would help users to see quickly if the calibration was successful and furthermore check if corner detection failed in some images and remove it from the dataset for further calibrations.

Tested with mono and stereo datasets.

### DIFF
--- a/camera_calibration/scripts/tarfile_calibration.py
+++ b/camera_calibration/scripts/tarfile_calibration.py
@@ -49,10 +49,10 @@ import rospy
 import sensor_msgs.srv
 
 def waitkey():
-   k = cv.WaitKey(6)
-   if k in [27, ord('q')]:
-      rospy.signal_shutdown('Quit')
-   return k
+    k = cv.WaitKey(6)
+    if k in [27, ord('q')]:
+        rospy.signal_shutdown('Quit')
+    return k
 
 def cal_from_tarfile(boards, tarname, mono = False, upload = False, calib_flags = 0, alpha=1.0):
     if mono:
@@ -86,7 +86,7 @@ def cal_from_tarfile(boards, tarname, mono = False, upload = False, calib_flags 
 
     archive = tarfile.open(tarname, 'r')
     if mono:
-       for f in archive.getnames():
+        for f in archive.getnames():
             if f.startswith('left') and (f.endswith('.pgm') or f.endswith('png')):
                 filedata = archive.extractfile(f).read()
                 file_bytes = numpy.asarray(bytearray(filedata), dtype=numpy.uint8)
@@ -106,7 +106,7 @@ def cal_from_tarfile(boards, tarname, mono = False, upload = False, calib_flags 
                 cv2.imshow( f,  numpy.asarray( drawable.scrib[:,:] ))
                 k=-1;
                 while k ==-1:
-                  k=waitkey();
+                    k=waitkey();
                 cv2.destroyWindow(f) 
     else:
         limages = [ f for f in archive.getnames() if (f.startswith('left') and (f.endswith('pgm') or f.endswith('png'))) ]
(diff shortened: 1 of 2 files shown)


camera_calibration/scripts/tarfile_calibration.py
```diff
@@ -69,6 +81,82 @@ def cal_from_tarfile(boards, tarname, mono = False, upload = False):
...
+        for i in range(len(limages)):
+            l=limages[i];
+            r=rimages[i];
+
+            if l.startswith('left') and (l.endswith('.pgm') or l.endswith('png')) and r.startswith('right') and (r.endswith('.pgm') or r.endswith('png')):
+                # LEFT IMAGE
+                filedata = archive.extractfile(l).read()
+                file_bytes = numpy.asarray(bytearray(filedata), dtype=numpy.uint8)
+                im_left=cv2.imdecode(file_bytes,cv2.cv.CV_LOAD_IMAGE_COLOR)
+                im_left_rect=calibrator.l.remap(cv.fromarray(im_left));
```
vrabaud: that variable seems unused no ?

camera_calibration/scripts/tarfile_calibration.py
```diff
@@ -69,6 +81,82 @@ def cal_from_tarfile(boards, tarname, mono = False, upload = False):
...
+   
+                bridge = cv_bridge.CvBridge()
+                try:
+                    msg_left=bridge.cv_to_imgmsg(cv.fromarray(im_left), "bgr8")
```
vrabaud: I just added an API you might like in bridge: cv2_to_imgmsg  in https://github.com/ros-perception/vision_opencv 
I'll release it asap (I was waiting on some feedback but if you can give me some too, that would be nice. Thx !).

It's just that we need to kill "import cv", it's very old and not as well maintained.

camera_calibration/scripts/tarfile_calibration.py
```diff
@@ -36,17 +36,29 @@
...
+import tarfile
 
 from camera_calibration.calibrator import cvmat_iterator, MonoCalibrator, StereoCalibrator, CalibrationException, ChessboardInfo
 
 import rospy
 import sensor_msgs.srv
 
-def cal_from_tarfile(boards, tarname, mono = False, upload = False):
+def waitkey():
+   k = cv.WaitKey(6)
```
vrabaud: bad indentation no ? 3 spaces instead of 4.

camera_calibration/scripts/tarfile_calibration.py
jbosch: true

camera_calibration/scripts/tarfile_calibration.py
jbosch: your re right

camera_calibration/scripts/tarfile_calibration.py
jbosch: I've checked it. It works fine. Should I push it with this new version?

vrabaud:
Hi, thx for your contribution, it looks good. Things I would like to be improved:
- add an option to display things or not (sometimes you run that on a robot and cannot display anything thus the script would crash). An option like --visualize would do. Put the display code in a separate function
- Try removing "import cv" by using the proper cv2 API (like in waitKey) and using the updated bridge API: bridge.cv2_to_imgmsg (not released yet but please use master in groovy_devel there for testing or just wait for a few days).
  Thx !

vrabaud:
looks good, thx ! I was waiting for vision_opencv 1.10.13 to be out (it has the numpy support we mentioned before)

PR #49, opened 2013-11-19
adds conversion (if necessary) of the images for the calculation of disparity

diegodgs:
adds conversion (if necessary) using cvBridge the rectified image that will be used for the calculation of disparity

### DIFF
--- a/camera_calibration/package.xml
+++ b/camera_calibration/package.xml
@@ -13,6 +13,7 @@
 
   <test_depend>rostest</test_depend>
 
+  <run_depend>cv_bridge</run_depend>
   <run_depend>image_geometry</run_depend>
   <run_depend>message_filters</run_depend>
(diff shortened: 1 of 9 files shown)


vrabaud:
Two things:
- l_ptr = cv_bridge::toCvShare(l_image_msg, sensor_msgs::image_encodings::MONO8); actually takes care of not copying the data if the encodign is the same so you do not need the condition
- why not use l_image = cv_bridge::toCvShare(l_image_msg, sensor_msgs::image_encodings::MONO8).image;

vrabaud:
Hi, thx for your contribution: I am very sorry I had to rewrite history (there was an experimental commit that should have been on my personal branch and reverting did not feel write) but I did re-implement your feature in 751954e5e4d09daa4661b0c3d170eaa748d35c6b

PR #53, opened 2014-01-28
Added OpenCV libraries to linking.

jjh42:
I found that if the system OpenCV differs from OpenCV specified using `OpenCV_DIR` I had to add this to the CMake file to avoid linking errors.


vrabaud:
hi, thx a bunch for the PR: I actually realized the include_dirs were not even considered ..... so I wrote a bigger patch. Thx !

vrabaud:
fyi, just released it

PR #54, opened 2014-02-03
Added requirement for core.

jjh42:
I also required this line in order to successfully compile with a non-system opencv 2.4.8. Otherwise, I had linking errors.


vrabaud:
wow, that's weird: highgui should include core. That's a regression from upstream.

PR #56, opened 2014-02-23
Fix crash in approxsync.py

contradict:
First, remove some dead (if 0) code. It appears to be broken too.

If an input message in queues[0] was within slop of two messages in queues[1], both of these cases would appear since the for loop has already created the list to iterate over. When the second one is processed, the message in queues[0] has already been deleted, resulting in a crash when the list of messages was built.

Fix this by always checking to see if we still have all the messages to satisfy the slop condition.


vrabaud:
looks good, thx !

PR #63, opened 2014-04-09
Image rotate using tf2

garaemon:
image_rotate subscribes /tf and building tf tree consumes CPU resources a lot.
this patch supports tf2 to decrease CPU consumption. 
- ~use_tf2 parameter to switch tf and tf2

### DIFF
--- a/image_rotate/CMakeLists.txt
+++ b/image_rotate/CMakeLists.txt
@@ -1,7 +1,7 @@
 cmake_minimum_required(VERSION 2.8)
 project(image_rotate)
 
-find_package(catkin REQUIRED COMPONENTS cv_bridge dynamic_reconfigure image_transport roscpp tf nodelet)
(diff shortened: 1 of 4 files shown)


vrabaud:
what's the point of adding a flag for tf2 ? This is all hapennig behind the scene so you can switch to tf2 by default no ? (and remove tf: tf actually uses tf2 behind the scene).
I have done it in indigo with 7900a4536900386bab8ea288b729adab9c1660c1 . Should I backport it ?

garaemon:
> I have done it in indigo with 7900a45 . Should I backport it ?

cool!

the reason why I add tf2 flag is I don't know how many ros users uses tf2_buffer_server.

ahendrix:
@vrabaud I think this is reasonable; the use_tf2 flag uses the tf2 buffer server instead of subscribing directly to the /tf fire hose.

:+1:

garaemon:
thanks!

is it difficult to apply this patch to groovy-devel?

vrabaud:
@garaemon  , merged in groovy

PR #66, opened 2014-04-27
depth_image_proc: fix missing symbols in nodelets

mikeferguson:
Fixes several missing symbols I hit when running the openni2/rgbd pipeline. Tested under 14.04 and Indigo.


vrabaud:
thx !

PR #87, opened 2014-07-09
fix argument passing

iluetkeb:
Very simply fix for groovy-branch. Otherwise the cameracalibrator.py node will not work as described in the calibration tutorials!


vrabaud:
thx !

iluetkeb:
btw, I've got a few more changes coming up, there are many minor things broken. will submit once the whole thing works.

vrabaud:
thx ! On Groovy ?

iluetkeb:
yeah, on groovy (it's what our PR2 runs...).

I just noted that all the other changes are in the camera_pose stack, though, because they involve the camera_pose_calibration package.

PR #88, opened 2014-07-19
Add point_cloud_xyzi nodelet

jolting:
This is for cameras that output depth and intensity images.
It's based on the point_cloud_xyzrgb nodelet.


vrabaud:
I am not crazy about this one as it duplicates a lot of code. I'll try to generalize the idea so that we can add more point types.

PR #91, opened 2014-08-02
Fixes approximate time synchronization.

wnoise:
When the approximate time synchronizer moved to
ros_comm/utilities/message_filters, the argument order was changed so
that the slop argument was moved to the end, breaking the use of
functools.partial to give it the same interface as the exact
synchronizer.


vrabaud:
good catch, thx

PR #92, opened 2014-08-02
Handle rescaling.

wnoise:
No type argument in cv2.resize().


vrabaud:
thx. Will release those two fixes asap.

PR #96, opened 2014-08-26
Modifications for OpenCV 3.0

edgarriba:
Changed some interfaces for the new OpenCV release

### DIFF
--- a/image_proc/include/image_proc/processor.h
+++ b/image_proc/include/image_proc/processor.h
@@ -53,7 +53,7 @@ class Processor
 {
 public:
   Processor()
-    : interpolation_(CV_INTER_LINEAR)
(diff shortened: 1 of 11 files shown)


vrabaud:
hi, thx. I was on a break sorry: this is valuabl but you could please make sure it compiles on OpenCV 2.4.5 (the minimum one supported on Ubuntu). Thx. I get errors like: fourcc does not exist, or StereoBM does not have the same API. If it compiles with OpenCV 2.4.9 and 3.0, just use an #if as follows:

```
#if (CV_MAJOR_VERSION > 2) || ((CV_MAJOR_VERSION==2) && ((CV_MINOR_VERSION>4) || (CV_MINOR_VERSION==4 && CV_SUBMINOR_VERSION>=9)))
```

PR #104, opened 2014-10-24
Publish range_max for invalid readings

efernandez:
This is useful for clearing in `costmap_2d`.

### DIFF
--- a/depth_image_proc/src/nodelets/point_cloud_xyz.cpp
+++ b/depth_image_proc/src/nodelets/point_cloud_xyz.cpp
@@ -53,6 +53,7 @@ class PointCloudXyzNodelet : public nodelet::Nodelet
   boost::shared_ptr<image_transport::ImageTransport> it_;
   image_transport::CameraSubscriber sub_depth_;
   int queue_size_;
+  double range_max_;
 
   // Publications
   boost::mutex connect_mutex_;
@@ -81,6 +82,7 @@ void PointCloudXyzNodelet::onInit()
 
   // Read parameters
   private_nh.param("queue_size", queue_size_, 5);
+  private_nh.param("range_max", range_max_, 0.0);
 
   // Monitor whether anyone is subscribed to the output
   ros::SubscriberStatusCallback connect_cb = boost::bind(&PointCloudXyzNodelet::connectCb, this);
@@ -160,8 +162,15 @@ void PointCloudXyzNodelet::convert(const sensor_msgs::ImageConstPtr& depth_msg, 
       // Missing points denoted by NaNs
       if (!DepthTraits<T>::valid(depth))
       {
-        pt.x = pt.y = pt.z = bad_point;
-        continue;
+        if (range_max_ != 0.0)
+        {
+          depth = DepthTraits<T>::fromMeters(range_max_);
+        }
+        else
+        {
+          pt.x = pt.y = pt.z = bad_point;
+          continue;
+        }
       }
 
       // Fill in XYZ
--- a/depth_image_proc/src/nodelets/point_cloud_xyzrgb.cpp
+++ b/depth_image_proc/src/nodelets/point_cloud_xyzrgb.cpp
@@ -82,6 +82,7 @@ class PointCloudXyzrgbNodelet : public nodelet::Nodelet
   typedef ApproximateTime<sensor_msgs::Image, sensor_msgs::Image, sensor_msgs::CameraInfo> SyncPolicy;
   typedef message_filters::Synchronizer<SyncPolicy> Synchronizer;
   boost::shared_ptr<Synchronizer> sync_;
+  double range_max_;
 
   // Publications
   boost::mutex connect_mutex_;
@@ -117,6 +118,7 @@ void PointCloudXyzrgbNodelet::onInit()
   // Read parameters
   int queue_size;
   private_nh.param("queue_size", queue_size, 5);
+  private_nh.param("range_max", range_max_, 0.0);
 
   // Synchronize inputs. Topic subscriptions happen on demand in the connection callback.
   sync_.reset( new Synchronizer(SyncPolicy(queue_size), sub_depth_, sub_rgb_, sub_info_) );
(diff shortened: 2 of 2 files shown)


vrabaud:
thx ! looks good.

trainman419:
I'm not convinced this is a good idea. There are lots of cases where a depth sensor can get depth dropouts, and maximum range is only one of those.

Does the depthimage format have any specific notations that differentiate between readings at the sensor's maximum range and depth points that are missing for another reason?

mikeferguson:
I think I agree with Austin, I see this as a regression against [REP117](http://www.ros.org/reps/rep-0117.html), previously it did conform with "Erroneous detections shall be represented by quiet (non-signaling) NaNs" (which is what bad_point is).

Really, this should probably be implemented in the navigation stack, already the obstacle layer has "inf_is_valid" as a parameter, which "Allows for Inf values in "LaserScan" observation messages. The Inf values are converted to the laser maximum range."

chadrockey:
@efernandez @trainman419 and @mikeferguson are all correct.

With primesense devices (since the uint16_t format only has noresponse/error == 0), you end up having to use NaNs to clear the costmap.  Since this is reported as uints from the hardware @trainman419, I haven't found a way to determine the different cases.

However, it also reports all 0s if you say.... blocked the projector.  Or stepped into the deadband.  So if you step into the deadband, you were an obstacle at 0.5m, now you get cleared at 0.5m, and now the robots plans through you and can't sense you with the primesense.

Ultimately @mikeferguson is correct that this has to be implmented for the various cases people want by the nav stack.  Some hints may be adding other sensors like a laser rangefinder to cover the deadzone or some other approach you develop.

@vrabaud For image_pipeline this should be reverted because it will lead to clearly broken cases and isn't "a difference of implementation" between robots.

efernandez:
@trainman419 @mikeferguson @chadrockey You're right when you say that there's no way to distinguish between the invalid readings coming from far away readings (no "echo"/"return") and close ones (blanking area). I'm happy that you've discussed about this.

The main problem of implemented this on the `costmap_2d` observation buffers is that AFAIK there's no camera calibration information (principle point and focal distances) to compute the reprojection from the image plane (at the `range_max`) into the 3D PC. However, if anybody knows a clean (and efficient) way to implement this there, I'd be happy to do it. But still, the problem to distinguish far away from close invalid readings is still there. I'd say it's intrinsic to some sensors, but with this PR I tried to give the option (note that with `range_max == 0.0` the implementation does the same as before) to have points for clearing in `costmap_2d`.

@DLu I think you might have something to say here. I also have the following fundamental question: How does `costmap_2d` manage the clearing when using PCs coming from RGBD sensors? I mean, when there's no reading because they're so far away, there's no endpoint to clear using raytracing, so AFAIK nothing is cleared.

mikeferguson:
RE: implementing "inf_is_valid" or "nan_is_valid" for point clouds in the costmap_2d: The camera info needed for the reprojection is in the camera_info topic (neither image nor point clouds contain the reprojection info).

You are correct that currently the costmap_2d does not currently implement "nan_is_clearing" or similar for point clouds, but I think the costmap_2d is the right place to implement that as an option -- not the image_pipeline. If you put it into the image pipeline, then it will affect all applications that depend on the particular point cloud you are modifying (which could cause all sorts of unintended side effects).

efernandez:
@mikeferguson Thanks for your reply. Sure, I understan the problem of doing this on the image_pipeline. I'll try to implement something in the costmap_2d, where I'll have to subscribe to the camera_info topic... maybe, it make sense to add a new observation source: the depth image. What do you think?

@DLu

DLu:
I concur with having this as part of costmap_2d. To keep layers from becoming too complicated, I would lean toward making a new layer to subscribe to DepthImages and CameraInfos.

efernandez:
@DLu Could you explain a little how and where you would implement this?
For me it doesn't look like a new layer, but just like another observation source type.
Anyway, I have to look in more detail into it before implementing it.

DLu:
I think implementing it as another observation source would be fine too.

efernandez:
@DLu @vrabaud @trainman419 @mikeferguson @chadrockey I've added a `DepthImage` `data_type` to the `costmap_2d` (into the `ObstacleLayer` plugin). There are two things I'd like to clean before sending the PR:
- I use the `convert` method and the `DepthTraits` class from `depth_image_proc`; I had to copy them because there's no library exporting them. Where could I put them in order to have a library for the conversion from depth image to point cloud? Somewhere in the `image_pipeline` of maybe in `image_geometry`?
- I have to synchronize the `sensor_msgs::Image`, `sensor_msgs::CameraInfo` and the TF for the global frame in the `costmap_2d`. Because of that, I cannot use `image_transport`. I tried, but it wasn't possible because there's no `CameraSubscriberFilter` and the `SubscriberFilter` cannot inherit from `message_filters::SubscriberBase`. I've ended up using a `TimeSynchronizer` in order to synchronize these 3 elements (the callback receives the `sensor_msgs::Image` and `sensor_msgs::CameraInfo`, apart from the observation buffer). It would be much cleaner if I only synchronize the `sensor_msgs::Image` with the TF for the global frame, and I add a callback for the `sensor_msgs::CameraInfo`, so I can take the last camera info and save it on an attribute. After all, the camera info doesn't change normally, so IMHO it's a good compromise to add the `DepthImage` with the minimal changes.

mikeferguson:
- I think I've seen DepthTraits copied into both the openni and openni2_camera packages as well -- exporting as a library would be good -- isn't it templated though? If so, all we need is to install the header?
- I see no need to sync the CameraInfo with the Image -- in fact, I'm not sure all drivers publish CameraInfo at the same time/rate as the Image topic, and I'm pretty sure most ROS code existing today does not sync them.

efernandez:
@mikeferguson Thanks! I'll send both PR tomorrow.

efernandez:
@DLu @mikeferguson @trainman419 @vrabaud @chadrockey I've create the following PRs:
- https://github.com/ros-perception/image_pipeline/pull/108 : Exports two headers to avoid code duplication and adds the `range_max` argument with a default value that keeps the previous behavior.
- https://github.com/ros-planning/navigation/pull/265 : Adds the `DepthImage` `data_type` in `costmap_2d` (in the `ObstacleLayer`).

PR #108, opened 2014-10-29
Export headers for DepthTraits and convert

efernandez:
This PR exports `DepthTraits<T>` and `convert` for `xyz` PC in two separate headers to avoid code duplication in other places, in particular in an upcoming PR following https://github.com/ros-perception/image_pipeline/pull/104.

The `convert` function is also extended to support to allow for setting `range_max` for invalid (when there's no "echo"/"return") pixels.

The `CMakeLists.txt` is also slightly cleaned up.


efernandez:
If needed, I can also export all the other `convert` methods that are used by the other nodelets; I don't need them in https://github.com/ros-planning/navigation/pull/265 though.

vrabaud:
fine for now. The other conversions are actually higher-level between data types. We could make those public at some point. I'll forward port that to Indigo. Thx.

mikeferguson:
@efernandez This also needs to be added to indigo, and then we need a new hydro and indigo release before the upstream navigation change can be merged.

efernandez:
@mikeferguson Do you want me to send a PR rebasing my changes on top of `indigo-devel`? Or is  @vrabaud taking care of that?

mikeferguson:
I didn't notice until after I posted, but @vrabaud said he would take care of it -- I presume he'll just cherry pick it into indigo, but he is a busy guy, so a PR for indigo-devel might make that easier for him.

vrabaud:
indeed :) But actually, I removed any PCL dependency in Indigo. What would you need there ?

efernandez:
@mikeferguson @vrabaud I've cherry-pick'd the changes for `indigo` in https://github.com/ros-perception/image_pipeline/pull/109

Note that I also take care of removing `PCL`, but I couldn't compile it on my ROS `hydro` system.
Could you check that it compiles correctly? I think there are automatic jobs for the PRs, so I guess I'll have a warning anyway if something is broken.

PR #109, opened 2014-10-31
Export headers for DepthTraits and convert for indigo

efernandez:
This PR cherry-picks https://github.com/ros-perception/image_pipeline/pull/108 onto `indigo` branch.

Note that it doesn't use `PCL`.


efernandez:
@mikeferguson @vrabaud Please, let me know whether something else is needed here or not.
Once this is merged, we can also merge https://github.com/ros-planning/navigation/pull/265

efernandez:
I've just verified that this compiles without errors in a clean ROS indigo installation; only one header update was missed, and I've just updated the PR.

depth_image_proc/include/depth_image_proc/depth_conversions.h
```diff
@@ -0,0 +1,101 @@
...
+
+// Handles float or uint16 depths
+template<typename T>
+void convert(
+    const sensor_msgs::ImageConstPtr& depth_msg,
+    PointCloud::Ptr& cloud_msg,
+    const image_geometry::PinholeCameraModel& model,
+    double range_max = 0.0)
+{
+  // Use correct principal point from calibration
```
vrabaud, line 56: it would be better if that function could take an empty cloud_msg. Conditions need to be made and initializations too (basically what is done in the other file you modified).

vrabaud:
sorry for my late comments, I did not get much bandwidth lately.

vrabaud:
any comment on the fact it should take an empty point cloud as input ?

vrabaud:
@efernandez  ?

vrabaud:
merging for now, will add the emptiness checks manually

PR #113, opened 2014-12-15
Change duplicated function name

garyservin:
When linking statically (in our case on Android) and because image_view is linked against image_transport, the increment function is duplicated (https://github.com/ros-perception/image_common/blob/483312b85671ec61e8aa15f96a52e0e493df53d2/image_transport/src/camera_subscriber.cpp#L41).

Is this the correct approach or would someone suggest to handle differently?

### DIFF
--- a/image_view/src/nodes/stereo_view.cpp
+++ b/image_view/src/nodes/stereo_view.cpp
@@ -323,7 +323,7 @@ static unsigned char colormap[768] =
     255,  0, 0,
   };
 
-void increment(int* value)
+inline void increment(int* value)
 {
   ++(*value);
 }


vrabaud:
would inlining solve the problem ?
(I can merge this PR, I am just curious)

garyservin:
You're right! It worked. Should I create a new PR or should I just squash these changes?

vrabaud:
does not matter, squashing will make that disappear so please go ahead with
that.

garyservin:
@vrabaud done!

vrabaud:
thx ! While you're at it, could you please make a PR for image_common ? thx.

vrabaud:
thx, I will release image_pipeline as soon as I have confirmation other bugs are fixed (should be this week). Please let me know if this is actually more urgent.

PR #115, opened 2015-02-27
change point_step to 32 and rgb offset to 16, as depth_image_proc

hyaguchijsk:
According to this evaluation,
https://github.com/garaemon/pcl_ros_conversion_benchmark
conversion from `sensor_msgs/PointCloud2` to `pcl::PointCloud` is too slow when fields mismatched.

I found this conversion problem using `stereo_image_proc`, and compared message format with `depth_image_proc`.
Then, there are 2 diffs;
- `point_step` is 16, `depth_image_proc` is 32
- `offset` of `rgb` is 12, `depth_image_proc` is 16

I changed these parameters and check the conversion is fast enough with this modification.

### DIFF
--- a/stereo_image_proc/src/nodelets/point_cloud2.cpp
+++ b/stereo_image_proc/src/nodelets/point_cloud2.cpp
@@ -49,6 +49,7 @@
 #include <stereo_msgs/DisparityImage.h>
 #include <sensor_msgs/PointCloud2.h>
 #include <sensor_msgs/image_encodings.h>
+#include <sensor_msgs/point_cloud2_iterator.h>
 
 namespace stereo_image_proc {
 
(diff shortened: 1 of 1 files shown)


vrabaud:
Thx for your help. Would you mind actually re-writing that code with the PointCloud2 iterator ? It would get rid of such errors and would simplify the code (it automatically takes care of the padding problems you solve).
https://github.com/ros/common_msgs/blob/jade-devel/sensor_msgs/include/sensor_msgs/point_cloud2_iterator.h
If you're interested in performance, you could also reqrite the isValidPoint(mat(v,u)) and use a proper matrix iterator. Thx !

hyaguchijsk:
Ok, I try to rewrite. Thanks!
2015/02/27 21:34 "Vincent Rabaud" notifications@github.com:

hyaguchijsk:
I modified using `sensor_msgs/point_cloud2_iterator.h`, referring to `depth_image_proc`.

stereo_image_proc/src/nodelets/point_cloud2.cpp
```diff
@@ -174,48 +175,34 @@ void PointCloud2Nodelet::imageCb(const ImageConstPtr& l_image_msg,
...
-  points_msg->fields[2].datatype = PointField::FLOAT32;
-  points_msg->fields[3].name = "rgb";
-  points_msg->fields[3].offset = 12;
-  points_msg->fields[3].count = 1;
-  points_msg->fields[3].datatype = PointField::FLOAT32;
-  //points_msg->is_bigendian = false; ???
-  static const int STEP = 16;
-  points_msg->point_step = STEP;
-  points_msg->row_step = points_msg->point_step * points_msg->width;
-  points_msg->data.resize (points_msg->row_step * points_msg->height);
```
vrabaud, line 198: you also need to resize the point cloud before hand

hyaguchijsk, line 198: In `PointCloud2Modifier::setPointCloud2FieldsByString(int n_fields, ...)`,
`data.resize()` is called at last column.
https://github.com/ros/common_msgs/blob/indigo-devel/sensor_msgs/include/sensor_msgs/impl/point_cloud2_iterator.h#L227
So I think it doesn't need to call `data.resize()` expressly.

vrabaud, line 198: you're totally right :) thx !

stereo_image_proc/src/nodelets/point_cloud2.cpp
```diff
@@ -231,18 +217,10 @@ void PointCloud2Nodelet::imageCb(const ImageConstPtr& l_image_msg,
...
-          uint8_t g = color(v,u);
-          int32_t rgb = (g << 16) | (g << 8) | g;
-          memcpy (&points_msg->data[offset + 12], &rgb, sizeof (int32_t));
-        }
-        else
-        {
-          memcpy (&points_msg->data[offset + 12], &bad_point, sizeof (float));
-        }
+        uint8_t g = color(v,u);
+        *iter_r = *iter_g = *iter_b = g;
```
vrabaud, line 223: that seems to change the behavior: before, when the point was not valid (e.g. MISSING_Z which is 10000), it would be set to infinity. Now, it is se to 10000

hyaguchijsk, line 223: You mean the behavior of xyz copying I guess, I did not change behavior.
It is around L196, please check it.
https://github.com/ros-perception/image_pipeline/pull/115/files#diff-7ad2f03337d702facd01ce7e3b65a7e0R196

This section is color information copying,
but, yes, I changed the behavior.

Before, if the point is invalid, then `NaN(float)` will be copied into `rgb(uint8_t[3] + 1 byte)` and it has no mean in `uint8_t[3]`. It is strange I think, so after, I changed all color information will be copied whether the point is valid or not.

And I'd like to change behavior to copy color if the point is invalid,
because we can reconstruct color image from only point cloud like `depth_image_proc`.

vrabaud, line 223: I was talking about color. Your point makes sense. Plus, it did not make sense to copy a float NaN in an int that is supposed to store RGB ....

vrabaud:
Actually, if performance is an issue for you, you could open a new pull request were mat(v,u) is not used but an iterator or pointer per line: that would make things muuuuuuch faster.

hyaguchijsk:
OK, I'll try. Thanks!

2015-03-05 19:46 GMT+09:00 Vincent Rabaud notifications@github.com:

PR #120, opened 2015-03-18
Bugfix/119 camera info

cbames:
This is a fix for issue 119.


cbames:
I must be missing something... after looking at the output of the rectifier it appears i've messed something up.

cbames:
It appears that the rectification node bins the image based on binning_x...

cbames:
found my solution already the API. ignore this.

PR #121, opened 2015-03-19
Bound amount of memory used by camera calibration

### LINKED ISSUE #112: memory grows unbounded during calibration
I have a problem where, during calibration, memory usage increases very fast. If I manage to avoid receiving a segmentation fault do to excessive memory usage, I see that after calibration process ends, the calibration window displays all past images captured during the calibration process. 

I've looked at the code and I've seen that images are queued and I think the queue has no limit. I wonder if this is the cause. If so, the queuing should be stopped before starting calibration to avoid this problem.

bobbens:
Added the possibility to limit the size of the image queues in cameracalibrator.py. Should close #112.

### DIFF
--- a/camera_calibration/nodes/cameracalibrator.py
+++ b/camera_calibration/nodes/cameracalibrator.py
@@ -95,7 +95,7 @@ class ConsumerThread(threading.Thread):
 
 class CalibrationNode:
     def __init__(self, boards, service_check = True, synchronizer = message_filters.TimeSynchronizer, flags = 0,
-                 pattern=Patterns.Chessboard, camera_name='', checkerboard_flags = 0):
+                 pattern=Patterns.Chessboard, camera_name='', checkerboard_flags = 0, queue_size = 0):
         if service_check:
             # assume any non-default service names have been set.  Wait for the service to become ready
             for svcname in ["camera", "left_camera", "right_camera"]:
@@ -110,6 +110,7 @@ class CalibrationNode:
                         print("Service not found")
                         rospy.signal_shutdown('Quit')
 
+        self._queue_size = queue_size
         self._boards = boards
         self._calib_flags = flags
         self._checkerboard_flags = checkerboard_flags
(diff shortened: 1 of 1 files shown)


bobbens:
I set the default to 0 so it behaves like the old way, but I'm guessing it might be a good idea to set it to some other value so nobody runs out of memory by default.

camera_calibration/nodes/cameracalibrator.py
```diff
@@ -370,6 +371,9 @@ def main():
     parser.add_option("-c", "--camera_name",
                      type="string", default='narrow_stereo',
                      help="name of the camera to appear in the calibration file")
+    parser.add_option("-Q", "--queue_size",
+                     type="int", default=0,
+                     help="maximum size of the image queues to use")
```
vrabaud: can you please just add: "0 means infinity"

vrabaud:
I would be ok with a default value of 30. Does that fix #112  for anybody ?

bobbens:
Done. This fixes #112 for me, which is why I wrote the patch in the first place ;)

vrabaud:
ok, sorry to come on that one again, but it would be fine to have a queue of size 1. So how about having, as an only change, a queue of size 1 (like 134/135). Does that work for you ?

bobbens:
While for exact timing that would work, it will not work well when doing approximate sync I believe. If you think that won't be a problem I don't have any issue with changing it, but I believe it will be.

vrabaud:
by synchronization you mean that if the image takes 2s to process, what's in the queue is from 2s ago when there is free CPU, or do you mean something else ? if that's the case, I'll just change the queue to a unique image with a lock.

bobbens:
I mean the approximate timer sync appears to use some sort of fancy algorithm in which tries to find the optimal mapping between the two different streams. If you don't have a queue it can't actually do this optimization and thus might not be able to find good pairs of synchronized images.

vrabaud:
the synchronization is done independently from those queues (it's internal to the ROS queues and TimeSynchronizer). The Queue is simply not adapted. Let me implement a lock instead to only have the latest image / pair of images.

vrabaud:
can you please check out master and see if that works for you ? I replaced all queues with deques of size 1.

YoshuaNava:
I had the same problem, and now it is solved.

Thank you!!! :D

PS: My PC has an AMD Phenom II N830 CPU and 4GB's of RAM. My CPU doesn't have SSSE3, so I have experienced some problems with really instruction-set-optimized ROS packages. But now everything is working alright with this package.

vrabaud:
@bobbens , can I close that pull request ?

bobbens:
No problem.

PR #122, opened 2015-03-23
Nodelet for handling radial depth images

jolting:
I have a camera that outputs radial depth images.

theta = angle to principle axis for a particular point
r = range on a particular point
Z= cos(theta) \* r

It's easier just to calculate the points from the range image directly instead of going to the whole image pipeline.

cv::undistortPoints gives you points in normalized image coordinates. This is useful because we can assume the focal length = 1. Since our data is radial we can just generate 3d vectors from the focal point to the point on the image plane and normalize the vector. I think this technique should work for axial based depth images as well. Just don't normalize the vector.


depth_image_proc/src/nodelets/point_cloud_xyz_radial.cpp
```diff
@@ -0,0 +1,235 @@
...
+	// Handles float or uint16 depths
+	template<typename T>
+	void convert(const sensor_msgs::ImageConstPtr& depth_msg, PointCloud::Ptr& cloud_msg);
+    };
+
+    cv::Mat initMatrix(cv::Mat cameraMatrix, cv::Mat distCoeffs, int width, int height, bool radial)
+    {
+	int i,j;
+	int totalsize = width*height;
+	cv::Mat pixelVectors(1,totalsize,CV_32FC3);
```
vrabaud, line 85: please use templated matrices all over for maintainability: cv::Mat_cv::Vec3f pixelVectors(1, totalsize);

vrabaud:
looks good except for the comment. Thx

PR #131, opened 2015-05-12
Correct three errors that prevented the node to work properly.

bassofil:
The node wasn't working in Ubuntu 14.04 with ROS Indigo and Python 2.7.6.


vrabaud:
thx, it works !

PR #137, opened 2015-06-12
Don't use GTK for android

garyservin:
Avoid using GTK when building for Android

### DIFF
--- a/image_view/CMakeLists.txt
+++ b/image_view/CMakeLists.txt
@@ -17,7 +17,9 @@ include_directories(SYSTEM ${Boost_INCLUDE_DIRS}
                            ${GTK2_INCLUDE_DIRS}
                            ${OpenCV_INCLUDE_DIRS}
 )
-add_definitions(-DHAVE_GTK)
+if (NOT ANDROID)
+  add_definitions(-DHAVE_GTK)
+endif()
 
 
 # Nodelet library


trainman419:
It's probably better to do this by checking the cmake `GTK_FOUND` variable and setting `-DHAVE_GTK` appropriately.

vrabaud:
what's the behavior on Android for `find_package(GTK2)` ? Should that be a `find_package(GTK2 QUIET)` ?

vrabaud:
actually, image_view is a GUI. For Android, we should simply skip the GUI executables / libs no ?

vrabaud:
@garyservin , any comments on the above ?

garyservin:
Sorry for the delay!

I agree with skipping GUI executables and libs.

Just to confirm. We should skip image_view, image_view_exe disparity_view, stereo_view, right?

This leaves extract_images image_saver and video_recorder to be compiled on Android. What's the best approach for  installing the targets? To separate the GUI ones from the extra tools perhaps?

vrabaud:
can you pleae confirm master works for you ? Thx

trainman419:
I still think that this should be done differently:

```
find_package(GTK2)
if(GTK2_Found)
   ... GUI executables here
endif()
```

This will better handle other platforms where a GUI or GTK2 may not be available.

vrabaud:
well, the package has a dependency on GTK so it will be installed anyway during build time. Or you prevent that on Android ?

garyservin:
I can confirm that it builds on Android!

GTK does not create any issue when building and I haven't seen anything in the logs about it

PR #145, opened 2015-07-20
Initial version of image_flip

awesomebytes:
I thought it may make sense to make a standalone node for the flipping of images so here is my approach.


vrabaud:
superseded by #173

awesomebytes:
I ended up finding my own PR as I needed the same thing again to flip an arbitrary topic, so I made I a little Python node for it:
https://gist.github.com/awesomebytes/44164730f38bea1a3b3de5b4eacc5d4c

PR #147, opened 2015-08-20
Added sensor_msgs::Image conversion to cv::Mat from rqt_image_view in order to be able to create videos from kinect depth images (cv_bridge currently doesn't support 16UC1 image encoding).

carlosmccosta:
Code adapted from:
https://github.com/ros-visualization/rqt_common_plugins/blob/groovy-devel/rqt_image_view/src/rqt_image_view/image_view.cpp

### DIFF
--- a/image_view/src/nodes/video_recorder.cpp
+++ b/image_view/src/nodes/video_recorder.cpp
@@ -18,8 +18,8 @@
 *****************************************************************************/
 
 #include <opencv2/highgui/highgui.hpp>
-
 #include <ros/ros.h>
+#include <sensor_msgs/image_encodings.h>
 #include <cv_bridge/cv_bridge.h>
 #include <image_transport/image_transport.h>
 #include <camera_calibration_parsers/parse.h>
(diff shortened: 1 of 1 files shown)


vrabaud:
hmmm, what do you think of actually having this functionality be part of cv_bridge ? Maybe as a function like toDisplayCvShare or something ?

carlosmccosta:
Thanks for the feedback :)
Integration within cv_bridge would be ideal to reduce code duplication in the future.
I created a PR with the implementation of the function
CvImageConstPtr cvtColorForDisplay(const sensor_msgs::ImageConstPtr& source,
                                   const std::string& encoding,
                                   bool use_dynamic_image_value,
                                   double min_image_value,
                                   double max_image_value)

PR:
https://github.com/ros-perception/vision_opencv/pull/75

Have a nice day :)

vrabaud:
Hey, can you please squash your commits ? Thx !

carlosmccosta:
Squashed the 3 commits into 1 as requested.
Thanks for improving cv_bridge :)

Have a nice day :)

PR #157, opened 2015-11-09
Convert 16UC1 to RGB to visualize depth image

wkentaro:
image

### DIFF
--- a/image_view/src/nodelets/image_nodelet.cpp
+++ b/image_view/src/nodelets/image_nodelet.cpp
@@ -34,6 +34,7 @@
 #include <ros/ros.h>
 #include <nodelet/nodelet.h>
 #include <image_transport/image_transport.h>
+#include <sensor_msgs/image_encodings.h>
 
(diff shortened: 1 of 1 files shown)


k-okada:
+1

wkentaro:
Close via https://github.com/ros-perception/vision_opencv/pull/81

vrabaud:
well, actually, I did not close your PR because you have dynamic scaling for depth images which https://github.com/ros-perception/image_pipeline/commit/46667b4a0255fe15d9c17da0b1734a5e411dd0b9 does not introduce.

image_view would need a parameter to enable/disable that (but then again, rqt_image_view has it). But before adding it, image_view needs to be re-written to remove its dependency on the nodelet: this created a lot of bugs with image_view because of OpenCV's threading and the nodelet threading.

If you feel like it, please re-write image_view to simply have an image subscriber and display the image using cvtColorForDisplay. This is probably 20 lines, I just haven't had time ....

PR #159, opened 2015-11-20
Extract images which are synchronized with message_filters

wkentaro:
This is a synchronizing two images version of extract_images.cpp on image_view.

### DIFF
--- a/image_view/CMakeLists.txt
+++ b/image_view/CMakeLists.txt
@@ -80,3 +80,9 @@ install(TARGETS disparity_view image_view_exe stereo_view
 install(FILES nodelet_plugins.xml
         DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION}
 )
+
+# Python programs
+catkin_install_python(
+  PROGRAMS scripts/extract_images_sync
+  DESTINATION ${CATKIN_PACKAGE_BIN_DESTINATION}
+)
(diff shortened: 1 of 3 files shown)


vrabaud:
Please add a license, and a doc on top of the file to say what it does.
You also need a catkin_install_python to install the script.

wkentaro:
Thanks, i will change it.

I think there is high possibility that someone will need a node which subscribes more than 2 topics.
So will change it to set subscribing topics by rosparam.

wkentaro:
I have done the changes.

wkentaro:
Could you please review and merge?

vrabaud:
ok, took a while to get the bandwidth: couldn't that be an extra parameter in https://github.com/ros-perception/image_pipeline/blob/indigo/image_view/src/nodes/extract_images.cpp#L144 ? Like called "images".

wkentaro:
That's possible but it's hard to write, because c++ does not have variable length arguments.
https://github.com/ros-perception/perception_pcl/blob/jade-devel/pcl_ros/src/pcl_ros/io/concatenate_data.cpp#L118-L181

Could I replace c++ `extract_indices` to python?

vrabaud:
ok, I indeed do not see any better solution. Maybe ROS2 and C++11 with variadic templates will help. In which case, we can remove the Python script and merge that behavior into C++.
Thx !

wkentaro:
> ok, I indeed do not see any better solution. Maybe ROS2 and C++11 with variadic templates will help. 

I'm happy to hear that, thanks.

PR #163, opened 2015-11-24
[image_view] Request start/end saving request by std_srvs/Trigger

wkentaro:


wkentaro:
Could you please review and merge?

vrabaud:
a header seems like a heavy interface ? Why not a dynamic parameter that is a ros::Time ?

wkentaro:
so user should send request like http://wiki.ros.org/hokuyo_node/Tutorials/UsingDynparamToChangeHokuyoLaserParameters#PythonAPI ?

wkentaro:
In that case, dynamic parameter interface is heavier I think.

vrabaud:
ok, you are right about dynamic_reconfigure.
The simplest you should use is a trigger service: http://docs.ros.org/api/std_srvs/html/srv/Trigger.html Please also need to rebase.
Very very sorry about the slowness these days, life gets in the way ...

wkentaro:
The reason I used header as the request is to pass the timestamp to the image_view.
With trigger service, it cannot pass the timestamp so there will be a time lag between the time user really requests and actual one.

wkentaro:
But I can use the rostime when the Trigger service is requested as the time user desires, so I will do that.

wkentaro:
I have done. Please review again.

wkentaro:
Friendly, ping.

wkentaro:
ping.

wkentaro:
ping

wkentaro:
ping

vrabaud:
Very very sorry for the delay, life got inthe way ...

wkentaro:
Never mind. Thanks!

PR #164, opened 2015-11-24
[image_view] Save multiple image topic images by request

wkentaro:


wkentaro:
Could you please review and merge this?

vrabaud:
couldn't a dynamic reconfigure be used instead of an extra header ? Couldn't that be merged with #159 and #163 ?

wkentaro:
> couldn't a dynamic reconfigure be used instead of an extra header ?

that's possible.

> couldn't a dynamic reconfigure be used instead of an extra header ?

the difference between `image_saver` and `extract_images` is whether saving images with request or not.
so #159 and #163 is different.

wkentaro:
c++ `image_saver` can be replaced by python version as well?
https://github.com/ros-perception/image_pipeline/pull/159#issuecomment-162412878

vrabaud:
Pleae also use a trigger service: http://docs.ros.org/api/std_srvs/html/srv/Trigger.html

wkentaro:
Rebased.

wkentaro:
Could you please merge this?

wkentaro:
Any comments, please?

wkentaro:
Any comments, please?

wkentaro:
friendly, ping.

wkentaro:
Could you please merge this?

vrabaud:
so this is like extract_images_sync.py except you can trigger it right ?
Why not add a parameter to extract_images_sync.py to tell it to start recording as soon as the script start ? It would be true by default (to not change the current behavior), and false when you want to mimic your current script. You would then add the same start/end services.
It seems like a lot to re-add a new file (that also does not handle properly colorspaces like extract_images_sync.py

wkentaro:
Yes, you're right.
These nodes are quite similar, and adding new image_save_sync is not a good idea.
So I'm closing this PR now, and will send PR about the trigger when I need that feature next time.

PR #166, opened 2015-11-24
[image_view/image_view] Make as a node

wkentaro:


vrabaud:
ok, thx for that ! So it's a good first try but it does not solve the threading issues that I actually never explained to you ...

OpenCV's GUI control can have issues if it's not the main thread of a process that creates and updates the window. That's it. Your code most likely works but maybe under Qt, GTK or some other conditions, things can get crazy.

So, your code needs to be re-written as follows to let OpenCV handle the whole threading:
- remove all the GTK stuff, you should not need it anymore: purely depend on OpenCV.
- in your main, have a while ros::ok loop
- have a call back that just fills in a cv::Mat variable and unlocks a condition variable
- in the main while loop, after the condition variable is set, display that cv::Mat and wait for the next one.

This way, the creation of window happens in the main thread (with the ImageViewer constructor) and it's update too (right know, it's happening in the callback queue thread).

Thx for updating !

wkentaro:
Updated please review. 
There is an image (which is brown or deep blue) at first time I open image_view with this commit.. I don't know why.

wkentaro:
image
image

image_view/src/nodes/image_view.cpp
```diff
@@ -42,13 +97,40 @@ int main(int argc, char **argv)
...
+  image_transport::TransportHints hints(transport, ros::TransportHints(), local_nh);
+  image_transport::Subscriber sub = it.subscribe(topic, 1, imageCb, hints);
+
+  while (ros::ok()) {
+    if (g_new_msg && !g_last_image.empty()) {
+      const cv::Mat &image = g_last_image;
+      cv::imshow(window_name, image);
+      g_new_msg = false;
+    }
+    ros::spinOnce();
```
vrabaud: you probably want to add a sleep to avoid busy idling no ?

image_view/src/nodes/image_view.cpp
```diff
@@ -32,7 +32,62 @@
...
+
+  const cv::Mat &image = g_last_image;
+
+  if (image.empty()) {
+    ROS_WARN("Couldn't save image, no data!");
+    return;
+  }
+
+  std::string filename = (g_filename_format % g_count).str();
+  if (cv::imwrite(filename, image)) {
```
vrabaud, line 90: dangerous in case the cb thread write to g_last_image: you probably want to protect all that with a mutex (same with imshow).

image_view/src/nodes/image_view.cpp
```diff
@@ -42,13 +97,40 @@ int main(int argc, char **argv)
...
+  std::string window_name;
+  local_nh.param("window_name", window_name, topic);
+
+  std::string format_string;
+  local_nh.param("filename_format", format_string, std::string("frame%04i.jpg"));
+  g_filename_format.parse(format_string);
+
+  bool autosize;
+  local_nh.param("autosize", autosize, false);
+  cv::namedWindow(window_name, autosize ? cv::WND_PROP_AUTOSIZE : 0);
```
vrabaud: also use keep_ratio and maybe gui_expanded ? http://docs.opencv.org/2.4/modules/highgui/doc/user_interface.html#namedwindow

vrabaud:
That black window is just the default because there is no image yet. How long does it stay there ?

image_view/src/nodes/image_view.cpp
```diff
@@ -32,7 +32,62 @@
...
 *********************************************************************/
 #include <ros/ros.h>
-#include <nodelet/loader.h>
+#include <image_transport/image_transport.h>
+
+#include <cv_bridge/cv_bridge.h>
+#include <opencv2/highgui/highgui.hpp>
+
+#include <boost/format.hpp>
+#include <boost/thread.hpp>
```
vrabaud, line 41: not needed

wkentaro:
> That black window is just the default because there is no image yet. How long does it stay there ?

Actually the window is gray if there is no image.
It takes a few ~~minutes~~ seconds.

image_view/src/nodes/image_view.cpp
wkentaro, line 41: Is is needed for boost::mutex.

wkentaro:
I updated, but still have black images.

vrabaud:
For the synchronization, you might want to use a condition variable: it will simplify things. An example here: http://stackoverflow.com/questions/16907072/how-do-i-use-a-boost-condition-variable-to-wait-for-a-thread-to-complete-process
For the grey window, try a cv::waitKey(10) after the imshow (this will also help have a refresh rate of 100Hz at max and therefore use less CPU in case we send images at 1000Hz).

wkentaro:
which below change, I found main and callback function is in same thread without using multiple thread spinner, so no need to run `imshow` in main but in callback.

```
diff --git a/image_view/src/nodes/image_view.cpp b/image_view/src/nodes/image_view.cpp
index 51ae517..64ce03b 100644
--- a/image_view/src/nodes/image_view.cpp
+++ b/image_view/src/nodes/image_view.cpp
@@ -48,6 +48,7 @@ std::string g_window_name;

 void imageCb(const sensor_msgs::ImageConstPtr& msg)
 {
+  ROS_INFO_STREAM("imageCb: " << boost::this_thread::get_id());
   // We want to scale floating point images so that they display nicely
   bool do_dynamic_scaling = (msg->encoding.find("F") != std::string::npos);

@@ -69,6 +70,7 @@ void imageCb(const sensor_msgs::ImageConstPtr& msg)

 static void mouseCb(int event, int x, int y, int flags, void* param)
 {
+  ROS_INFO_STREAM("mouseCb: " << boost::this_thread::get_id());
   if (event == cv::EVENT_LBUTTONDOWN) {
     ROS_WARN_ONCE("Left-clicking no longer saves images. Right-click instead.");
     return;
@@ -128,6 +130,7 @@ int main(int argc, char **argv)
   image_transport::TransportHints hints(transport, ros::TransportHints(), local_nh);
   image_transport::Subscriber sub = it.subscribe(topic, 1, imageCb, hints);

+  ROS_INFO_STREAM("main: " << boost::this_thread::get_id());
   ros::spin();

   cv::destroyWindow(g_window_name);
```

wkentaro:
I have done. Please review.

vrabaud:
right, the default spinner is mono threaded. Cool ! So you don't have the gray image problem anymore ?

wkentaro:
no, I don't. It works as I expect.

wkentaro:
Could you please review and merge this?

vrabaud:
can you please rebase ? I am trying to see if travis works. Thx.

wkentaro:
I rebased.

vrabaud:
Yes ! Thx, Travis validated. Thx a bunch, this will close a few bugs.

PR #169, opened 2015-12-24
[image_view] Scale 16UC1 depth image

wkentaro:
In most cases, it is depth image, isn't it?


vrabaud:
Well yes. Then again, as a user I'd rather have the choice whether I want dynamic scaling or not: if you have a static scene, the colors might change because of the noise, and this can be weird to watch.
rqt_image_view is good for that. We simply need dynamic reconfigure here.

wkentaro:
image

image_view_dynamic_scaling_before

image_view_dynamic_scaling

PR #172, opened 2016-01-04
remove libopencv-dev and find_package(OpenCV)

k-okada:
when we install ros-indigo-image-proc and try to compile cv_bridge against opencv3, image_proc in /opt/ros/indigo try to link cv_bridge in catkin_ws and that is linked against opencv3.

in this PR, we changed
- remove libopencv-dev from package.xml, depending on cv_bridge is ok
- if we depend on cv_bridge, we do not have to `find_package(OpenCV)` in each CMakeLists.txt
- add `SET(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)` , so /opt/ros/indigo/lib/libimage_proc.so uses cv_bridge in /opt/ros/indigo, not in catkin_ws, this may need discussion.

this pr requires https://github.com/ros-perception/vision_opencv/pull/105

### DIFF
--- a/camera_calibration/package.xml
+++ b/camera_calibration/package.xml
@@ -15,7 +15,6 @@
 
   <run_depend>cv_bridge</run_depend>
   <run_depend>image_geometry</run_depend>
-  <run_depend>libopencv-dev</run_depend>
   <run_depend>message_filters</run_depend>
   <run_depend>rospy</run_depend>
   <run_depend>std_srvs</run_depend>
--- a/depth_image_proc/CMakeLists.txt
+++ b/depth_image_proc/CMakeLists.txt
@@ -9,12 +9,14 @@ catkin_package(
 
 find_package(Boost REQUIRED)
 find_package(Eigen REQUIRED)
-find_package(OpenCV)
 include_directories(include
         SYSTEM ${BOOST_INCLUDE_DIRS}
                ${catkin_INCLUDE_DIRS}
                ${EIGEN_INCLUDE_DIRS}
 )
+
+# set rpath
+SET(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
 
 add_library(${PROJECT_NAME} src/nodelets/convert_metric.cpp
                              src/nodelets/disparity.cpp
--- a/image_proc/CMakeLists.txt
+++ b/image_proc/CMakeLists.txt
@@ -4,7 +4,6 @@ project(image_proc)
 find_package(catkin REQUIRED)
 
 find_package(catkin REQUIRED cv_bridge dynamic_reconfigure image_geometry image_transport nodelet roscpp sensor_msgs)
-find_package(OpenCV REQUIRED)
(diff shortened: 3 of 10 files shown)


trainman419:
This issue is an architectural feature/problem with newer versions of ROS, and affects all ROS packages which provide C/C++ libraries. (You'll run into similar issues if your try to build roscpp in a workspace that is overlaid on /opt/ros)

rosbuild used RPATH to guarantee link consistency, but RPATH is not allowed according to the debian packaging guidelines, so the use of RPATH was removed and replaced with the use of the LD_LIBRARY_PATH environment variable.

Modifying core ROS libraries to RPATH will negatively affect their ability to be accepted into mainline Debian or Ubuntu in the future, so it probably isn't a good idea. In addition, setting the cmake flag for RPATH globally may affect other packages in the workspace, which is generally not a good idea.

trainman419:
Many of these packages depend directly on OpenCV by including the OpenCV headers and linking directly against the OpenCV libraries. Removing the explicit dependency on OpenCV is not a good idea; it means that these packages depend on a transient dependency, which is brittle.

trainman419:
The standard ROS practice for building a package from source is to also build all downstream packages from source in the same workspace.

vrabaud:
@trainman419 pointed at a lot of very good reasons not to accept this PR (thx !).
Now, how can we fix your problem @k-okada : you will (unfortunately) always have some OpenCV around and rosdep only allows for one version of a lib. Also, REP 003 defined only one version of OpenCV for Indigo and Jade. I packaged OpenCV3 as a "bonus".
Unfortunately, the only solution I see is to recompile everything that depends on OpenCV. I know, that can be big but the low-level libs (image_transport, image_proc) were really cleaned from useless dependencies. I am trying to get those compiling with OpenCV3 as it goes too.
I am pushing for OpenCV3 in Kinetic by default. Please voice up ! https://github.com/ros-infrastructure/rep/pull/106
Letting it open if somebody has a better suggestion.

k-okada:
Thanks for comments,
1) I thought using RPATH in ROS system was good feature and realized that was removed at some point, so your explanation really help me to understand situation,
2) as for removing explicit dependency to OpenCV, I think this will help you, specially when you want to switch from upstream opencv package to opencv3 ros package. so how about removing `libopencv-dev` and `find_package(OpenCV)` from package.xml and CMakeLists.txt. If you `find_package(cv_bridge)` then I think we still can use `OpenCV_INCLUDE_DIRS`;

One possibility is to provide ros-indigo-cv-bridge2 and ros-indigo-cv-bridge3 and set conflict each other.

vrabaud:
2) makes sense: especially if OpenCV3 becomes the default in Kinetic, packages will most likely rely on cv_bridge to not maintain several branches.

To get ros-indigo-cv-bridge2 and ros-indigo-cv-bridge3 conflict sure, but then we would need image_pipeline, image_common, image_transport and many more to have both versions. And it's an all or nothing situation.

PR #174, opened 2016-01-10
[image_view] Do dynamic scaling for not 8UC1 image

wkentaro:


vrabaud:
I don't see why it should be the default: there should be an option too. We're starting to have hand-picked rules in Python too, that's bad. Add cvColorForDisplay to the bridge and use it, it will make things simpler no ?

wkentaro:
Yes, you're right.
Converting to 8UC1 is necessary to save image as jpg file with OpenCV.
Adding cvtColorForDisplay in cv_bridge for python is better, so I will do so later.

vrabaud:
As hints for you:
https://github.com/ros-perception/vision_opencv/blob/indigo/cv_bridge/src/module_opencv2.cpp
would be the place to add something like:
https://github.com/ros-perception/image_common/pull/44

wkentaro:
Updated the commit.

frame0000_1

PR #181, opened 2016-03-02
when no image is saved, do not save camera info

jkerfs:
When the images are not recorded because "save_all_image" is false and "save_image_service" is false, the frame count should not be incremented and the camera info should not be written to disk.


vrabaud:
makes sense, thx.

Issue #180, opened 2016-02-29
Need to check if all distortion coefficients are zero, not just first one


lucasw:
https://github.com/ros-perception/image_pipeline/blob/indigo/image_proc/src/nodelets/rectify.cpp#L125

```
// If zero distortion, just pass the message along
if (info_msg->D.empty() || info_msg->D[0] == 0.0)
{
...
```

That check ought to test every element of info_msg->D to see if they are all 0.0.  This check probably always works with real camera calibrations, but synthetic ones (which could be used in tests or simulations) can have an exactly zero first coefficient and non-zero other coefficients.

lucasw:
I can provide a pull request in a day or two.

vrabaud:
Thx, please do. And please include a test.

lucasw:
A corresponding change is needed in vision_opencv to make a fix here work:

https://github.com/ros-perception/vision_opencv/blob/indigo/image_geometry/src/pinhole_camera_model.cpp#L130

```
cache_->distortion_state = (cam_info_.D.size() == 0 || (cam_info_.D[0] == 0.0)) ? NONE : CALIBRATED;
```

vrabaud:
greatly fixed in #117

PR #182, opened 2016-03-06
Check if all distortion coefficients are zero.

lucasw:
For issue #180

Test with:

rostest --reuse-master --text image_proc test_rectify.xml

Can also test interactively with vimjay image_rect.launch, which brings up an rqt gui and camera info distortion coefficients can be dynamically reconfigured.

### DIFF
--- a/image_proc/src/nodelets/rectify.cpp
+++ b/image_proc/src/nodelets/rectify.cpp
@@ -122,7 +122,17 @@ void RectifyNodelet::imageCb(const sensor_msgs::ImageConstPtr& image_msg,
   }
 
   // If zero distortion, just pass the message along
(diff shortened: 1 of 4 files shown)


vrabaud:
tests do not pass on Travis: https://travis-ci.org/ros-perception/image_pipeline/jobs/114052915
If you keep pushing commits, they will be tested automatically.

lucasw:
If this is using the off the shelf vision_opencv and not what what merged for https://github.com/ros-perception/vision_opencv/pull/118 this will keep failing until that dependency gets updated.

https://travis-ci.org/ros-perception/image_pipeline/jobs/114052919#L1354

```
executing command [sudo -H apt-get install -y ros-jade-image-geometry]
```

vrabaud:
ok thx, I should probably release that.

PR #183, opened 2016-03-08
image_view: ImageView.cfg: Do not try to load image_view python packa…

aballier:
…ge as this is not needed and would fail if it is not installed.

### DIFF
--- a/image_view/cfg/ImageView.cfg
+++ b/image_view/cfg/ImageView.cfg
@@ -2,8 +2,6 @@
 
 PACKAGE='image_view'
 try:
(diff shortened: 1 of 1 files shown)


vrabaud:
That got fixed in 3ad75a15430ce46b25341233665bcce40e3afa70 , does that work for you ?

aballier:
yes, thanks

i was thinking that maybe https://github.com/ros/dynamic_reconfigure/commit/82b104cad8f309cadaa22f0badf64815feaad88c would help too

PR #184, opened 2016-03-09
image_view freeze fix for OpenCV 3 installed from Source

iKrishneel:
image_view freezes when opencv is installed from source. 
Before fix:
screenshot from 2016-03-09 15 28 33

After fix:
screenshot from 2016-03-09 15 32 43


vrabaud:
actually, can you please make it a `cv::waitKey(5);` and remove the `#if`, there should be one for all release. Oh, and can you also please squash your commits ? Thx

iKrishneel:
Done.

vrabaud:
3 is a bit low, that's why I mentioned 5. But let's try with that.

vrabaud:
thx !

PR #192, opened 2016-04-19
[camera_calbration] fix typo np -> numpy

knorth55:
`np` is note declared. Standardize into `np` -> `numpy`

### DIFF
--- a/camera_calibration/src/camera_calibration/calibrator.py
+++ b/camera_calibration/src/camera_calibration/calibrator.py
@@ -248,9 +248,9 @@ class Calibrator(object):
             _, max_val, _, _ = cv2.minMaxLoc(img)
             if max_val > 0:
                 scale = 255.0 / max_val
-                mono_img = (img * scale).astype(np.uint8)
+                mono_img = (img * scale).astype(numpy.uint8)
             else:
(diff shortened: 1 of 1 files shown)


vrabaud:
I'll also write a test. This error has always been here as far as I can tell no ? Or did it just appear recently.
Thx for the fix !

knorth55:
I roughly searched its change history, but it is not a recent change.
I found this when I tried to calibrate camera with depth image, and it does occurred for color image calibration, so most people don't care about this.

knorth55:
@vrabaud travis fails because of short storage.

```
The log length has exceeded the limit of 4 MB (this usually means that the test suite is raising the same exception over and over).
```

knorth55:
@vrabaud ping
on indigo branches,  it passes. but on hydro branches, it doesn't.
i think this error is because of opencv setting

```
libdc1394 error: Failed to initialize libdc1394
```

vrabaud:
Travis was failing in the first place because it needed an update from cv_bridge (which just got synced but only for Indigo).
Then there was the problem that you mentioned as well as other ones on Travis.
Indigo is fine for now at least so I'll merge. Thx !

PR #198, opened 2016-06-02
changed file numbering from %04i to %06i due to the need of saving mo…

zwenga:
Changed file numbering from %04i to %06i due to the need of saving more than 10000 images. There is no possibility known to me to change the numbering from outside, so the numbering should be at least as high as %06i.

### DIFF
--- a/camera_calibration/src/camera_calibration/calibrator.py
+++ b/camera_calibration/src/camera_calibration/calibrator.py
@@ -501,7 +501,7 @@ class Calibrator(object):
 
 def image_from_archive(archive, name):
     """
-    Load image PGM file from tar archive. 
+    Load image PGM file from tar archive.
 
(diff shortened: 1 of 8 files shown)


wkentaro:
I have been thinking about the filename rules of image saving in this repo,
and in my opinion the node should support image saving with timestamp in filename.
What do you think about this, there is no duplication in filename in this case?

wkentaro:
I found the solution for that in `video_recorder.cpp` : https://github.com/ros-perception/image_pipeline/pull/203
This issue seems fixed with same approach.

JWhitleyWork:
Is this still an issue? I know it's been a long time but we're trying to clear the backlog.

JWhitleyWork:
Closing stale issue.

PR #206, opened 2016-07-08
[image_view] Refactor: Remove subscription of camera_info in video_recorder

wkentaro:
It is not necessary.

### DIFF
--- a/image_view/src/nodes/video_recorder.cpp
+++ b/image_view/src/nodes/video_recorder.cpp
@@ -39,12 +39,11 @@ double max_depth_range = 5.5;
 bool use_dynamic_range = false;
 
 
-void callback(const sensor_msgs::ImageConstPtr& image_msg, const sensor_msgs::CameraInfoConstPtr& info)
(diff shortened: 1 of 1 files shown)


image_view/src/nodes/video_recorder.cpp
```diff
@@ -107,9 +105,7 @@ int main(int argc, char** argv)
 
     image_transport::ImageTransport it(nh);
     std::string topic = nh.resolveName("image");
-    image_transport::CameraSubscriber sub_camera = it.subscribeCamera(topic, 1, &callback);
-    image_transport::Subscriber sub_image = it.subscribe(topic, 1,
-            boost::bind(callback, _1, sensor_msgs::CameraInfoConstPtr()));
+    ros::Subscriber sub_image = nh.subscribe(topic, 1, &callback);
```
vrabaud: you still want to use an image_transport::Subscriber so that you can view compressed images. Agree on removing CameraInfo though.

PR #214, opened 2016-07-15
Remove testing with HEAD version of cv_bridge that is released on apt

wkentaro:


vrabaud:
Thx for tracking all that !

PR #218, opened 2016-08-10
launch image view with a predefined window size

wiseware-ros:
We came across a use case where we want to launch an image_view window with a predefined window size. The solution checks if autoscale is not set and introduces two new ROS parameters (width and height). If both are present the window is resized. This way, previous functionality is maintained and the window is only resized if you specifically tell the node to at launch time.

### DIFF
--- a/image_view/src/nodes/image_view.cpp
+++ b/image_view/src/nodes/image_view.cpp
@@ -135,6 +135,18 @@ int main(int argc, char **argv)
   cv::namedWindow(g_window_name, autosize ? (CV_WINDOW_AUTOSIZE | CV_WINDOW_KEEPRATIO | CV_GUI_EXPANDED) : 0);
   cv::setMouseCallback(g_window_name, &mouseCb);
 
+  if(autosize == false)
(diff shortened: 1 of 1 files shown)


image_view/src/nodes/image_view.cpp
```diff
@@ -135,6 +135,18 @@ int main(int argc, char **argv)
...
+	  if(local_nh.hasParam("width") && local_nh.hasParam("height"))
+	  {
+		  int width;
+		  local_nh.getParam("width", width);
+		  int height;
+		  local_nh.getParam("height", height);
+		  cv::resizeWindow(g_window_name, width, height);
+	  }
+  }
+
```
wkentaro, line 149: It should be 2 space, I think.

vrabaud: yes, please don't use tabs but spaces.

vrabaud (changes requested):

vrabaud:
Please update to spaces.

PR #219, opened 2016-08-12
Include missing headers

mryellow:
Specified headers for `sensor_msgs/image_encodings` and `opencv2/imgproc` in `crop_non_zero.cpp`.


mryellow:
Was using an older version of `cv_bridge`, updating to latest source fixed this and other build errors.

PR #223, opened 2016-09-28
enable to set camera_info_url from rosparam, enable to publish images of depth other than bgr8

k-okada:
@wkentaro , I think this PR enables what you want to do in https://github.com/ros-perception/image_common/pull/54#issuecomment-248081140

### DIFF
--- a/image_publisher/src/nodelet/image_publisher_nodelet.cpp
+++ b/image_publisher/src/nodelet/image_publisher_nodelet.cpp
@@ -100,7 +100,22 @@ class ImagePublisherNodelet : public nodelet::Nodelet
       if (flip_image_)
         cv::flip(image_, image_, flip_value_);
 
-      sensor_msgs::ImagePtr out_img = cv_bridge::CvImage(std_msgs::Header(), "bgr8", image_).toImageMsg();
+      std::string encoding = sensor_msgs::image_encodings::BGR8;
+      switch ( image_.type() ) {
+        case CV_8UC3:
+          encoding = sensor_msgs::image_encodings::BGR8; break;
+        case CV_8UC1:
+          encoding = sensor_msgs::image_encodings::MONO8; break;
+        case CV_16UC1:
+          encoding = sensor_msgs::image_encodings::MONO16; break;
+        case CV_16UC3:
+          encoding = sensor_msgs::image_encodings::BGR16; break;
+        case CV_8UC4:
+          encoding = sensor_msgs::image_encodings::BGRA8; break;
+        case CV_16UC4:
+          encoding = sensor_msgs::image_encodings::BGRA16; break;
+      }
+      sensor_msgs::ImagePtr out_img = cv_bridge::CvImage(std_msgs::Header(), encoding, image_).toImageMsg();
       out_img->header.frame_id = frame_id_;
       out_img->header.stamp = ros::Time::now();
       camera_info_.header.frame_id = out_img->header.frame_id;
@@ -124,6 +139,23 @@ class ImagePublisherNodelet : public nodelet::Nodelet
     subscriber_count_--;
   }
 
+  std::string depth2str(int depth) {
+    std::string r;
+
+    switch ( depth ) {
+      case CV_8U:  r = "8U"; break;
+      case CV_8S:  r = "8S"; break;
+      case CV_16U: r = "16U"; break;
+      case CV_16S: r = "16S"; break;
+      case CV_32S: r = "32S"; break;
+      case CV_32F: r = "32F"; break;
+      case CV_64F: r = "64F"; break;
+      default:     r = "User"; break;
+    }
+
+    return r;
+  }
+
 public:
   virtual void onInit()
   {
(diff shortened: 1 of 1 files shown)


wkentaro:
This would be a nice feature, but there is still needs to publish pcd and camera_info with synchronization,
which should be added to `pcd_to_pointcloud` as a feature.
Also, in some situation I prefer image and points are synchronized, in that case something like a virtual_camera is required.

wkentaro:
Also, we need saving node for camera_info, image, depth_image and pcd with synchronization, and we need to create a node to handle all of them.
I'm planning to place that node in `virtual_camera` package, and I think it helps users (including me) to find and use the node in the same package.

vrabaud:
lgtm. @wkentaro, can I merge ?

image_publisher/src/nodelet/image_publisher_nodelet.cpp
```diff
@@ -174,14 +207,27 @@ class ImagePublisherNodelet : public nodelet::Nodelet
...
-    camera_info_.K = list_of(1)(0)(camera_info_.width/2)(0)(1)(camera_info_.height/2)(0)(0)(1);
-    camera_info_.R = list_of(1)(0)(0)(0)(1)(0)(0)(0)(1);
-    camera_info_.P = list_of(1)(0)(camera_info_.width/2)(0)(0)(1)(camera_info_.height/2)(0)(0)(0)(1)(0);
-
+    std::string camera_info_url_;
+    nh_.param("camera_info_url", camera_info_url_);
+    if ( !camera_info_url_.empty() ) {
+      camera_info_manager::CameraInfoManager c(nh_);
+      try {
+        nh_.param("camera_info_url", camera_info_url_);
```
wkentaro: Why duplication of `nh_.param("camera_info_url", camera_info_url_);`?

image_publisher/src/nodelet/image_publisher_nodelet.cpp
```diff
@@ -174,14 +207,27 @@ class ImagePublisherNodelet : public nodelet::Nodelet
...
 
-    camera_info_.width = image_.cols;
-    camera_info_.height = image_.rows;
-    camera_info_.distortion_model = "plumb_bob";
-    camera_info_.D = list_of(0)(0)(0)(0)(0);
-    camera_info_.K = list_of(1)(0)(camera_info_.width/2)(0)(1)(camera_info_.height/2)(0)(0)(1);
-    camera_info_.R = list_of(1)(0)(0)(0)(1)(0)(0)(0)(1);
-    camera_info_.P = list_of(1)(0)(camera_info_.width/2)(0)(0)(1)(camera_info_.height/2)(0)(0)(0)(1)(0);
-
+    std::string camera_info_url_;
```
wkentaro: Why do you add `_` suffix to variable that is not the data field of the class?

wkentaro (changes requested):

image_publisher/src/nodelet/image_publisher_nodelet.cpp
k-okada: removed

image_publisher/src/nodelet/image_publisher_nodelet.cpp
k-okada: that't mistake, removed

wkentaro:
LGTM

image_publisher/src/nodelet/image_publisher_nodelet.cpp
```diff
@@ -100,7 +100,22 @@ class ImagePublisherNodelet : public nodelet::Nodelet
...
+        case CV_8UC3:
+          encoding = sensor_msgs::image_encodings::BGR8; break;
+        case CV_8UC1:
+          encoding = sensor_msgs::image_encodings::MONO8; break;
+        case CV_16UC1:
+          encoding = sensor_msgs::image_encodings::MONO16; break;
+        case CV_16UC3:
+          encoding = sensor_msgs::image_encodings::BGR16; break;
+        case CV_8UC4:
+          encoding = sensor_msgs::image_encodings::BGRA16; break;
```
vrabaud: should be BGRA8

image_publisher/src/nodelet/image_publisher_nodelet.cpp
```diff
@@ -100,7 +100,22 @@ class ImagePublisherNodelet : public nodelet::Nodelet
       if (flip_image_)
         cv::flip(image_, image_, flip_value_);
 
-      sensor_msgs::ImagePtr out_img = cv_bridge::CvImage(std_msgs::Header(), "bgr8", image_).toImageMsg();
+      std::string encoding = "bgr8";
```
vrabaud: instead of "bgr8", it should be sensor_msgs::image_encodings::BGR8 (yes, it was wrong before).

vrabaud:
BTW, the title of this PR is "enable to publish depth image", I guess you mean "enable to publish images of depth other than bgr8", right ?

vrabaud:
Please address the latest fixes.

k-okada:
@vrabaud I'm very sorry for super late response, fixed the code.

JWhitleyWork:
@k-okada - Is this still a problem and, if so, could you rebase on Melodic?

JWhitleyWork:
Closing based on lack of response.

PR #224, opened 2016-09-28
Convert depth image metric from [m] to [mm]

wkentaro:


depth_image_proc/src/nodelets/convert_metric.cpp
```diff
@@ -63,6 +67,17 @@ void ConvertMetricNodelet::onInit()
   ros::NodeHandle& nh = getNodeHandle();
   it_.reset(new image_transport::ImageTransport(nh));
 
+  // Get rosparams and validate them
+  ros::NodeHandle& pnh = getPrivateNodeHandle();
+  pnh.param("encoding_from", encoding_from_, enc::TYPE_16UC1);
+  pnh.param("encoding_to", encoding_to_, enc::TYPE_32FC1);
+  if ((encoding_from_ == enc::TYPE_16UC1 && encoding_to_ == enc::TYPE_32FC1) ||
```
vrabaud: if(!(    no ?

wkentaro: You're right. Thanks.

wkentaro:
Maybe is that better if the encoding from-to is estimated from input message?

depth_image_proc/src/nodelets/convert_metric.cpp
```diff
@@ -144,9 +123,10 @@ void ConvertMetricNodelet::depthCb(const sensor_msgs::ImageConstPtr& raw_msg)
   }
   else
   {
-    ROS_ERROR("Unsupported image conversion. from: %s to: %s.", encoding_from_.c_str(), encoding_to_.c_str());
+    ROS_ERROR("Unsupported image conversion from %s.", raw_msg->encoding);
     return;
   }
+  depth_msg->step     = raw_msg->width * (enc::bitDepth(depth_msg->encoding) / 8);
```
tmoinel: you want to compute this before line 95 :
depth_msg->data.resize( depth_msg->height \* depth_msg->step);
otherwise the length of data is 0 and it seg fault when you loop to fill it.

A possible enhancement would be change queue size with a parameter.
Apart from that it worked perfectly for me :)

wkentaro: Thanks. I fixed it.

vrabaud:
LGTM, can you please squash your commits? Thx.

wkentaro:
You can squash them with GitHub GUI. I don't mind if they will become a single commit.

vrabaud:
thx, I did not know the GUI could do that :)

Issue #225, opened 2016-10-04
[camera_calibration] OpenCV Error: Assertion failed (ni > 0 && ni == ni1) 


MartinPeris:
When calibrating a stereo camera with multiple calibration patterns, the following error occurs under some circumstances:

```
OpenCV Error: Assertion failed (ni > 0 && ni == ni1) in collectCalibrationData, file /build/buildd/opencv-2.4.8+dfsg1/modules/calib3d/src/calibration.cpp, line 3193
Traceback (most recent call last):
  File "/opt/ros/indigo/lib/camera_calibration/cameracalibrator.py", line 244, in on_mouse
    self.c.do_calibration()
  File "/opt/ros/indigo/lib/python2.7/dist-packages/camera_calibration/calibrator.py", line 1093, in do_calibration
    self.cal_fromcorners(self.good_corners)
  File "/opt/ros/indigo/lib/python2.7/dist-packages/camera_calibration/calibrator.py", line 865, in cal_fromcorners
    self.r.cal_fromcorners(rcorners)
  File "/opt/ros/indigo/lib/python2.7/dist-packages/camera_calibration/calibrator.py", line 603, in cal_fromcorners
    flags = self.calib_flags)
cv2.error: /build/buildd/opencv-2.4.8+dfsg1/modules/calib3d/src/calibration.cpp:3193: error: (-215) ni > 0 && ni == ni1 in function collectCalibrationData


```

Steps to reproduce the error:
- Print two calibration chessboards: 9x6 and 8x6 (in my case the square size for 9x6 was 0.0315m and for 8x6 was 0.021m)
- Start your stereo camera (with whatever rosrun/roslaunch command appropriate for your camera)
- Start the calibration node: 

> rosrun camera_calibration cameracalibrator.py --size=9x6 --square 0.0315 --size=8x6 --square 0.021
- Swing the 9x6 calibration chessboard around until the "Calibrate" button is activated.
- Move the 9x6 calibration chessboard towards the edge of one of the images until you get something similar to the following screenshot (notice that on the left image the chessboard is detected as 9x6 but on the right image the chessboard is detected as 8x6):

![bad_chessboard_detection](https://cloud.githubusercontent.com/assets/5905362/19075424/cd228bb0-8a7f-11e6-8e66-53ca188abeda.jpg)
- Click the "Calibrate" button and enjoy your "Assertion failed" error.

The stereo calibration node is not checking if the calibration pattern detected in left and right images are the same.

PR #226, opened 2016-10-04
During stereo calibration check that the number of corners detected i…

MartinPeris:
…n the left and right images are the same. This fixes ros-perception/image_pipeline#225


vrabaud:
Thx ! A test would be nice too :)
That was pretty bad, I am surprised it worked decently before.

PR #231, opened 2016-10-31
Added an argument to the calibration script for user defined folder location.

v-mehta:
This PR enables the user of camera _calibration to save the calculated parameters to custom location using an argument. By default it goes to /tmp folder. 

The editor updated the files with indentation changes according to https://www.python.org/dev/peps/pep-0008/. Please let me know if you want them to be reverted to avoid confusion.

### DIFF
--- a/camera_calibration/nodes/cameracalibrator.py
+++ b/camera_calibration/nodes/cameracalibrator.py
@@ -43,6 +43,7 @@ from collections import deque
 import threading
 import functools
 import time
+import tempfile
 
 import cv2
 import numpy
@@ -95,7 +96,7 @@ class ConsumerThread(threading.Thread):
 
 class CalibrationNode:
     def __init__(self, boards, service_check = True, synchronizer = message_filters.TimeSynchronizer, flags = 0,
-                 pattern=Patterns.Chessboard, camera_name='', folder_location='/tmp/', checkerboard_flags = 0):
+                 pattern=Patterns.Chessboard, camera_name='', folder_location=tempfile.gettempdir(), checkerboard_flags = 0):
         if service_check:
             # assume any non-default service names have been set.  Wait for the service to become ready
             for svcname in ["camera", "left_camera", "right_camera"]:
@@ -279,9 +280,9 @@ class OpenCVCalibrationNode(CalibrationNode):
 
     def screendump(self, im):
         i = 0
-        while os.access("/tmp/dump%d.png" % i, os.R_OK):
+        while os.access(tempfile.gettempdir()+"/dump%d.png" % i, os.R_OK):
             i += 1
-        cv2.imwrite("/tmp/dump%d.png" % i, im)
+        cv2.imwrite(tempfile.gettempdir()+"/dump%d.png" % i, im)
 
     def redraw_monocular(self, drawable):
         height = drawable.scrib.shape[0]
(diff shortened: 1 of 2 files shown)


camera_calibration/nodes/cameracalibrator.py
```diff
@@ -95,7 +95,7 @@ def run(self):
 
 class CalibrationNode:
     def __init__(self, boards, service_check = True, synchronizer = message_filters.TimeSynchronizer, flags = 0,
-                 pattern=Patterns.Chessboard, camera_name='', checkerboard_flags = 0):
+                 pattern=Patterns.Chessboard, camera_name='', folder_location='/tmp/', checkerboard_flags = 0):
```
vrabaud: please use the tempfile module (and gettempdir()) so that it can work on multiple OSes (and yes, the current implementation is definitely not working on other OSes :) ).

camera_calibration/nodes/cameracalibrator.py
```diff
@@ -274,7 +276,7 @@ def buttons(self, display):
     def y(self, i):
         """Set up right-size images"""
         return 30 + 40 * i
-        
+
     def screendump(self, im):
         i = 0
         while os.access("/tmp/dump%d.png" % i, os.R_OK):
```
vrabaud: how about the tmp here ? (not sure ... have you checked other mentions of the tmp path?)

camera_calibration/nodes/cameracalibrator.py
```diff
@@ -362,6 +364,9 @@ def main():
     parser.add_option("-c", "--camera_name",
                      type="string", default='narrow_stereo',
                      help="name of the camera to appear in the calibration file")
+    parser.add_option("-f", "--folder_location",
+                     type="string", default='/tmp/',
```
vrabaud: indentation is off by 1.

camera_calibration/src/camera_calibration/calibrator.py
```diff
@@ -202,7 +202,8 @@ class Calibrator(object):
     """
     Base class for calibration system
     """
-    def __init__(self, boards, flags=0, pattern=Patterns.Chessboard, name='', checkerboard_flags=cv2.CALIB_CB_FAST_CHECK):
+    def __init__(self, boards, flags=0, pattern=Patterns.Chessboard, name='',
+                folder_location='', checkerboard_flags=cv2.CALIB_CB_FAST_CHECK):
```
vrabaud: indentation is off by 1

vrabaud (commented):
Very useful overall. For spaces, I am all for pep8-ifying code but next time, please make a separate commit, it makes things easier to review (it could be part of another PR). It's ok for this one. Thx.

camera_calibration/nodes/cameracalibrator.py
v-mehta: Good idea. Will update it in a bit.

camera_calibration/nodes/cameracalibrator.py
v-mehta: I will update all the tmp path mentions. Looks like I missed a few.

vrabaud:
please remove your last commit and make a pull request.
BW, how is your last commit (about yaml) different from the lryaml function of the yaml() that those classes already have ?

vrabaud:
I still don't understand the use of do_yaml_save: do_tarfile_save already creates the .yaml file. Plus do_yaml_save does not have a similar API as do_tarfile_save (one takes a file object, the other one a path)

PR #233, opened 2016-11-03
Remove rosdoc.yaml as the external docs URL is no longer valid.

mikepurvis:


vrabaud:
Hi, I fixed the doc jobs instead, thx for noticing !

PR #236, opened 2016-11-18
Fix #235 for kinetic

### LINKED ISSUE #235: Can't save image through image_view on Kinetic
On kinetic, when use mouse to right click image_view screen, it was failed to save the image.
And also image_view screen will be hang, even can't terminate it by press ctrl+c on terminal

k-okada:
This is workaround for #235, my guess is that it locks `g_image_mutex` within `mouseCb`, which I think called from `waitKey` function in `imageCb` that also locks `g_imgae_mutex`, so that it tryied to lock mutex whcih is already locked within same thread.

on Kinetic, Right click seems to assigned to other function as the picture, so I think it's better to us Left click for this purpose.

screenshot from 2016-11-18 15-43-03

### DIFF
--- a/image_view/src/nodes/image_view.cpp
+++ b/image_view/src/nodes/image_view.cpp
@@ -90,8 +90,6 @@ static void mouseCb(int event, int x, int y, int flags, void* param)
     return;
   }
 
-  boost::mutex::scoped_lock lock(g_image_mutex);
-
   const cv::Mat &image = g_last_image;
 
   if (image.empty()) {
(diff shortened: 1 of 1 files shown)


wangji1:
Is this change only for Kinetic, or both for Kinetic and Indigo?

wangji1:
When should I test the new changes?

k-okada:
- I think this works for both indigo/kinetic, but this may change the UI behavior so I suppose this will only released to kinetic, 
- you can merge this branch into your source tree and check if this works for you too.

JWhitleyWork:
@k-okada Is this still relevant? If so, can you rebase on `melodic`?

JWhitleyWork:
Closing for now due to lack of response.

PR #240, opened 2016-12-14
Apply value scaling to depth/float image with specified/default min/max range

wkentaro:
This is to visualize depth images without `_do_dynamic_scaling:=true` rosparam specification.

- 32FC1: we assume meter metric, and 10[m] as the max range.
- 16UC1: we assume milimeter metric, and 10 * 1000[mm] as the max range.

### DIFF
--- a/image_view/cfg/ImageView.cfg
+++ b/image_view/cfg/ImageView.cfg
@@ -22,6 +22,8 @@ edit_method_colormap = gen.enum([
 ], "colormap")
 
 gen.add('do_dynamic_scaling', bool_t, 0, 'Do dynamic scaling about pixel values or not', False)
-gen.add('colormap', int_t, -1, "colormap", -1, -1, 11, edit_method=edit_method_colormap);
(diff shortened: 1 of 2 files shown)


wkentaro:
## dynamic scaling
dynscaling

dynscaling_jet

## non-dynamic scaling

scaling

scaling_jet

wkentaro:
## Update

Sometimes I'd like to visualize probability image whose image value range is from 0 to 1, and max_image_value should be `1` in that case. Previous implementation is not sufficient to achieve this.

I hit another idea with specifying min/max image_value with rosparam to solve this problem.
We can assume the min/max values are not specified if they are equal, in that case we use 10[m] as the default value. And if either min or max image value specified we should use it.

**RGB image**


**not specified**
not_specified

**max=3.0**
max_3

**min=0.5,max=1.0**
min_0 5_max_1

vrabaud:
LGTM

PR #242, opened 2016-12-20
stereo_view: fixed empty left, right, disparity windows with opencv3

### LINKED ISSUE #188: stereo_view showing empty left, right, disparity windows with opencv3
hello, myself and at least the poster of the question at this url: 

http://answers.ros.org/question/225012/image_view-stereo_view-do-not-show-image-or-disparity/?answer=229690#post-id-229690

have experienced stereo view showing empty windows for the left, right and disparity images. 

I  got it to work again by changing some code in the image call back function. My answer to the question at the above url details the solution. 

I was also asked to post to the issue here by the moderator but I've never done this before so i don't know how much more detail is needed sorry!

cwecht:
Fixed #188 (the solution is described there).


vrabaud:
Thx !

PR #244, opened 2017-01-03
DisparityViewNodelet: fixed freeze

cwecht:
A similar problem as in #242 has been fixed in DisparityViewNodelet. OpenCV with QT needs `namedWindow` and  `imshow` to be called in the same thread. DisparityViewNodelet `onInit` and the DisparityImage-callback are not called in the same thread, so it necessary to move the `namedWindow`-call in the callback-function.


leolchat:
Please merge that :)
PS: Just putting a call to waitKey fixes the things for me (opencv3 + qt).

PR #247, opened 2017-01-24
[python3] Use floor division where necessary.

hgaiser:
As per [the docs](https://docs.python.org/3/howto/pyporting.html#division), some divisions need to be floored (as opencv expects integers in these places). This change is python 2 / 3 compatible.

### DIFF
--- a/camera_calibration/nodes/cameracalibrator.py
+++ b/camera_calibration/nodes/cameracalibrator.py
@@ -261,9 +261,9 @@ class OpenCVCalibrationNode(CalibrationNode):
             color = (155, 155, 80)
         else:
             color = (224, 224, 224)
-        cv2.circle(dst, (size[0] / 2, size[1] / 2), min(size) / 2, color, -1)
+        cv2.circle(dst, (size[0] // 2, size[1] // 2), min(size) // 2, color, -1)
         (w, h) = self.getTextSize(label)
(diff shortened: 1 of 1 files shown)


vrabaud:
Thx for //
For BytesIO, would you be able to read something that was written with StringIO ?

hgaiser:
Not sure to be honest. The main issue was that the StringIO was used to write an encoded image (as a sequence of bytes) _and_ the yaml and text files (as a string). In python2, a bytes array or string array are the same thing, which is why StringIO was applicable. In python3 however they are two separate things (try to evaluate `bytes is str` in python2/3). Therefore, StringIO was giving an error when writing the image, since it was expecting a string but received a bytesarray. The solution I chose here was to write the yaml files as bytes too and use a BytesIO instead.

Apologies for not explaining this change earlier.

vrabaud:
Can you just remove that commit for now and make another one with StringIO please ? I want to merge the first commit for now and try the other one.

hgaiser:
Done.

hgaiser:
Any update on this ?

vrabaud:
Thx for cleaning it !

vrabaud:
time for doing one with StringIO :)

hgaiser:
No problem :) thanks for merging.

PR #253, opened 2017-02-18
Fix encoding error message

leolchat:


vrabaud:
Nice, let's be precise :)

PR #254, opened 2017-02-18
Fix and Improve Camera Calibration Checker Node

leolchat:
The current camerachecker node doesn't work because it passes a list of boards when the calibrator.py code expect only one board.

It is also improved by allowing the usage of the approximate argument in the same way as other image pipeline nodes.


vrabaud:
Nice, thx ! Was there any bug open related to that ?

leolchat:
I don't think so, seems like nobody used that node :) Who cares about checking calibration once it is done !

PR #256, opened 2017-03-07
Use BytesIO to save calibration.

hgaiser:
As discussed in https://github.com/ros-perception/image_pipeline/pull/247 :

The main issue was that StringIO was used to write an encoded image (as a sequence of bytes) and the yaml and text files (as a string). In python2, a bytes array or string array are the same thing, which is why StringIO was applicable. In python3 however they are two separate things (try to evaluate `bytes is str` in python2/3). Therefore, StringIO was giving an error when writing the image, since it was expecting a `str` but received a `bytesarray`. The solution I chose here was to write the yaml files as bytes too and use a BytesIO instead.


vrabaud:
Hi, we can't change the format and have text files now be binary files. I believe I fixed it with my commit, please confirm.

PR #257, opened 2017-03-08
launch image view with a predefined window size

k-okada:
rewrited (fixed) vesrion of #218

close #218


octvs:
Hey there,
is this functionality later abandoned? There is no mention on the wiki, although I was hopeful that this commit would address my use-case. I tried to follow up the code with my limited understanding of cpp but couldn't find it in the new version of the file, or in the nodelets that might have received it after the refactoring.
Cheers

Issue #262, opened 2017-04-03
Missing documentation for image_publisher


trainman419:
image_publisher looks like a nice package, but there's nothing on the ROS wiki about it. It would be very nice if there was a bit of documentation.

cc @k-okada since it looks like he was the original author

k-okada:
ok, done -> http://wiki.ros.org/image_publisher, we need https://github.com/ros-perception/image_pipeline/pull/263 to have link from image_pipeline page

vrabaud:
Fixed by #263

PR #263, opened 2017-04-04
Update package.xml to include image_publisher

k-okada:
see https://github.com/ros-perception/image_pipeline/issues/262


vrabaud:
thx !

PR #266, opened 2017-04-10
remove find_package(OpenCV), but to use cv_bridge

k-okada:
if we install ros-indigo-opencv3 and compile this package, the executbale segfaults because cv_bridge is linked against cv2 but this node is linked against cv3


vrabaud:
This would set precedence for a very dangerous path: if you install opencv3, then all should be compiled against it. cv_bridge included.

PR #270, opened 2017-04-14
Remove python-opencv run_depend for image_view

kartikmohta:
The `python-opencv` dependency pulls in the system OpenCV v2.4 which is not required since the `image_view` package depends on `cv_bridge` which pulls in `opencv3` and `opencv3` provides the python library that `image_view` can use.

### DIFF
--- a/image_view/package.xml
+++ b/image_view/package.xml
@@ -18,7 +18,7 @@
   <buildtool_depend>catkin</buildtool_depend>
 
   <test_depend>rostest</test_depend>
-  
+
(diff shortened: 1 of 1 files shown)


kartikmohta:
The `opencv3` comments are only valid for `kinetic` but `cv_bridge` pulls in `python-opencv` for `indigo` anyway, so we can safely remove it here and avoid creating a new branch for `kinetic`.

vrabaud:
great catch !

PR #276, opened 2017-06-05
[image_proc][crop_decimate] support changing target image frame_id

furushchev:
This PR provides new parameter `~target_frame_id` that specifies frame_id name of output image topic.


vrabaud:
Thx !

Issue #285, opened 2017-07-18
tarfile_calibration does not visualize rectified images after calibration


trigal:
in tarfile_calibration.py there is a waitKey define with 6 ms as delay (as reported in [opencv documentation](http://docs.opencv.org/2.4/modules/highgui/doc/user_interface.html#waitkey) . 
This doesn't make sense since waitkey() should wait for a pressed key, then releasing the while inside the display() define. 

```python
def waitkey():
    k = cv2.waitKey(6)
    return k

def display(win_name, img):
    cv2.namedWindow(win_name, cv2.WINDOW_NORMAL)
    cv2.imshow( win_name,  numpy.asarray( img[:,:] ))
    k=-1
    while k ==-1:
        k=waitkey()
    cv2.destroyWindow(win_name)
    if k in [27, ord('q')]:
        rospy.signal_shutdown('Quit')
```

this should be the desired behavior: 
```python
def waitkey():
    k = cv2.waitKey(0)
    return k
```

k-okada:
I have rewrite #286  and post new PR #340, but not sure how this related to the "does not visualize rectified images after calibration" can you provide sample tar file and instructions to how to reproduce this problem?

JWhitleyWork:
Closing stale issue which appears to also be fixed by #340. Can re-open if OP replies or offers feedback.

PR #286, opened 2017-07-18
Update tarfile_calibration.py

trigal:
this commit refs #285  and solve "does not visualize rectified images after calibration" issue

### DIFF
--- a/camera_calibration/scripts/tarfile_calibration.py
+++ b/camera_calibration/scripts/tarfile_calibration.py
@@ -46,7 +46,7 @@ import rospy
 import sensor_msgs.srv
 
 def waitkey():
-    k = cv2.waitKey(6)
+    k = cv2.waitKey(0)
     return k
 
 def display(win_name, img):


vrabaud:
we don't need the while anymore right ? waitKey is a blocker . We just need k = cv2.waitKey() no ?

trigal:
technically speaking, yes. I changed that line only to change the lowest number of lines, but they're useless. I can update the pull request if needed. 

Moreover, I just tried pressing 'q' for quit and nothing happens (but only in the monocular case) but this is a new issue ...

vrabaud:
Yes, please change. I checked the history and it seems it comes from cv.waitKeye. Not sure on how it worked in OpenCV1 but the behavior has changed anyway so we might as well change it.
For 'q', it is supposed to quit according to the doc ? If that's the case, please fix too as it is in the same part of the code.

k-okada:
added updated version https://github.com/ros-perception/image_pipeline/pull/340

JWhitleyWork:
This PR is super stale and @k-okada seems to have written replacements. Closing.

PR #293, opened 2017-09-12
Small change to image_rotate transform lookups.

ghost:
Changed calls to lookupTransform to include a timeout equal to the previous frame delay. This requires the transform to be only newer than the previous frame and significantly reduces extrapolation errors.

### DIFF
--- a/image_rotate/src/nodelet/image_rotate_nodelet.cpp
+++ b/image_rotate/src/nodelet/image_rotate_nodelet.cpp
@@ -124,14 +124,14 @@ class ImageRotateNodelet : public nodelet::Nodelet
       target_vector_.header.stamp = msg->header.stamp;
       target_vector_.header.frame_id = target_frame_id;
       geometry_msgs::Vector3Stamped target_vector_transformed;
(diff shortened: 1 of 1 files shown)


vrabaud:
Thx !

mintar:
I just stumbled across this PR when looking at the recent commits and was confused by the commit message, so I checked. TL;DR: The commit is correct, only the justification is wrong. I'm only posting this here in case somebody else comes across the same thing. @mhosmar-cpr : Please excuse me for being nitpickish! :)

> This requires the transform to be only newer than the previous frame

That's not actually correct. Adding a timeout doesn't change the semantics of the lookout, i.e. tf still requires one tf message before the requested time and one after, so it can interpolate. What adding a timeout does is that lookupTransform won't fail immediately if the second tf message isn't in the buffer yet; it will wait up to the specified timeout before failing. This is what reduces the extrapolation errors.

There's nothing special about waiting for exactly the time between the two previous frames however. You could as well wait for 0.1 seconds or 2 seconds or whatever; you should set this to the maximum latency that you're willing to accept. Usually the tfs will arrive much earlier than that, so it usually doesn't hurt setting a larger timeout.

PR #298, opened 2017-09-29
Removing code from nodes

davelkan-zz:
Currently several classes exist in the camera_calibration/nodes directory. This prevents anyone from importing these classes for use in other projects. Moving the classes to their own file in the camera_calibration/src directory would allow importing of these classes without impacting functionality of the package. Besides doing this I noticed a few imports that did not appear to be used, so I removed these. 

I have tested successfully the cameracalibrator.py and camerachecker.py and using a monocular camera but do not have a stereo set up to test with.


davelkan-zz:
build failures appears unrelated to this pr

vrabaud:
agreed. Can you please use "git mv" so that we can keep the history of those classes ? Thx.

davelkan-zz:
Tried using git mv but same result. Looked into and apparently git history is based on filename. Using git mv tells the history to look at a new filename, however adding back the original filename, which is necessary in this case in order to preserve the functionality without requiring users to suddenly memorize a new executable. I'm looking into solutions now, but i'm currently not sure how to get the history attached to src/ files as requested.

davelkan-zz:
This pr has been replaced by https://github.com/ros-perception/image_pipeline/pull/301 and will be closed shortly.

vrabaud:
Thx for trying. Changes were too big for git. I replicated your method with different commits. Thx !

PR #299, opened 2017-10-06
 Fix image_resize nodelet

bikramak:
image_resize nodelet implementation was incomplete.
This patch fixes missing parts to use image_resize nodelet.

### DIFF
--- a/image_proc/src/nodelets/resize.cpp
+++ b/image_proc/src/nodelets/resize.cpp
@@ -80,7 +80,7 @@ void ResizeNodelet::onInit()
   reconfigure_server_->setCallback(f);
 
   pub_info_ = advertise<sensor_msgs::CameraInfo>(*pnh_, "camera_info", 1);
-  pub_image_ = advertise<sensor_msgs::Image>(*pnh_, "image", 1);
+  pub_image_ = advertise<sensor_msgs::Image>(*pnh_, "out_image", 1);
 
   onInitPostProcess();
 }
@@ -93,7 +93,7 @@ void ResizeNodelet::configCb(Config &config, uint32_t level)
 void ResizeNodelet::subscribe()
 {
   sub_info_ = nh_->subscribe("camera_info", 1, &ResizeNodelet::infoCb, this);
-  sub_image_ = nh_->subscribe("image", 1, &ResizeNodelet::imageCb, this);
+  sub_image_ = nh_->subscribe("in_image", 1, &ResizeNodelet::imageCb, this);
 }
 
 void ResizeNodelet::unsubscribe()
@@ -151,21 +151,32 @@ void ResizeNodelet::imageCb(const sensor_msgs::ImageConstPtr& image_msg)
     config = config_;
   }
 
-  cv_bridge::CvImagePtr cv_ptr = cv_bridge::toCvCopy(image_msg);
+  cv_bridge::CvImageConstPtr cv_ptr;
+  try
+  {
+    cv_ptr = cv_bridge::toCvShare(image_msg, sensor_msgs::image_encodings::BGR8);
+  }
+  catch (cv_bridge::Exception& e)
+  {
+    ROS_ERROR("cv_bridge exception: %s", e.what());
+    return;
+  }
+  cv::Mat buffer;
 
   if (config.use_scale)
   {
-    cv::resize(cv_ptr->image, cv_ptr->image, cv::Size(0, 0),
+    cv::resize(cv_ptr->image, buffer, cv::Size(0, 0),
                config.scale_width, config.scale_height, config.interpolation);
   }
   else
   {
     int height = config.height == -1 ? image_msg->height : config.height;
     int width = config.width == -1 ? image_msg->width : config.width;
-    cv::resize(cv_ptr->image, cv_ptr->image, cv::Size(width, height), 0, 0, config.interpolation);
+    cv::resize(cv_ptr->image, buffer, cv::Size(width, height), 0, 0, config.interpolation);
   }
 
-  pub_image_.publish(cv_ptr->toImageMsg());
+  sensor_msgs::ImagePtr out_msg = cv_bridge::CvImage(std_msgs::Header(), "bgr8", buffer).toImageMsg();  
+  pub_image_.publish(out_msg);
 }
 
 }  // namespace image_proc
--- a/image_proc/CMakeLists.txt
+++ b/image_proc/CMakeLists.txt
@@ -1,9 +1,16 @@
 cmake_minimum_required(VERSION 2.8)
 project(image_proc)
 
-find_package(catkin REQUIRED)
+find_package(catkin REQUIRED COMPONENTS
+  cv_bridge
+  dynamic_reconfigure
+  image_geometry
+  image_transport
+  nodelet
+  nodelet_topic_tools
+  roscpp
+  sensor_msgs)
 
-find_package(catkin REQUIRED cv_bridge dynamic_reconfigure image_geometry image_transport nodelet nodelet_topic_tools roscpp sensor_msgs)
 find_package(OpenCV REQUIRED)
 find_package(Boost REQUIRED COMPONENTS thread)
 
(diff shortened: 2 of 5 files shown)


bikramak:
@vrabaud I fixed image_resize nodelet whose implementation was not complete as far as I understand. I have also refactored some parts of the code to expose parameters. I have tried to keep the code backwards compatible. Please let me know, if you suggest any modifications before approving this PR.
Thanks for your maintainership.

vrabaud:
It was pretty incomplete indeed ! Thx !

efernandez:
Why is this adding a nodelet manager? https://github.com/ros-perception/image_pipeline/pull/299/files#diff-49d9b5ca47c27aafe0b87d7dac3e26edR30

This breaks anything that was running in a nodelet manager inside a namespace, because the `manager` arg now is also passed to the nodelet manager `name`, and node names aren't allowed to contain namespaces. Before it was only used for nodelet slaves.

efernandez:
@vrabaud Could we revert this, or at least don't run (or pass an arg to enable/disable, which should default to false) the nodelet manager in the `image_proc.launch`?
That breaks backwards compatibility.

I you agree on any of my suggestions, I can put up a PR.

efernandez:
FYI @teyvoniaperception

efernandez:
I've just sent #309 to address the issue I mentioned here. We can continue the discussion there.

image_proc/src/nodelets/resize.cpp
```diff
@@ -93,7 +93,7 @@ void ResizeNodelet::configCb(Config &config, uint32_t level)
 void ResizeNodelet::subscribe()
 {
   sub_info_ = nh_->subscribe("camera_info", 1, &ResizeNodelet::infoCb, this);
-  sub_image_ = nh_->subscribe("image", 1, &ResizeNodelet::imageCb, this);
+  sub_image_ = nh_->subscribe("in_image", 1, &ResizeNodelet::imageCb, this);
```
knorth55, line 96: !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
@vrabaud 
This causes a lot of error in our system
!!!!!!!!!!!!!!!!!!!

knorth55:
@vrabaud 
We strongly want to revert this PR and release fixed version.
`topic` name is converted with no comment, and we suffered a lot.

- `image` -> `in_image`
- `~image` -> `out_image`

furushchev:
-100

I also have the same situation with @knorth55 
This breaks existing api (e.g. `image` is changed to `in_image`). By these changes, all the configuration that use this nodelet must change existing settings.
Since this repository is fundamental and depended by huge packages, this type of changes should be discussed more and more carefully.
(At least the title is not enough for this change, IMO...)

knorth55:
Although this PR is good enough to try to implement new feature, it collapse all the other surroundings.
Now, new version with this PR is released, and everyone is suffering the topic problems.

IMO, we should split this PR commit (actually it only has one commit!) into several commit, and discuss  for all of them.
I think we need to discuss about topic name because it is old-fashion now, but we should not modify it in this PR.

furushchev:
FYI, this change is not yet synced with wiki http://wiki.ros.org/image_proc

130s:
For Kinetic, public sync is planned to happen soon so I just [sent a headsup](https://discourse.ros.org/t/preparing-for-kinetic-sync-2017-11-29/3275/3?u=130s).

wkentaro:
@bikramak @vrabaud Well, what was the **incomplete** part of `image_resize`? I don't understand your point. I have used it without any problems.

image_proc/src/nodelets/resize.cpp
efernandez: I could also fix this in https://github.com/ros-perception/image_pipeline/pull/309, but I'd wait for some feedback from the maintainers.

Is there anything else broken, other than the nodelet manager I had to removed in https://github.com/ros-perception/image_pipeline/pull/309?

knorth55: IMO, we should revert this PR first to get 'no-bug' repo, then we can start debugging.

furushchev: +1 for just reverting and then clarifying which part is 'incomplete'

bikramak:
Apologies team for breaking the backwards compatibility. I would like to explain the changes that I have made so that the future corrections are in the right direction.

@wkentaro and @vrabaud  This PR was submitted on top of [98670bdfc9f](https://github.com/ros-perception/image_pipeline/tree/98670bdfc9fbd22a3ee7f4e75c3ac66965fcbb5d/image_proc).

This PR loads  image_resize nodelet into image_proc. In [nodes/image_proc.cpp:L89](https://github.com/ros-perception/image_pipeline/blob/indigo/image_proc/src/nodes/image_proc.cpp#L89), I attempted to fix the issue that did not load`image_proc/resize` into the nodelet manager. This meant that the image_resize nodelet was not integrated into image_proc pipeline. I forked the branch to diagnose the issue and added these two lines. Following that fix, the image_resize nodelet came into action which opened up another issue in [nodelets/resize.cpp](https://github.com/ros-perception/image_pipeline/blob/98670bdfc9fbd22a3ee7f4e75c3ac66965fcbb5d/image_proc/src/nodelets/resize.cpp) file.

In [nodelets/resize.cpp](https://github.com/ros-perception/image_pipeline/blob/98670bdfc9fbd22a3ee7f4e75c3ac66965fcbb5d/image_proc/src/nodelets/resize.cpp) the nodelet would crash at [line:158](https://github.com/ros-perception/image_pipeline/blob/98670bdfc9fbd22a3ee7f4e75c3ac66965fcbb5d/image_proc/src/nodelets/resize.cpp#L158). IMO the `cv_bridge::CvImagePtr ` instance was not implemented in complete. First it should have been `cv_bridge::CvImageConstPtr`  and  it might have also been that the `sensor_msgs::image_encodings::BGR8` was not specified when instantiated ([ref](https://github.com/ros-perception/vision_opencv/issues/20)) .
I refactored the code to use `cv_bridge::toCvShare` instead of `cv_bridge::toCvCopy` and used a `cv::Mat buffer` to store result from `cv::resize`.

The refactoring of the input and output topics was made backwards compatible at the roslaunch file level. This happens to have caused an issue in backwards compatibility when using this package without the given launch files. 

The resolution could include the above mentioned changes apart from removing the topic remappings at the `node.cpp` file level.

knorth55:
@bikramak 
IMO,
all the refactoring should be splitted into another commit, because it is difficult to trace back who is responsible for the change.
Refactoring is good, but `git blame` shows that you change `CMakeLists.txt` or `Resize.cfg` for 'Fix image_resize nodelet ', which means that these files contain some error.
However, you change these files just for refactoring and not fixing error.
We should not do like this because future users get confused at mismatching between commit message and commit changes.
Actually, it took me some hours to realize that this PR modify the topic name.

https://github.com/ros-perception/image_pipeline/pull/299/commits/0e4745a37ee72884fc8d26fa23f855b137b92450
Your commit is too big, and as you write, it contains 5 big changes (and also refactoring of quotation and topic name change).
Keeping this commit, and putting debugging commit may cause other new errors.
We should first go backward to keep backcompatibility and make sure that this repo works fine.
And then, we can move forward.

knorth55:
@bikramak 

>This PR loads image_resize nodelet into image_proc. In nodes/image_proc.cpp:L89, I attempted to fix the issue that did not loadimage_proc/resize into the nodelet manager. This meant that the image_resize nodelet was not integrated into image_proc pipeline. I forked the branch to diagnose the issue and added these two lines. Following that fix, the image_resize nodelet came into action which opened up another issue in nodelets/resize.cpp file.

According to above, what you mainly want to change is these lines below, is it right?
https://github.com/akrobotics/image_pipeline/blob/0e4745a37ee72884fc8d26fa23f855b137b92450/image_proc/src/nodes/image_proc.cpp#L88-L91

knorth55:
@bikramak 
#311 is merged so that all this PR is reverted.
You can once again apply this change, but I recommend you to split commits and make several PRs.

One PR is about this change.
https://github.com/akrobotics/image_pipeline/blob/0e4745a37ee72884fc8d26fa23f855b137b92450/image_proc/src/nodes/image_proc.cpp#L88-L91

>In nodelets/resize.cpp the nodelet would crash at line:158. IMO the cv_bridge::CvImagePtr instance was not implemented in complete. First it should have been cv_bridge::CvImageConstPtr and it might have also been that the sensor_msgs::image_encodings::BGR8 was not specified when instantiated (ref) .
I refactored the code to use cv_bridge::toCvShare instead of cv_bridge::toCvCopy and used a cv::Mat buffer to store result from cv::resize.

The other PR is about this.

Other refactoring requires discussion, but backward compatibility is really important, IMO.

PR #302, opened 2017-10-30
Opencv datatype deprecation fix

kdhansen:
In OpenCV 3.3.1, `cv::DataType::type` was deprecated, which broke `image_proc` and `stereo_image_proc`.

`image_proc` was easily fixed by migrating to the recommended `cv::traits::Type::value`. `stereo_image_proc`however, uses the `mat.hpp` which has not been updated to use `cv::traits::Type::value`. My fix for that is to add the `OPENCV_TRAITS_ENABLE_DEPRECATED`definition to the `CMakeLists.txt`, which is a temporary fix until OpenCV is fixed.

### DIFF
--- a/image_proc/src/nodelets/crop_decimate.cpp
+++ b/image_proc/src/nodelets/crop_decimate.cpp
@@ -120,7 +120,7 @@ template <typename T>
 void debayer2x2toBGR(const cv::Mat& src, cv::Mat& dst, int R, int G1, int G2, int B)
 {
   typedef cv::Vec<T, 3> DstPixel; // 8- or 16-bit BGR
-  dst.create(src.rows / 2, src.cols / 2, cv::DataType<DstPixel>::type);
+  dst.create(src.rows / 2, src.cols / 2, cv::traits::Type<DstPixel>::value);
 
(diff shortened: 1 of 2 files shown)


kdhansen:
The CI fails as it has not an updated OpenCV.

kdhansen:
The problem with mat.hpp in OpenCV has been addressed in https://github.com/opencv/opencv/commit/72f789bf34575ed9b9756028d25ef883568c9fd4#diff-bc1d784738cd852f5b1e95ce10a56d06 but is not yet in the release.

mikepurvis:
This is the `indigo` branch being targeted, so either we need an `ifdef` to check the version of OpenCV and only use the new definitions when applicable, or this repo needs to be forked off to a new branch for Lunar+.

Alternatively, it's probably safe to add the `OPENCV_TRAITS_ENABLE_DEPRECATED` definition in Indigo+, but that just kicks the ball forward until the deprecated version really is removed.

@vrabaud What's your preferred path forward on this?

vrabaud:
Closed in favor of https://github.com/ros-perception/image_pipeline/pull/303

PR #304, opened 2017-11-07
link libimage_publisher against OpenCV

mbreyer:
When compiling image_publisher, I got linker errors with symbols in the cv::VideoCapture namespace. Linking against OpenCV solved the problem.

My Setup: macOS 10.13 with OpenCV version 3.3.1.

### DIFF
--- a/image_publisher/CMakeLists.txt
+++ b/image_publisher/CMakeLists.txt
@@ -8,10 +8,12 @@ generate_dynamic_reconfigure_options(cfg/ImagePublisher.cfg)
 
 catkin_package()
 
+find_package(OpenCV REQUIRED core videoio)
+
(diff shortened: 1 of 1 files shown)


oceanusxiv:
@mbreyer May I ask if you have XCode itself or XCode CLT installed? I've just encountered this issue on my macOS 10.13 using XCode CLT, but managed a successful build when suing XCode app toolchain

mbreyer:
@eric1221bday I got the same error with both XCode 9 and XCode CLT. However, I'm also trying to install everything with python3, which might indirectly break something.

vrabaud:
Fixed in https://github.com/ros-perception/image_pipeline/pull/318

PR #307, opened 2017-11-14
Adding image_transport implementation to image_proc/resize nodelet

doisyg:


doisyg:
Please do not merge yet, i think that the changes break the resize functionality with some cameras

doisyg:
fixed

afdaniele:
Hi, any news about this? Thanks.

mistoll:
I needed this, too. And it woks great. However you changed the topic names and thus it's not backward compatible. I changed them back: https://github.com/mistoll/image_pipeline/commit/b3c32277f555b5a6818c2e429c5a4a505e4a5041
IMO that commit will make this PR more likely to be merged.

Tuebel:
I used the @mistoll code and found that the nodelet would not work on raw depth images.
The reason is that the encoding has been hardcoded to "bgr8".

I propose to go the old way and use the plain `cv_bridge::toCvCopy` without any arguments to keep the input encoding (the way it was done before nodeleting it). A new cv::Mat has been used for `cv::resize` anyway so it shouldn't make a difference in performance.

Checkout my version: https://github.com/Tuebel/image_pipeline/blob/530281a12fc17de483e8dfb06a04eeb6ab397c72/image_proc/src/nodelets/resize.cpp

Edit: I also need this functionality to record resized compressed images via rosbag.

mistoll:
Great to see some action here. I'm going to try your code. 

Using `cv_bridge::toCvCopy` sounds resonable. However I'd like to know why @bikramak changed it to `cv_bridge::toCvShare` in #299. (https://github.com/ros-perception/image_pipeline/pull/299#issuecomment-349413632)

mistoll:
I've been using https://github.com/Tuebel/image_pipeline/blob/530281a12fc17de483e8dfb06a04eeb6ab397c72/image_proc/src/nodelets/resize.cpp for a while. Works perfectly and it's much cleaner than this PR. IMO you (@Tuebel) should create a new PR for that version.

image_proc/src/nodelets/resize.cpp
```diff
@@ -31,27 +31,37 @@
...
 namespace image_proc
 {
 
-class ResizeNodelet : public nodelet_topic_tools::NodeletLazy
+class ResizeNodelet : public nodelet::Nodelet
 {
 protected:
   // ROS communication
-  ros::Publisher pub_image_;
+  image_transport::Publisher pub_image_;
```
SteveMacenski, line 57: Any reason this isnt a CameraPublisher (deals with the info stuff for you)?

SteveMacenski (commented):
I have no issue in concept here, but I'm not a direct user of this nodelet. I'd love some feedback if this API change effects folks. Certainly this should not be backported as its a significant change. I'd love to see the new PR as well.

mistoll:
IMO we should take special care to not change any existing topic names. However currently there are new topics added for image transport. To be 100% compatible we could introduce a parameter which enables the image_transport support. The default would be false for current releases and true for future releases.

SteveMacenski:
@mistoll 100% agree. There shouldnt be any topic name changes, and I dont think this would do that. @doisyg to verify please. 

Image transport adds _new_ topics based on the availability of the compression plugins, but the raw topics shouldnt change at all. It will just expose additional topics should you want to use a compressed version

image_proc/src/nodelets/resize.cpp
Tuebel: I don't see a reason why the CameraPublisher shouldn't be used. In my version I'm already using it which makes it much cleaner. I will create a new PR as mistoll suggested.

JWhitleyWork:
Closing in favor of #409.

PR #309, opened 2017-11-27
Fix image proc launch

efernandez:
This fixes `image_proc.launch` after https://github.com/ros-perception/image_pipeline/pull/299, where the nodelet manager was added to the launch file. I guess it was added by mistake by @bikramak, causing existing code/launch files to fail.

In my case the failure is due to the nodelet manager being in a namespace:
``` bash
RLException: while processing /opt/clearpath/2.8devel/sdk/share/image_proc/launch/image_proc.launch:
Invalid <node> tag: node name cannot contain a namespace. 

Node xml is <node args="manager" launch-prefix="$(arg launch_prefix)" name="$(arg manager)" output="screen" pkg="nodelet" type="nodelet">
     <param name="num_worker_threads" value="$(arg num_worker_threads)"/>
  </node>
The traceback for the exception was written to the log file
```

But in general the main issue is that the manager is run from another launch file, so at the end it gets run twice.

I've also changed three lines with a `respawn` that were indented with a tab instead of spaces, but that actually fit in 120 columns; I assume this is the desired max number columns because there was one > 80 columns. If you want I can remove this from the PR though.

FYI @vrabaud

### DIFF
--- a/image_proc/launch/image_proc.launch
+++ b/image_proc/launch/image_proc.launch
@@ -22,18 +22,15 @@
   <!-- Debayered images -->
   <!-- TODO Arguments for debayer, interpolation methods? -->
   <node if="$(arg debayer)" pkg="nodelet" type="nodelet" name="debayer"
-        args="load image_proc/debayer $(arg manager) $(arg bond)"
-	respawn="$(arg respawn)" />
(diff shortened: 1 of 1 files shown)


efernandez:
Since #311 has reverted #299 I'm going to close this one #309

PR #314, opened 2017-12-15
Fixed bug in cv_bridge conversion - deal with 8UC1 encoding to mono8.

jeffdelmerico:
This came up when trying to calibrate one of the cameras on an Intel Realsense sensor.  The image was encoded as 8UC1, and cv_bridge complained about it not being a color format.  Forcing it to be interpreted as uint8 resolves this problem.


vrabaud:
8UC1 is not a color format unfortunately. mono8 is. Yes, data is the same but what if I send heat temperature or depth as 8UC1 and it gets converted to mono8 automatically thus becoming a color ? The driver is wrong and you unfortunately need to create a wrapper around it.

PR #318, opened 2018-02-19
fix 'VideoCapture' undefined symbol error

hannometer:
The following error occured when trying to run image_publisher:
[...]/devel/lib/image_publisher/image_publisher: symbol lookup error: [...]/devel/lib//libimage_publisher.so: undefined symbol: _ZN2cv12VideoCaptureC1Ev

Probably, changes in cv_bridge reducing the OpenCV component dependencies led to the error. See
https://github.com/ros-perception/vision_opencv/commit/8b5bbcbc1ce65734dc600695487909e0c67c1033

This is fixed by manually finding OpenCV with the required components and adding the dependencies to the library, not just the node.


trigal:
this needs to be merged _asap_ ! :+1:

PR #320, opened 2018-03-16
Added dashcam mode to video_recorder

IvanVN:
Added an optional "dashcam mode" for video recording. It will store a sequence of videos in a loop, deleting the oldest one.
Several parameters have been added:
- dashcam_mode: (bool) Enabled/disabled dashcam mode.
- video_length: (double) Length in minutes of each video in dashcam mode.
- n_videos: (int) Number of videos in the loop of the dashcam mode.
- video_topic: (string) input image topic
- use_posix_timestamp: (bool) Use posix formated (human readable) timestamps in filenames.

### DIFF
--- a/image_view/src/nodes/video_recorder.cpp
+++ b/image_view/src/nodes/video_recorder.cpp
@@ -26,6 +26,8 @@
 #if CV_MAJOR_VERSION == 3
 #include <opencv2/videoio.hpp>
 #endif
+#include <stdio.h>
+#include "boost/date_time/posix_time/posix_time.hpp"
 
 cv::VideoWriter outputVideo;
 
@@ -34,20 +36,60 @@ ros::Time g_last_wrote_time = ros::Time(0);
 std::string encoding;
 std::string codec;
 int fps;
-std::string filename;
+std::string base_filename, filename;
 double min_depth_range;
 double max_depth_range;
 bool use_dynamic_range;
 int colormap;
+bool stamped_filename;
+bool dashcam;
+ros::Duration video_length;
+int n_videos;
+ros::Time video_creation_stamp;
+std::vector<std::string> filenames;
+std::string video_topic;
+bool use_posix;
 
+void generateStampedFilename(std::string &filename) {
+  std::size_t found = base_filename.find_last_of("/\\");
+  std::string path = base_filename.substr(0, found + 1);
+  std::string basename = base_filename.substr(found + 1);
+  std::stringstream ss;
+  if (use_posix) {
+    boost::posix_time::ptime my_posix_time = ros::Time::now().toBoost();
+    std::string iso_time_str = boost::posix_time::to_iso_extended_string(my_posix_time);
+    ss << iso_time_str << "_" << basename;
+  }
+  else {
+    ss << ros::Time::now().toNSec() << "_" << basename;
+  }
+  filename = path + ss.str();
+  ROS_INFO("Video recording to %s", filename.c_str());
+  // If dashcam mode, store filenames in a vector
+  if (dashcam) {
+    filenames.push_back(filename);
+    // Delete oldest video
+    if (filenames.size() > n_videos) {
+      remove(filenames.front().c_str());
+      filenames.erase(filenames.begin());
+    }
+  }
+}
 
 void callback(const sensor_msgs::ImageConstPtr& image_msg)
 {
     if (!outputVideo.isOpened()) {
 
+      // Generate stamped file name if required or in dashcam mode
+      if (stamped_filename or dashcam) {
+        generateStampedFilename(filename);
+      }
+      else {
+        filename = base_filename;
+      }
         cv::Size size(image_msg->width, image_msg->height);
 
-        outputVideo.open(filename, 
+        outputVideo.open(filename,
 #if CV_MAJOR_VERSION == 3
                 cv::VideoWriter::fourcc(codec.c_str()[0],
 #else
@@ -55,10 +97,11 @@ void callback(const sensor_msgs::ImageConstPtr& image_msg)
 #endif
                           codec.c_str()[1],
                           codec.c_str()[2],
-                          codec.c_str()[3]), 
+                          codec.c_str()[3]),
                 fps,
                 size,
                 true);
+        video_creation_stamp = ros::Time::now();
 
         if (!outputVideo.isOpened())
         {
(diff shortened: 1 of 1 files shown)


JWhitleyWork:
I know this is old (just over a year) but I like the idea. However, I would rather call it something like "rolling_buffer," "ring_buffer," or similar instead of "dashcam" which is much more specific and, IMO, has other implications. Opinions welcome from the community (and maintainers - @SteveMacenski / @mjcarroll )

@IvanVN - If you're still keen on this type of thing, would you mind rebasing on `melodic`?

mjcarroll:
Agree, I think that this is an interesting and useful feature.

I agree that the name should also be changed.  "rolling_buffer" seems like a suitable name.  I believe that the GoPro calls it looping.

One feature that would be nice to have is to have a service/topic to trigger saving the current contents and starting a new loop.  That way you could use something else in the system (say a failure incident or something) to cause it to store the duration of the video up to when the incident occurs.

SteveMacenski:
I agree with @mjcarroll. Rolling buffer would make the most sense to me, and the ability to save sections of it would be useful. However I wouldnt make that feature as a requirement for merging this PR, but rather a feature request on a new ticket after this has been rebased and merged.

IvanVN:
@JWhitleyAStuff I'll try to do it. Sadly right know I've to do it in my spare time and I don't have any melodic setup to test it, so I guess it will take me some time. If anyone wants to do it in the meantime, please feel free to do so.

SteveMacenski:
@IvanVN Kinetic and Melodic are essentially the same for all intents and purposes for your PR. If you make sure it compiles and works in Kinetic and it passes Melodic CI, I think its fine.

IvanVN:
I think it should be done. Compiled and tested in Kinetic. It also compiled OK in a Melodic docker.

image_view/src/nodes/video_recorder.cpp
```diff
@@ -115,15 +164,20 @@ int main(int argc, char** argv)
...
-      std::size_t found = filename.find_last_of("/\\");
-      std::string path = filename.substr(0, found + 1);
-      std::string basename = filename.substr(found + 1);
-      std::stringstream ss;
-      ss << ros::Time::now().toNSec() << basename;
-      filename = path + ss.str();
-      ROS_INFO("Video recording to %s", filename.c_str());
+    local_nh.param("rolling_buffer_mode", rolling_buffer, false); // rolling_buffer mode enabled/disabled
+    double v_duration;
+    local_nh.param("video_length", v_duration, 5.0);
```
JWhitleyWork: Can you please prefix the parameters that are only used for rolling_buffer_mode with `rb_` or `rbf_` to clarify that they aren't generic (e.g. `video_length` could be interpreted as "length of video to record" for any use of this node).

JWhitleyWork (changes requested):
See comment about parameters. I've looked over it more thoroughly and this should be the last change request from me before approval. Thanks!

image_view/src/nodes/video_recorder.cpp
IvanVN: Done at 690da7f :+1:

image_view/src/nodes/video_recorder.cpp
```diff
@@ -34,31 +36,73 @@ ros::Time g_last_wrote_time = ros::Time(0);
...
-std::string filename;
+std::string base_filename, filename;
 double min_depth_range;
 double max_depth_range;
 bool use_dynamic_range;
 int colormap;
-
+bool stamped_filename;
+bool rolling_buffer;
+ros::Duration rb_video_lenght;
```
SteveMacenski: Can you fix  this typo?

image_view/src/nodes/video_recorder.cpp
```diff
@@ -34,31 +36,73 @@ ros::Time g_last_wrote_time = ros::Time(0);
...
+int rb_n_videos;
+ros::Time video_creation_stamp;
+std::vector<std::string> filenames;
```
SteveMacenski: I'm a super huge not-fan of this. This is a global variable of variable size. The other variables while changing probably cannot cause additional memory allocation, but this will constantly. This shouldn't be global or should be set at runtime to a constant unchanging size.

SteveMacenski (changes requested):

image_view/src/nodes/video_recorder.cpp
IvanVN: Sorry, but my c++ skills are pretty basic and I'm not sure of how to exactly do what you suggest. I would need some help on this.

I guess that it would be enough to just do `filenames.reserve(rb_n_videos + 1)` once we read the `rb_n_videos` parameter for the `std:vector`. However, there is no way (that I know of) of reserving also the size of the stored `std::string`, right?

Second option that comes to my mind is to use something like a non global `std::array<char[filename.size()], rb_n_videos>  filenames` and use it as input/output parameter to `generateStampedFilename` function. But I really need to check if this can be done (and how).

Anyway (and probably I'll be beaten up for this comment), in the worst case you would have `rb_n_videos + 1` memory allocations... Is it really so much of a big deal? :sweat_smile: I though this was exactly the purpose of the `vector` type.

JWhitleyWork: @IvanVN - Just a couple of suggestions/question here:

1. How are you actually using this filenames list? It looks like you use a function called `remove()` on the first filename in the vector but I don't exactly know what this function is doing or where it is included.
2. This should probably be a `queue` instead of a `vector`. A vector is not designed for efficiently removing items from the front of the list. A `queue` is exactly for this - it's an implementation of a FIFO (first-in-first-out) buffer.
3. If you reserve the amount of memory ahead of time equal to `rb_n_videos`, then that's the majority of the allocation done up-front. No, it won't reserve the size of each `std::string` but a vector is essentially a list of pointers to the actual storage locations for the elements anyway so the `reserve(N)` function will reserve the memory for one pointer for each element N. The pointed-to elements may change in size or location but the allocated memory for each pointer remains the same unless you add N+1 elements.
4. Rather than allocating `rb_n_videos` + 1, you could check the size of the container before adding the new element.

SteveMacenski: All above work. I know this pipeline may be used on some of the delivery robots and those guys are more concerned about that type of thing than your typical indoor robot (but far less than the autonomous car people). Just making sure we're not introducing a changing memory allocation where there wasn't one before would be good. The other global variables are statics so on runtime they allocate once and just hang out, this buffer changes size which is what I'm wanting to resolve.

SteveMacenski:
What's the status of this @IvanVN

IvanVN:
> What's the status of this @IvanVN

Sorry, I haven't have the time yet to look into this...

JWhitleyWork:
@IvanVN Could I please get an update on this (and a rebase) if you're still interested in merging it? Sorry it's been forever since I checked back on it.

IvanVN:
> @IvanVN Could I please get an update on this (and a rebase) if you're still interested in merging it? Sorry it's been forever since I checked back on it.

I've replaced the vector for a queue, as suggested. I'm having troubles rebasing to noetic, though. It doesn't compile because of some missing package (my current setup is melodic). I will come back to this in a couple of days and test it in a noetic docker container.

JWhitleyWork:
@IvanVN Given that this has been hanging out for 2+ years, I'm going to close it as incomplete. If you come back to this at some time in the future, please re-open or create a new PR.

PR #322, opened 2018-03-27
Add option to filter out close points from point cloud

tylerlum:
Sometimes, the created point cloud has points too close to the camera that we do not want. This will allow the user to remove all data points too close to the camera. 

Use on command-line: `rosrun nodelet nodelet load depth_image_proc/point_cloud_xyz nodelet_manager _range_min:=1.4`

### DIFF
--- a/depth_image_proc/include/depth_image_proc/depth_conversions.h
+++ b/depth_image_proc/include/depth_image_proc/depth_conversions.h
@@ -51,7 +51,8 @@ void convert(
     const sensor_msgs::ImageConstPtr& depth_msg,
     PointCloud::Ptr& cloud_msg,
     const image_geometry::PinholeCameraModel& model,
-    double range_max = 0.0)
+    double range_max = 0.0,
+    double range_min = 0.0)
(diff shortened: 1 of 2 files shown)


SteveMacenski:
This can be handled in a nodelet filter after this pipeline, why add here? If its a valid point from matching I dont think this would be an appropriate place to add it.

I also see you added a maximum range field to the function defaulted to 0 but then unused. Is there something there I'm missing? 

@JWhitleyAStuff what do you think?

tylerlum:
Hi Steven,

Thanks for the feedback. A nodelet filter probably is a smarter solution. I
didn't see a nodelet for this at the time I made the request, so I thought
I'd share this, but I don't think it's needed.

Best regards,
Tyler

SteveMacenski:
Ok, there is a pass through filter nodelet you can take advantage of http://wiki.ros.org/pcl_ros/Tutorials/filters#PassThrough

tylerlum:
Great thank you!

PR #329, opened 2018-05-03
Implemented extracting raw image data

Fruchtzwerg94:
Implementation of the raw image extraction if the file extension is "*.raw". This file extension is not supported by cv::imwrite so there is be no conflict. The raw files only containing the pixel data without any meta data which allows usage in MATLAB or other tools.

I also added a warning if OpenCV failed to save an image file.

### DIFF
--- a/image_view/src/nodes/extract_images.cpp
+++ b/image_view/src/nodes/extract_images.cpp
@@ -42,6 +42,8 @@
 #include <boost/thread.hpp>
 #include <boost/format.hpp>
 
+#include <fstream>
+
 class ExtractImages
 {
 private:
@@ -117,7 +119,29 @@ public:
         std::string filename = (filename_format_ % count_).str();
 
 #if !defined(_VIDEO)
-        cv::imwrite(filename, image);
+        // Save raw image if the defined file extension is ".raw", otherwise use OpenCV
+        std::string file_extension = filename.substr(filename.length() - 4, 4);
+        if (filename.length() >= 4 && file_extension == ".raw")
+        {
+          std::ofstream raw_file;
+          raw_file.open(filename.c_str());
+          if (raw_file.is_open() == false)
+          {
+            ROS_WARN_STREAM("Failed to open file " << filename);
+          }
+          else
+          {
+            raw_file.write((char*)(msg->data.data()), msg->data.size());
+            raw_file.close();
+          }
+        }
+        else
+        {
+          if (cv::imwrite(filename, image) == false)
+          {
+            ROS_WARN_STREAM("Failed to save image " << filename);
+          }
+        }
 #else
         if(!video_writer)
         {


JWhitleyWork (approved):

SteveMacenski (commented):
This seems like a pretty strange special use-case that I don't think necessarily needs to be supported in this node. 

I think a better way to handle this situation would be a conversion script from a image to a matlab readable file rather than adding this. This seems to make the code less readable and confusing what ".raw" means and why that this is needed or how it should be properly used.

Fruchtzwerg94:
Sure @SteveMacenski , extracting raw images is a special use case but a very interesting one. I don't like the solution using additional scripts since the job can be done very easily in the extractor. Code readability is not as good as before, maybe some additional comments are the solution.

Nevertheless, the current implementation ignores the return value of `imwrite`. The check I've implemented in the commit should be added definitely:
`
if (cv::imwrite(filename, image) == false)
{
    ROS_WARN_STREAM("Failed to save image " << filename);
}
`

SteveMacenski:
@Fruchtzwerg94 that's true, but if we make it a script then we can make it a catkin global so it can be generally used in any number of other applications. Making it a separate script to open an openCV image -> convert to matlab readable -> save back could have alot of use in other packages or projects.

Fruchtzwerg94:
From my point of view, a direct integration in the existing code is more handy @SteveMacenski . But I will not force anybody to integrate the changes. What do you think @JWhitleyAStuff . Would it be better to create a script here? If yes, I could change  (simplify) this pull request to only handle the failure case already mentioned above.

SteveMacenski:
A third opinion is always good :-)

JWhitleyWork:
Honestly, I'm inclined to agree with @Fruchtzwerg94. It's a fairly small change but:

1. *.raw images are used outside of Matlab (e.g. Adobe products, other image editing software, etc.) and retain more image data than their compressed counterparts.
2. It won't affect any other types.
3. Saving directly from the node keeps from adding in a compression/decompression step which loses data.

Just my opinion.

SteveMacenski:
Ok that's fine, we can do that then.

I was trying ot think about portability. THis might be a useful little conversion script  for other people to be able to `./convert image.jpg image.raw` as a general use and could find value in other places.

SteveMacenski (approved):

Fruchtzwerg94:
Thanks for the discussion and merge @SteveMacenski and @JWhitleyAStuff . I've also created a pull request to backport the changes to kinetic. You can find the backport merge at #413 . I've done it similar to #406 . Would be great also to have the changes in kinetic, which is my mainly used distro.

JWhitleyWork:
@SteveMacenski - I get where you're coming from but the problem is that, if you've already saved the image as a JPG, you've already lost quality (JPG uses compression and is therefore a lossy format) which means converting it back to .raw gets you compatibility but doesn't retain the quality. Having it included here means that the raw images from the camera can be saved without degradation due to compression.

SteveMacenski:
huh, I didn't know it defaulted to jpg, we could add a flag for png which should be lossless: https://docs.opencv.org/3.0-alpha/modules/imgcodecs/doc/reading_and_writing_images.html#imwrite

Issue #246, opened 2017-01-23
[camera_calibration] Doesn't work well with mono16 images


mintar:
When running `cameracalibrator.py` on a `mono16` image stream, the image appears all white. This is because `cameracalibrator.py` simply clips all gray values to the lowest 256 values:

[calibrator.py:247](https://github.com/ros-perception/image_pipeline/blob/f94bfe7b03397a78cedd2686e8fb69d234b38447/camera_calibration/src/camera_calibration/calibrator.py#L247)

IMO, it should multiply each gray value by 1/256 instead.

I have verified this on my own image data: When I multiply the input image by 1/256 before feeding it into `cameracalibrator.py`, it works.

mintar:
There's an open pull request that solves this issue: #334

JWhitleyWork:
Closed by #334.

PR #334, opened 2018-05-09
Scale pixels down from 16 to 8 bits instead of just clipping

Fruchtzwerg94:
Clipping 16 bit pixels just to 8 bit pixels leads to white images if the original image uses the full range of the 16 bits. Instead the pixel should be scaled down to 8 bits.


JWhitleyWork:
Verified as working by #246. Merging.

PR #336, opened 2018-05-15
depth_image_proc/point_cloud_xyzi_radial Add intensity conversion (copy) for float

madsherlock:
This commit enables the generation of xyzi point clouds from 32-bit floating point intensity images.
The destination data type for intensity storage is 32-bit float, so all that is required is a data copy.
The change in this commit is simply an extension of the if-else statement to include the TYPE_32FC1 type and apply the usual convert_intensity() method.

This applies to the depth_image_proc/point_cloud_xyzi_radial and depth_image_proc/point_cloud_xyzi nodelets.

PS: I do have a 3D camera with a manufacturer-supported ROS driver providing me with 32-bit float intensity images.

### DIFF
--- a/depth_image_proc/src/nodelets/point_cloud_xyzi.cpp
+++ b/depth_image_proc/src/nodelets/point_cloud_xyzi.cpp
@@ -239,6 +239,11 @@ void PointCloudXyziNodelet::imageCb(const sensor_msgs::ImageConstPtr& depth_msg,
   {
     convert<uint16_t, uint16_t>(depth_msg, intensity_msg, cloud_msg);
   }
+  else if (depth_msg->encoding == enc::TYPE_16UC1 &&
+      intensity_msg->encoding == enc::TYPE_32FC1)
+  {
+    convert<uint16_t,float>(depth_msg, intensity_msg, cloud_msg);
+  }
   else if (depth_msg->encoding == enc::TYPE_32FC1 &&
       intensity_msg->encoding == enc::MONO8)
   {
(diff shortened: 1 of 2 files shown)


k-okada:
can you share your bug fiile so that everyone can share your problem?

--
◉ Kei Okada


2018-05-16 8:40 GMT+09:00 Mikael Westermann <notifications@github.com>:

madsherlock:
I would love to, but I don't understand your request.

I added a minor feature to the depth processing node.

Do you want to see the error message that would be generated if I tried using the unmodified node on my incompatible data?

k-okada:
> Do you want to see the error message

yes, error message and how the node behavies.

madsherlock:
I will try and generate such an error later.
In the meantime, I can tell you the error that _would_ be produced in the image callback of the **point_cloud_xyzi_radial** node:
`
Intensity image has unsupported encoding [32FC1]
`
This error would be thrown by a call to NODELET_ERROR_THROTTLE() at a maximum rate of 5 Hz and the image callback would return without publishing a point cloud.


A similar error message _would_ be produced in the image callback of the **point_cloud_xyzi** node:
`
Depth image has unsupported encoding [32FC1]
`
And otherwise the exact same behaviour would be observed.


To produce these error messages, you need the nodes to subscribe to an intensity image with the **TYPE_32FC1** data type.



Edit:
Here is a log excerpt from May 16th, right before I found out the feature was missing:
`
1526423924.533520542 ERROR [/media/data/catkin_ws/src/image_pipeline/depth_image_proc/src/nodelets/point_cloud_xyzi_radial.cpp:250(PointCloudXyziRadialNodelet::imageCb) [topics: /rosout, /ifm3d/camera/cloud_rect, /ifm3d/bond] Intensity image has unsupported encoding [32FC1]
`

SteveMacenski:
@JWhitleyAStuff any issue merging this?

JWhitleyWork:
LGTM! Onward!

Issue #331, opened 2018-05-08
[image_view] window hangs because callback can be dispatched from multiple threads


yoshito-okada:
I found that the image_view nodelet (image_view/image) does not update its window and hangs in some cases.  I think this is because the nodelet manager can execute a callback in multiple threads even if the callback is associated to a threaded callback queue (see https://github.com/ros/nodelet_core/blob/2f25b775fd0241fc949a5fc002e718dbd142d5d7/nodelet/src/callback_queue_manager.cpp#L206), although the callback must be called by a specific thread to successfully update the window.

The following is a launch script to reproduce the problem.  I am afraid that the problem may not be reproduced in different environment because the problem depends on environment-specific threading but I hope this is helpful.
```
<launch>
    <node name="image_view_node" pkg="nodelet" type="nodelet" 
          args="manager image_view_node" output="screen"/>

    <node name="image_view_loader" pkg="nodelet" type="nodelet" 
          args="load image_view/image image_view_node" output="screen"/>

    <node name="image_view_loader2" pkg="nodelet" type="nodelet" 
          args="load image_view/image image_view_node" output="screen"/>

    <!-- An external node publishing sensor_msgs/Image on the "image" topic is required -->
</launch>
```

I think the only solution is to add a new thread updating the window and am writing a workaround.  Please see if interested.
https://github.com/yoshito-okada/image_pipeline/tree/fix_image_view_nodelet

yoshito-okada:
Sorry, I put a wrong launch script in my original post.  The following one would reproduce the problem.

```
<launch>
    <node name="image_view_node" pkg="nodelet" type="nodelet" 
          args="manager image_view_node" output="screen"/>

    <node name="image_view_loader" pkg="nodelet" type="nodelet" 
          args="load image_view/image image_view_node" output="screen">
        <remap from="image" to="image_mono"/>
    </node>

    <node name="image_proc_loader" pkg="nodelet" type="nodelet" 
          args="load image_proc/debayer image_view_node" output="screen">
        <remap from="image_raw" to="image"/>
    </node>

    <!-- An external node publishing sensor_msgs/Image on the "image" topic is required -->
</launch>
```

k-okada:
[...]

PR #337, opened 2018-05-17
Fix image view nodelet

yoshito-okada:
Requested in the discussion #331


k-okada:
@vrabaud I think this is ok to merge, if we could not find better solution.

yoshito-okada:
how is this going?

JWhitleyWork:
@yoshito-okada - Maintenance has resumed on this repo. Could you possibly merge changes from `indigo` into your branch? I would like to also wait until we have #398 complete to verify cross-version compatibility.

yoshito-okada:
@JWhitleyAStuff - Thank you for the response!  Please check a merge commit I just made.

SteveMacenski:
Woah, I dont see any queue management here (i.e. queue size gets _really big_ because images are getting backed up). There needs to be something here to check and retain queue size so it doesn't infinitely lag on weak computers

JWhitleyWork:
@SteveMacenski - It isn't really a "queue" so much as it is two image storage locations which are thread-safe using mutexes. The `while(true)` loop is just to try to pull the latest image from the `queued_image_` for display. If it takes too long to display an image, the `queued_image_` will be `set()` again by the callback and the old image will be de-allocated. I don't see any "queue size" growing here.

SteveMacenski:
Ah got it. I saw get/set/pop and assumed the  data structure was a queue. All good :+1:

PR #338, opened 2018-05-18
add approximate_sync args in stereo_image_proc.launch

### LINKED ISSUE #323: [ WARN] [1522489818.929206010]: [image_transport] Topics '/zed/right/image_mono' and '/zed/right/camera_info' do not appear to be synchronized.
Hi, 
I am trying to get point clouds from zed camera but I got this warning on terminal and no point clouds present on rviz. I did some googling and found that that is caused by non sycned topics. I tried to place 
`
    <param name="approximate_sync" value="true"/>
    <param name="queue_size" value="10"/>
`
to sync the topics but the error still remains when I run `roslaunch stereo_image_proc some.launch`

here is what I have in `some.launch` file 

```
<launch>

  <!-- Start a nodelet manager; the name must be unqualified here -->
  <node pkg="nodelet" type="nodelet" name="pipeline" args="manager" />

  <!-- Include the stereo_image_proc.launch file and pass in a qualified manager name -->
  <include file="$(find stereo_image_proc)/launch/stereo_image_proc.launch" ns="zed">
    <arg name="manager" value="/pipeline" />
  </include>

</launch>
```
How I can overcome this issue and get my point clouds published ? 

Thanks

k-okada:
this will close https://github.com/ros-perception/image_pipeline/issues/323

### DIFF
--- a/stereo_image_proc/launch/stereo_image_proc.launch
+++ b/stereo_image_proc/launch/stereo_image_proc.launch
@@ -5,7 +5,7 @@
   <arg name="respawn" default="false" />
   <arg name="left" default="left" />
   <arg name="right" default="right" />
-  <!-- TODO Arguments for sync policy, etc? -->
+  <arg name="approximate_sync" default="false" />
 
   <arg     if="$(arg respawn)" name="bond" value="" />
   <arg unless="$(arg respawn)" name="bond" value="--no-bond" />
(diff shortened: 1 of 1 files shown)


JWhitleyWork:
Just a launch file change with default value being innocuous. Merging.

PR #340, opened 2018-05-19
use waitKey(0) instead of while loop

### LINKED ISSUE #285: tarfile_calibration does not visualize rectified images after calibration

k-okada:
rewrited version of #286

Closes #286
Closes #285

### DIFF
--- a/camera_calibration/scripts/tarfile_calibration.py
+++ b/camera_calibration/scripts/tarfile_calibration.py
@@ -45,16 +45,10 @@ from camera_calibration.calibrator import MonoCalibrator, StereoCalibrator, Cali
 import rospy
 import sensor_msgs.srv
 
-def waitkey():
-    k = cv2.waitKey(6)
-    return k
-
 def display(win_name, img):
     cv2.namedWindow(win_name, cv2.WINDOW_NORMAL)
     cv2.imshow( win_name,  numpy.asarray( img[:,:] ))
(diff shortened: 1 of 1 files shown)


JWhitleyWork:
@k-okada - Would you mind rebasing this on the `melodic` branch? We can then test and pull and consider backport to Kinetic since Indigo is EOL.

k-okada:
ok done

SteveMacenski (approved):

JWhitleyWork (approved):

PR #347, opened 2018-07-24
implement registration: rgb to depth

stereoboy:
I wrote additional nodelet for registration from rgb frame to depth frame. I am using Intel Realsense d435. RGB camara's fov is just 70 degree, even though depth camera's fov about 90 degree. I want to use a wider fov frame.


SteveMacenski:
@stereoboy doesn't the `depth_image_proc/point_cloud_xyzrgb` do something very similar?

SteveMacenski:
Closing from no response. Feel free to reopen if active.

PR #355, opened 2018-09-14
Add option to use the RGB image timestamp for the registered depth image

roehling:
For some applications, it is beneficial that the registered depth image
adopts the time stamp from the RGB image, e.g. if one wants to pair them
with ExactTime policy afterwards. It also makes sense from a theoretical
point of view, since the registration aligns the frames both spatially
and temporally.

The new behavior can be enabled on a opt-in basis by setting the
parameter "use_rgb_timestamp".

### DIFF
--- a/depth_image_proc/src/nodelets/register.cpp
+++ b/depth_image_proc/src/nodelets/register.cpp
@@ -70,6 +70,7 @@ class RegisterNodelet : public nodelet::Nodelet
   image_transport::CameraPublisher pub_registered_;
 
   image_geometry::PinholeCameraModel depth_model_, rgb_model_;
+  bool use_rgb_timestamp_;
 
   virtual void onInit();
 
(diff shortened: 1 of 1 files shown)


JWhitleyWork:
This LGTM. @SteveMacenski - Any issues here?

SteveMacenski:
@JWhitleyAStuff I dont see anything wrong with it as in breaking people's implementations. I think if you're using exact policy you would want to change it to an approximate one. Changing the timestamp here could mess with TF down the road for user that decide to use this option, especially in high speed or dynamic environments. 

I'm not sure I agree this belongs here. But if we're going to add it, I'd like clear and explicit documentation why its there and what the potential fallout for sychronization are. Reassigning timestamps isn't kosher. 

Sorry taking me so long to get to these, the 101 `image_pipeline` notifications the last couple weeks has pushed me behind :-).

roehling:
The ApproximateTime policy is somewhat unstable in its association, namely it is not guaranteed that subsequent nodes will associate the same RGB and depth frames again. This is an actual problem we have, not just a theoretical one. However, the register node is the authoritative one, since it transforms the depth image to the RGB TF frame, and no other association will make sense afterwards. And if someone needed the original timestamp, they would probably get it from the original input stream anyway.

I'm happy to add documentation if we can agree that the feature itself makes sense.

SteveMacenski:
@roehling I can agree to that.

JWhitleyWork:
@roehling This apparently completely dropped off my radar. Now that `melodic` is EoL, would you mind re-targeting this feature to `noetic`? If you don't have time, I totally understand and will close the PR for now.

JWhitleyWork:
Sorry, I was wrong - `melodic` is not EoL yet. However, this does need a rebase.

roehling:
I think I deleted the original fork already, so I'm going to reopen this here: https://github.com/ros-perception/image_pipeline/pull/791

Issue #357, opened 2018-09-22
image_publisher dynamic reconfigure server needs to use private namespace


lucasw:
Not doing so means the parameters are retrieved from the nodelet manager namespace and collisions between dynamic reconfigure servers arise if two image_publishers are loaded into same manager.

JWhitleyWork:
Closed by #358

PR #358, opened 2018-09-22
Use a shared_ptr for the dynamic reconfigure pointer, and create it w…

lucasw:
…ith the private node handle so that the parameters for the dynamic reconfigure server are in the private namespace and two image publishers can coexist in the same manager #357 

This may break parameters that were set in the nodelet manager  namespace as a workaround in some launch files out there.

### DIFF
--- a/image_publisher/src/nodelet/image_publisher_nodelet.cpp
+++ b/image_publisher/src/nodelet/image_publisher_nodelet.cpp
@@ -48,7 +48,8 @@ using namespace boost::assign;
 namespace image_publisher {
 class ImagePublisherNodelet : public nodelet::Nodelet
 {
-  dynamic_reconfigure::Server<image_publisher::ImagePublisherConfig> srv;
+  typedef dynamic_reconfigure::Server<image_publisher::ImagePublisherConfig> ReconfigureServer;
+  boost::shared_ptr<ReconfigureServer> srv;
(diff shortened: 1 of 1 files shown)


JWhitleyWork:
@lucasw - Thanks for the PR! I know this is a bit old but, would you mind removing your custom launch file and rebasing this on `melodic`? Also, I'm OK with the use of std::shared_ptr<>. We should be upgrading to it across the code-base for `melodic` anyway.

lucasw:
Wouldn't you rather take this as a boost shared_ptr, then have a separate PR that switches out all the boost shared_ptrs for all of image_pipeline?  (created an issue #407)  It seems odd to just have one std shared_ptr and then 40 boost shared_ptrs.

JWhitleyWork:
@lucasw - That's a good point. Thanks for the update. Would you mind rebasing this on `melodic` to grab the new CI? After, I'll have one other person review and then we'll merge.

mjcarroll (approved):

lucasw:
I've removed the launch file, rebased to melodic, and squashed.  I left a backup of the launch file for myself in https://github.com/lucasw/image_pipeline/tree/indigo_image_pub_dr_private_namespace .

JWhitleyWork (approved):

PR #362, opened 2018-10-22
Port depth image proc on ROS2

yechun1:


yechun1:
This PR is the first module(depth_image_proc) porting from image_pipeline. Before going ahead to port others, we hope the PR could be reviewed and merged, so that we could apply the same changes on later porting. And once the PR could be merged, we would also take effort to build and release depth_image_proc with bloom before Crystal release.

This PR passed the colcon test and most of nodes have be verified based on Realsense camera(as kinect  not work on ROS2 right now, some features such as radial could not be verified with Realsense). The nodes could be implemented by ros2 launch, for example: to generate pointcloud with rgb and depth, with the command:
`"ros2 launch depth_image_proc point_cloud_xyzrgb.launch.py"  `

Some dependent APIs are not implemented on ROS2, we have added TODOs with workaround in the code, and the depth_image_proc features will not be impact by those workaround, we think this PR is ready for merge, and we would submit other PRs in the future to fix TODOs.

@mjcarroll would you please help to review or ask someone help to review this patch?  many thanks.

depth_image_proc/CMakeLists.txt
```diff
@@ -1,44 +1,93 @@
...
+find_package(image_transport REQUIRED)
+find_package(message_filters REQUIRED)
+find_package(class_loader REQUIRED)
+find_package(sensor_msgs REQUIRED)
+find_package(stereo_msgs REQUIRED)
+find_package(tf2 REQUIRED)
+find_package(tf2_ros REQUIRED)
+
 
 if(cv_bridge_VERSION VERSION_GREATER "1.12.0")
```
mjcarroll: This logic is not needed now, as we're already setting the standard above.

depth_image_proc/CMakeLists.txt
```diff
@@ -1,44 +1,93 @@
...
+  src/nodelets/point_cloud_xyz.cpp
+  src/nodelets/point_cloud_xyzrgb.cpp
+  src/nodelets/point_cloud_xyzi.cpp
+  src/nodelets/point_cloud_xyz_radial.cpp
+  src/nodelets/point_cloud_xyzi_radial.cpp
+  src/nodelets/register.cpp
 )
-target_link_libraries(${PROJECT_NAME} ${catkin_LIBRARIES} ${OpenCV_LIBRARIES})
+ament_target_dependencies(${PROJECT_NAME}
+  "image_transport"
```
mjcarroll, line 46: Nit: The quotes aren't necessary here.

depth_image_proc/CMakeLists.txt
```diff
@@ -1,44 +1,93 @@
...
 
 if(cv_bridge_VERSION VERSION_GREATER "1.12.0")
   add_compile_options(-std=c++11)
 endif()
 
-catkin_package(
-    INCLUDE_DIRS include
-    LIBRARIES ${PROJECT_NAME})
-
 find_package(Boost REQUIRED)
```
mjcarroll: Is Boost still being used? It looks like it has been removed.

depth_image_proc/package.xml
```diff
@@ -14,33 +14,31 @@
   <license>BSD</license>
```
mjcarroll, line 15: Probably should update maintainer.

mjcarroll (changes requested):
Some initial feedback, I need to check this out and build/run as well.

depth_image_proc/CMakeLists.txt
yechun1: Removed cv_bridge_VERSION check logic, fixed in c13e2a3

depth_image_proc/CMakeLists.txt
yechun1: If remove "image_transport" from ament_target_dependencies, the build will fail with "fatal error: image_transport/image_transport.h: No such file or directory", so I just keep it.
```
ament_target_dependencies(${PROJECT_NAME}
#  "image_transport"
```

depth_image_proc/CMakeLists.txt
yechun1: removed.

depth_image_proc/package.xml
yechun1: updated in c13e2a3

yechun1:
@mjcarroll thanks for your effort to review the code, I have updated the patch, would you please help to check? Any comments are welcome, We'll hurry up and update the patch ASAP. We'll try best to make the PR ready for merge and build bloom package in Crystal release.

yechun1:
One more commit: 5e5b54b
Passed implementation tests based on Realsense camera and display image/points on RVIZ.

Enabled demo launcher
     ros2 launch depth_image_proc point_cloud_xyzrgb.launch.py
     ros2 launch depth_image_proc point_cloud_xyz.launch.py
     ros2 launch depth_image_proc convert_metric.launch.py
     ros2 launch depth_image_proc crop_foremost.launch.py
     ros2 launch depth_image_proc point_cloud_xyz_radial.launch.py
     ros2 launch depth_image_proc disparity.launch.py
     ros2 launch depth_image_proc register.launch.py
     ros2 launch depth_image_proc point_cloud_xyzi.launch.py
     ros2 launch depth_image_proc point_cloud_xyzi_radial.launch.py

mjcarroll:
Hey @mkhansen-intel I have been looking at this one this afternoon and it made me recognize a potential shortcoming in the way that `image_transport` and `message_filters` are implemented.

Currently, we make those two libraries take `rclcpp::Node::SharedPtr`, which makes the code a bit unweildy when it's used as part of an object that inherits from `rclcpp::Node`.  I have some proposed suggestions to those two packages as well as a bit of refactoring here incoming in the next few hours.

depth_image_proc/src/nodelets/convert_metric.cpp
```diff
@@ -1,140 +1,145 @@
...
+  rclcpp::Logger logger_ = rclcpp::get_logger("ConvertMetricNode");
 };
 
-void ConvertMetricNodelet::onInit()
+ConvertMetricNode::ConvertMetricNode()
+: Node("ConvertMetricNode")
 {
-  ros::NodeHandle& nh = getNodeHandle();
-  it_.reset(new image_transport::ImageTransport(nh));
+  rclcpp::Node::SharedPtr node = std::shared_ptr<rclcpp::Node>(this);
```
mjcarroll: While this appears to work for you, it looks to be unsafe based on the documentation is here: https://en.cppreference.com/w/cpp/memory/enable_shared_from_this

The docs indicate: 
> enable_shared_from_this provides the safe alternative to an expression like std::shared_ptr<T>(this), which is likely to result in this being destructed more than once by multiple owners that are unaware of each other (see example below)

The first workaround that I thought of is that we can just use `shared_from_this`, but using it in the constructor could also lead to potentially undefined behavior.

mjcarroll: My proposed workaround is to allow `message_filters` and `image_transport` take a raw pointer to `Node`, because it's unlikely that the `SharedPtr` is providing real benefit or safety here.

mjcarroll:
I've made some progress but have to stop for a bit, you can see what I have so far here: https://github.com/ros-perception/image_pipeline/tree/mjcarroll/pointer_api_updates

I think that shows most of the direction that I would expect it to go in.

In the meantime, please take a look at, which both block those changes:

* ros2/message_filters#14
* ros-perception/image_common#104

yechun1:
Let me continue the progress and complete the code changes by 202411f 

Block changes:
- ros2/message_filters#14
- ros-perception/image_common#104
- https://github.com/intel/ros2_intel_realsense/pull/15

mjcarroll:
@yechun1 Since we don't have any tests in here, it would be really helpful if you could test with your hardware or add tests.

I don't believe that any of these changes should fundamentally change behavior, but there is always an opportunity to introduce a bug or two in there.

mjcarroll:
There is also an update to the transport plugins here: ros-perception/image_transport_plugins#31

yechun1:
@mjcaroll, we have added launch demo for all depth_image_proc nodes, and verified the published topic (include pointcloud and images) could be displayed via RVIZ.

below commands are workable but depends on three block changes list above. 

 ros2 launch depth_image_proc point_cloud_xyzrgb.launch.py
 ros2 launch depth_image_proc point_cloud_xyz.launch.py
 ros2 launch depth_image_proc convert_metric.launch.py
 ros2 launch depth_image_proc crop_foremost.launch.py
 ros2 launch depth_image_proc point_cloud_xyz_radial.launch.py
 ros2 launch depth_image_proc disparity.launch.py
 ros2 launch depth_image_proc register.launch.py
 ros2 launch depth_image_proc point_cloud_xyzi.launch.py
 ros2 launch depth_image_proc point_cloud_xyzi_radial.launch.py

yechun1:
The launcher tests based on Intel Realsense R435 Camera Hardware as resource input.

mjcarroll (approved):

yechun1:
@mjcarroll thanks very much for merging this code :)
The dependent PR (https://github.com/ros-perception/image_common/pull/104)  is also require to merge

PR #368, opened 2018-11-29
enable rclcpp_register_node_plugins

yechun1:
this may be remarked while code debugging, should be enabled to build node plugin file


yechun1:
@mjcarroll donot know when rclcpp_register_node_plugins had been marked. This occurs launcher execution not work. The PR to fix the issue, would you please help to merge?

mjcarroll (approved):

PR #375, opened 2018-12-17
Fix OpenCV4 compatibility.

hgaiser:
This PR adds compatibility for OpenCV4.

### DIFF
--- a/image_view/src/nodes/image_view.cpp
+++ b/image_view/src/nodes/image_view.cpp
@@ -165,7 +165,7 @@ int main(int argc, char **argv)
     // Handle window size
     bool autosize;
     local_nh.param("autosize", autosize, false);
-    cv::namedWindow(g_window_name, autosize ? (CV_WINDOW_AUTOSIZE | CV_WINDOW_KEEPRATIO | CV_GUI_EXPANDED) : 0);
+    cv::namedWindow(g_window_name, autosize ? (cv::WINDOW_AUTOSIZE | cv::WINDOW_KEEPRATIO | cv::WINDOW_GUI_EXPANDED) : 0);
     cv::setMouseCallback(g_window_name, &mouseCb);
 
     if(autosize == false)
--- a/image_view/src/nodes/video_recorder.cpp
+++ b/image_view/src/nodes/video_recorder.cpp
@@ -48,7 +48,7 @@ void callback(const sensor_msgs::ImageConstPtr& image_msg)
         cv::Size size(image_msg->width, image_msg->height);
 
         outputVideo.open(filename, 
-#if CV_MAJOR_VERSION == 3
+#if CV_MAJOR_VERSION >= 3
                 cv::VideoWriter::fourcc(codec.c_str()[0],
 #else
                 CV_FOURCC(codec.c_str()[0],
--- a/stereo_image_proc/include/stereo_image_proc/processor.h
+++ b/stereo_image_proc/include/stereo_image_proc/processor.h
@@ -56,7 +56,7 @@ class StereoProcessor
 public:
   
   StereoProcessor()
-#if CV_MAJOR_VERSION == 3
+#if CV_MAJOR_VERSION >= 3
   {
     block_matcher_ = cv::StereoBM::create();
     sg_block_matcher_ = cv::StereoSGBM::create(1, 1, 10);
@@ -168,7 +168,7 @@ private:
   image_proc::Processor mono_processor_;
   
   mutable cv::Mat_<int16_t> disparity16_; // scratch buffer for 16-bit signed disparity image
-#if CV_MAJOR_VERSION == 3
+#if CV_MAJOR_VERSION >= 3
   mutable cv::Ptr<cv::StereoBM> block_matcher_; // contains scratch buffers for block matching
   mutable cv::Ptr<cv::StereoSGBM> sg_block_matcher_;
 #else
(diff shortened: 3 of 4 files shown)


hgaiser:
Any update on this? @vrabaud

SteveMacenski:
@hgaiser is openCV4 used in any ROS distribution?

JWhitleyWork:
@SteveMacenski - It isn't supported, even in ROS2 Crystal, certainly not Melodic. Maybe in Noetic, though. I'm going to close for now but we can re-open if Noetic adds support.

SteveMacenski:
Lets not be too hasty, the ROS distribution not supporting it isn't necessarily a reason ignore it. If it doesn't break compatibility for existing openCV1,2,3 we should consider it

hgaiser:
I agree with @SteveMacenski . There not being a ros distribution that uses opencv4 seems like a bad reason to close this. The PR is backwards compatible, there should be no harm in prematurely adding these changes, for when ros decides to support opencv4.

JWhitleyWork:
Sorry for the hasti-ness. @hgaiser - Would you mind rebasing from or merging the changes in the `indigo` branch so your PR goes through the new CI? Also, are you actually still using this for `indigo` or have you upgraded? I would like to keep as many changes as possible moved forward out of Indigo for stability and due to it being EOL'd.

SteveMacenski:
Indigo is EOL, we should be targeting all changes to Melodic and cherry picking down to Kinetic for big upgrades. @hgaiser thanks for the work, I'd like to see a testing plan / things you've tried to make sure it doesn't break anyone using default openCV versions in Kinetic/Melodic (opencv 2,3)

hgaiser:
I'll update this PR later this week. A testing plan seems a bit overkill. I've changed some ifdefs that tested for opencv3, to ifdefs that test for opencv2. Additionally the header that is included has existed since opencv2. Shouldn't the unit tests handle any testing?

JWhitleyWork:
@hgaiser - I think this is more about getting an environment set up in CI to do this test. Normally, we use the standard `ros:melodic` Docker which is going to include only OpenCV 3.2.0. Would it be possible for you to provide us with a Docker image (preferably on Docker Hub) that uses `ros:melodic` as a base and adds OpenCV 4? I don't think it's as simple as just installing OpenCV 4 (because of Melodic package dependencies) but maybe it is.

hgaiser:
I updated it to be based from melodic.

hgaiser:
@JWhitleyAStuff regarding the Docker image, the amount of effort to make this is unreasonable for this PR since it also requires all ROS packages linked to OpenCV to be recompiled. It has been tested on Arch Linux, which contains OpenCV4. I would wait with setting up CI for OpenCV4 until there is a ROS distribution that uses it.

SteveMacenski:
@hgaiser a couple of questions just for my openCV4 ignorance: 
- You'd changed the conditions to set 2's as a certain thing, and 3 otherwise. Is that to mean opencv4 has the same API for all these things?
- What does `highgui_c` contain that you needed `highgui` did not? I'm struggling to see where in this PR there's a new thing called outside rearranging conditions.

I dont think the docker file is absolutely necessary as long as its been tested on your side in opencv4, we can independently test in opencv3 and opencv2. While I dont think this is going to be an issue, releasing debians for this and having it crash would be *very very bad*

de-vri-es:
> @hgaiser a couple of questions just for my openCV4 ignorance:
> 
> You'd changed the conditions to set 2's as a certain thing, and 3 otherwise. Is that to mean opencv4 has the same+ API for all these things?

Yes, opencv4 is still mostly backwards compatible with opencv3. In general it also makes more sense to special-case older versions rather than the newest version, since it's much more likely that later versions will be similar to the current version rather than an older version.

> What does `highgui_c` contain that you needed `highgui` did not? I'm struggling to see where in this PR there's a new thing called outside rearranging conditions.

The header contains a C interface, including some non-namespaced enums that are used here and there: https://github.com/ros-perception/image_pipeline/blob/d53b457ad4512a98c0df51bb321d8e17cb66655b/image_view/src/nodes/image_view.cpp#L169

In opencv3 and earlier, the C++ interface automatically pulls in the C interface. This is no longer the case in opencv4.

Another option could be to use the namespaced enums instead of adding the `highui_c.h` header. These are also already present in opencv2. Including the header is a smaller change though.

SteveMacenski:
Got it. Ok, given that Melodic ships with openCV3, we should at minimum validate 
- [ ] opencv4 works for you on this
- [x] opencv3 works for all others in melodic (@JWhitleyAStuff do you run melodic to test)
- [ ] opencv2 works for all users in kinetic, though we probably wont port this to kinetic (I can take this)

SteveMacenski:
This is a great future proofing effort. I'm a fan.

hgaiser:
I refactored the PR to be as small as possible. I chose to use the `cv::WINDOW_` definitions, instead of the `CV_WINDOW_` definitions, since this PR is set to merge with melodic (which requires opencv3). Alternatively I can change the file to include `highgui_c.h` to use the backwards compatible definitions if desired.

JWhitleyWork:
@hgaiser Great work on this. Totally on-board with these changes.

JWhitleyWork:
@SteveMacenski - It built against Melodic and Kinetic correctly. No runtime testing is done in the CI builds (yet) but everything compiled just fine.

SteveMacenski:
Oh! With that refactor, yes I approve, very simple and clearly not going to break anything for non-4 users (which isnt officially supported, so my burden of proof is incredibly low)

SteveMacenski (approved):

SteveMacenski:
With a second maintainer's approval, go ahead and merge

JWhitleyWork (approved):

PR #379, opened 2019-01-17
Fix boost 1.69 compatibility

de-vri-es:
The signals library was not used at all, and it has been removed from boost 1.69. As a result, the package doesn't build anymore with boost 1.69 without this change.


de-vri-es:
Note: CI fails before actually testing these changes.

SteveMacenski (approved):
@JWhitleyAStuff LGTM

I want to make sure someone else takes a look at things and approves and not just 1 persons' opinion

JWhitleyWork (approved):
I'm not actually sure why this was here in the first place.

SteveMacenski:
May have been boiler plate at one point or used before CV windows

de-vri-es:
My guess: copy pasta from another `CMakeLists.txt`

PR #389, opened 2019-03-05
View Image in Fullscreen

dietriro:
I needed the functionality to view an image in fullscreen instead of in a window. Therefore I implemented a quick fix. Maybe others are in need of this functionality as well.


SteveMacenski (approved):
LGTM @JWhitleyAStuff

JWhitleyWork:
@dietriro - Has this been tested for strange interactions with the `width` and `height` parameters?

JWhitleyWork:
Since `image_view` has been replaced by the nodelet, this will need to be re-implemented. @dietriro - Please implement in `nodelets/image_nodelet.cpp` if you want to keep this functionality.

PR #391, opened 2019-03-14
image_saver: Support lazy subscribing

furushchev:
This pull request includes following updates:

- Support lazy subscribing input topic

  Subscribing image topic to be saved only when saving is requested.

- Add `~wait_for_save` & `~save_timeout` params

  `image_saver` has three ways for saving images (though not yet documented)

  1. Oneshot save by using `~save` service
  2. Durative save by using `~start` & `~end` service
  3. Continuous save by setting `~save_all_images` param

  On oneshot way of saving, call of service `~save` did never return until no image is published from publishers, which can lock callee forever.
  With `~wait_for_save` param, we can switch whether calling `~save` request blocks until image is saved or immediately returns from the request just like a non-blocking way.
  If we choose the `~wait_for_save` as true, then we can set `~save_timeout` param to set timeout for waiting.

### DIFF
--- a/image_view/src/nodes/image_saver.cpp
+++ b/image_view/src/nodes/image_saver.cpp
@@ -53,19 +53,21 @@ public:
             const std::string& filename_format,
             const std::string& encoding,
             const bool save_all_image,
-            const bool wait_for_save)
-      : it_(nh), topic_(topic), encoding_(encoding),
+            const bool wait_for_save,
+            const double save_timeout)
+      : nh_(nh), it_(nh_), topic_(topic), encoding_(encoding),
         count_(0), start_time_(0),
-        subscribing_(false), should_unsubscribe_(false), wait_for_save_(wait_for_save)
+        subscribing_(false), should_unsubscribe_(false),
+        wait_for_save_(wait_for_save), save_timeout_(save_timeout)
   {
     format_.parse(filename_format);
     if (save_all_image)
     {
       subscribe();
     } else {
-      srv_save_ = nh.advertiseService("save", &Callbacks::callbackSave, this);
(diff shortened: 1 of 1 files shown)


knorth55:
@furushchev this PR needs minor bug fix. I made PR for the bug fix.
https://github.com/furushchev/image_pipeline/pull/1

JWhitleyWork:
@furushchev - Would you mind rebasing this on `melodic`?

@knorth55 - If we don't hear back from @furushchev within a week or two, would you mind picking up this PR with your fix and rebasing on `melodic`?

knorth55:
@JWhitleyAStuff sure, I can.

JWhitleyWork:
@knorth55 - Would you mind taking over this one and creating a new PR with these changes and your fixes, based on `melodic`? We really appreciate your help.

knorth55:
We want this feature in `indigo`, too because we still use `indigo` robot.
Can i also make a backport PR?

JWhitleyWork:
@knorth55 - I'm sorry but `indigo` (and Ubuntu 14.04) are EOL. We are no longer approving new feature additions to `indigo`, just major bug fixes. Exactly for your use case: we don't want to break long-established installs.

knorth55:
Indigo is not yet EOL, but actually today (2019/04/30) is the last day of indigo Trusty.
OK, i will open to kinetic branch.

SteveMacenski:
@knorth55 Indigo is EOL. http://wiki.ros.org/Distributions as far as I can tell from that document is that modern ROS distros go EOL with Canonical

knorth55:
I open #412

Issue #396, opened 2019-04-05
Fix travis CI


SteveMacenski:

JWhitleyWork:
@SteveMacenski - For the repositories to which I have maintenance responsibilities, I have been switching to CircleCI which is more flexible, updated considerably more often, and has much lower queue times (see https://github.com/ros-drivers/velodyne/blob/master/.circleci/config.yml and https://circleci.com/gh/ros-drivers/velodyne as an example). Any opposition?

SteveMacenski:
None at all. I haven't used it personally, but that's not a reason _not_ to learn something :)

JWhitleyWork:
Thanks. I'll get on it as soon as I'm back from lunch. :)

JWhitleyWork:
Fixed in #398 by changing to CircleCI and using new build setup. Since Indigo recently EOL'd, we're only running CI on Kinetic and Melodic branches.

PR #398, opened 2019-04-08
Maint/ci circleci test

JWhitleyWork:
Created a new melodic-devel branch where we can work on features and bug fixes for release into Melodic.


JWhitleyWork:
Addresses #396.

JWhitleyWork:
@SteveMacenski - Do you think this should be pulled into `indigo`? The down-side is that it currently tests against ROS releases after Indigo - the upside is that it guarantees forward-compatibility as well. Or maybe I should just test the `indigo` branch against Indigo and have the new `melodic-devel` test against all non-EOL distros? Thoughts?

SteveMacenski:
I think we should release the changes we've just merged in Indigo, but all further development in melodic. Indigo shouldnt get updates anymore, it EOLd 8 days ago. Kinetic it would be nice to get updates if its not a pain.

SteveMacenski:
Lunar EOLs in 20 days, I'd just remove that here too. Do one last release on Indigo/Lunar or leave them as is

JWhitleyWork:
@SteveMacenski - I agree. Let me create a Kinetic branch from the current Indigo branch, then Melodic from Kinetic. I'll make this PR for Kinetic and then pull it into Melodic. I'll also make Melodic the default branch.

SteveMacenski (approved):

PR #409, opened 2019-04-26
Removed hard coded image encoding.

Tuebel:
Fixed the hard-coded image format of #307 by using `cv_bridge::toCvCopy` as a new `cv::Mat has` been used for `cv::resize` anyway.

### DIFF
--- a/image_proc/src/nodelets/resize.cpp
+++ b/image_proc/src/nodelets/resize.cpp
@@ -31,27 +31,35 @@
 *  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 *  POSSIBILITY OF SUCH DAMAGE.
 *********************************************************************/
+#include <boost/make_shared.hpp>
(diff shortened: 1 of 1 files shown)


JWhitleyWork:
@Tuebel - Two things:

1. It looks like you didn't base this change off of the current `melodic` branch. To fix this, you can either use the commands below to update your fork or create a new fork, re-apply the changes, and create a new PR.

    git remote add upstream https://github.com/ros-perception/image_pipeline
    git fetch upstream
    git checkout -b melodic
    git merge upstream melodic
    git push -u origin melodic
    git checkout -b fix/409
    git cherry-pick 530281a
    git push -u origin fix/409

Then either change this PR to use fix/409 from your fork or create a new PR.

2. Since this will be against `melodic`, and it is basically a re-write of the node, there is no reason we shouldn't use `std::shared_ptr` and `std::mutex` instead of the `boost` versions.

Tuebel:
Closed for #411

Issue #359, opened 2018-09-22
Put image_publisher filename, flip_horizontal, flip_vertical into .cfg


lucasw:

JWhitleyWork:
@lucasw - Is this something you would still like to see? If so, can you provide a PR?

mikeferguson:
There is a PR targeting ROS 1 for this - and it actually exists in ROS 2 (the param callback is setup for flip horizont/vertical and filename)

PR #410, opened 2019-04-28
Set the flip and filename through dynamic reconfigure

lucasw:
For  #359

Lots of changes needed to make this work.

Handling of camera info urls needs more consideration- if a provided url doesn't work it erases the old one, also a new image loaded erases the loaded camera info url.  What to do if the camera info url width and height don't match the image?  Generate a distortion free camera info of correct size and let the user try again?  I'll add a commit that solves both of those next.

Could consider deprecating providing the filename as a non-ros param arg in the future.

It also looks like flip vertical and flip both are backwards, I didn't fix it here.


JWhitleyWork:
@lucasw This PR was generated before I took over maintainership of this repo. Are you still able to rebase this so it can be reviewed/tested? If not, I can close the PR.

lucasw:
I'll give rebasing it a try

mikeferguson:
Melodic is EOL and no response on rebase

PR #411, opened 2019-04-29
Fix 409 based on melodic branch

Tuebel:
Based #409 on branch melodic.
Also removed  `boost::mutex` and  `boost::shared_ptr` in favor of the std types.

The constructor of the `dynamic_reconfigure::Server` requires a `boost::recursive_mutex` so in this version I use the constructor without the mutex parameter. But as far as I can tell from the code of the server the mutex is only required when calling `dynamic_reconfigure::Server::updateConfig(const ConfigType &config)` which we don't.


image_proc/src/nodelets/resize.cpp
```diff
@@ -1,171 +1,183 @@
...
+  void configCb(Config& config, uint32_t level);
 };
 
 void ResizeNodelet::onInit()
 {
-  nodelet_topic_tools::NodeletLazy::onInit();
+  ros::NodeHandle& nh = getNodeHandle();
+  ros::NodeHandle& private_nh = getPrivateNodeHandle();
+  it_.reset(new image_transport::ImageTransport(nh));
+  private_it_.reset(new image_transport::ImageTransport(private_nh));
```
JWhitleyWork: I'm not super familiar with the implementation differences between a regular subscriber and an ImageTransport subscriber. Is there a reason that this one uses the private node handle?

image_proc/src/nodelets/resize.cpp
```diff
@@ -1,171 +1,183 @@
...
 {
-  sub_info_.shutdown();
-  sub_image_.shutdown();
+  std::lock_guard<std::mutex> lock(config_mutex_);
+  config_ = config;
 }
 
-void ResizeNodelet::infoCb(const sensor_msgs::CameraInfoConstPtr& info_msg)
+void ResizeNodelet::imageCb(const sensor_msgs::ImageConstPtr& image_msg,
+                            const sensor_msgs::CameraInfoConstPtr& info_msg)
```
JWhitleyWork: What happened to the implementation of `infoCb`?

Tuebel: The callback of `image_transport::subscribeCamera` synchronizes the image and camera info. This makes the second callback unnecessary. Maybe we should rename it to `ResizeNodelet::cameraCb` to make this clear.

image_proc/src/nodelets/resize.cpp
Tuebel: When I use the regular node handle I get errors like this:
```bash
[ERROR] [1556808041.833527432]: Tried to advertise a service that is already advertised in this node [/camera/depth/image_rect_raw/compressedDepth/set_parameters]
[ERROR] [1556808041.837569377]: Tried to advertise a service that is already advertised in this node [/camera/depth/image_rect_raw/compressed/set_parameters]
[ERROR] [1556808041.841201998]: Tried to advertise a service that is already advertised in this node [/camera/depth/image_rect_raw/theora/set_parameters]
```
I think the source of this Error is located in the in image_transport_plugins library. The CompressedPublisher and the CompressedSubscriber are initialized with the same base topic so their dynamic_reconfigure services collide. By using the private namespace for the CameraPublisher this issue is avoided.

JWhitleyWork: Check out https://github.com/Tuebel/image_pipeline/pull/1

SteveMacenski: The image transport framework should use private nodehandles for the reasons above. It exposes the available image_transport_plugins for compression under the main namespace to offer `*/compressed` `*/theora` etc so if you dont make them private, they'll collide, but for organizational purposes they won't belong to the sensor namespace they are representing

image_proc/src/nodelets/resize.cpp
```diff
@@ -31,27 +31,35 @@
...
 #include <sensor_msgs/Image.h>
+#include <image_transport/image_transport.h>
 
 #include "image_proc/ResizeConfig.h"
 
 namespace image_proc
 {
 
-class ResizeNodelet : public nodelet_topic_tools::NodeletLazy
+class ResizeNodelet : public nodelet::Nodelet
```
knorth55, line 47: why do you need to change from `NodeletLazy` ?

knorth55:
I'm not familiar with `image_transport`, but does this PR work when there is no `camera_info` topic?
It seems `image_transport` uses `message_filter` so that this PR change to require both `camera_info` and `image`.
Current node can resize with only `image` topic or  `camera_info` topic.

image_proc/src/nodelets/resize.cpp
```diff
@@ -1,171 +1,180 @@
...
+ *  FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+ *  COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+ *  INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+ *  BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ *  LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+ *  CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ *  LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ *  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ *********************************************************************/
```
knorth55: which part of license did you change?

Tuebel: None of it. I guess my formatter added some spaces to align the stars.

image_proc/src/nodelets/resize.cpp
Tuebel: I based my work on #307 where doisyg changed it to `Nodelet`.
I support his decision for consistency with the other nodelets from image_proc.
Additionally the `NodeletLazy` class is not documented that well so I prefer choosing the commonly used `Nodelet` class.
Do you have any idea what could be the benefit of the `NodeletLazy` class?

knorth55: it is better to have answer from the author.
@wkentaro can you tell me why you choose to use `NodeletLazy` for this node?

image_proc/src/nodelets/resize.cpp
knorth55: If there is no need to change, it is confusing to add this kind of change for other people.
I think it is better to revert this license change.

Tuebel: Makes sense I changed it back!

image_proc/src/nodelets/resize.cpp
Tuebel: @knorth55 Having looked into his code I see why it was chosen: It automatically unsubscribes topics if no one is subscribing the output of the node. The problem is that this class has a protected method `advertise` which must be used to advertise topics. This prohibits of a `image_transport::Publisher` which is the main goal of this PR.

Tuebel:
@knorth55 `image_transport::CameraSubscriber` indeed needs both topics synchronized and you do have a good point. We should not brake compatibility, so I will try to implement the best of both by using `image_transport::Subscriber` for images and a `ros::Subscriber` for the CameraInfo.

JWhitleyWork:
@Tuebel - Since this is a back-port, can you also create a PR to apply any changes you make here to `melodic`?

JWhitleyWork (approved):

knorth55:
@Tuebel 
Thank you for checking it.
Many people use `resize` nodelet to resize `image` only, so it is good to support previous behavior.

@JWhitleyAStuff 
This PR is a back-port, but current PR changes previous behavior (message synchronizing part).
I think it is better to wait @Tuebel update and merge after it.

Tuebel:
@knorth55 I pushed an implementation, which allows to resize the image and the camera_info independently. The images are published and subscribed using the image_transport classes so we still get the functionality of the plugins. I also kept an eye on making only the necessary changes compared to the old resize.cpp.

@JWhitleyAStuff I'm sorry but I don't understand what kind of different PR should I create?

JWhitleyWork:
@Tuebel - Nevermind. I was confused. There are many tickets in-flight right now and I got them mixed up.

PR #412, opened 2019-04-30
image_saver: Support lazy subscribing 

knorth55:
Move from #391 

This pull request includes following updates:
- Support lazy subscribing input topic
- Subscribing image topic to be saved only when saving is requested.
- Add `~wait_for_save` & `~save_timeout` params
  - image_saver has three ways for saving images (though not yet documented)
    - Oneshot save by using `~save service`
    - Durative save by using `~start` & `~end` service
    - Continuous save by setting `~save_all_image` param
  - On oneshot way of saving, call of service `~save` did never return until no image is published from publishers, which can lock calleeforever.
  - With `~wait_for_save` param, we can switch whether calling `~save` request blocks until image is saved or immediately returns from the request just like a non-blocking way.
  - If we choose the `~wait_for_save` as true, then we can set `~save_timeout` param to set timeout for waiting.

### DIFF
--- a/image_view/src/nodes/image_saver.cpp
+++ b/image_view/src/nodes/image_saver.cpp
@@ -53,19 +53,21 @@ public:
             const std::string& filename_format,
             const std::string& encoding,
             const bool save_all_image,
-            const bool wait_for_save)
-      : it_(nh), topic_(topic), encoding_(encoding),
+            const bool wait_for_save,
+            const double save_timeout)
+      : nh_(nh), it_(nh_), topic_(topic), encoding_(encoding),
         count_(0), start_time_(0),
-        subscribing_(false), should_unsubscribe_(false), wait_for_save_(wait_for_save)
+        subscribing_(false), should_unsubscribe_(false),
+        wait_for_save_(wait_for_save), save_timeout_(save_timeout)
   {
     format_.parse(filename_format);
     if (save_all_image)
     {
       subscribe();
     } else {
-      srv_save_ = nh.advertiseService("save", &Callbacks::callbackSave, this);
-      srv_start_ = nh.advertiseService("start", &Callbacks::callbackStartSave, this);
-      srv_end_ = nh.advertiseService("end", &Callbacks::callbackEndSave, this);
+      srv_save_ = nh_.advertiseService("save", &Callbacks::callbackSave, this);
+      srv_start_ = nh_.advertiseService("start", &Callbacks::callbackStartSave, this);
+      srv_end_ = nh_.advertiseService("end", &Callbacks::callbackEndSave, this);
     }
   }
 
@@ -107,22 +109,45 @@ public:
     sub_image_.shutdown();
   }
 
+  void callbackTimeoutTimer(const ros::TimerEvent&)
+  {
+    ROS_DEBUG("callback timeout timer");
+    unsubscribe();
+  }
+
   bool callbackSave(std_srvs::Empty::Request &req,
                     std_srvs::Empty::Response &res)
   {
-    if (start_time_ != ros::Time(0)) {
-      ROS_ERROR("The service '~start' was called. Try calling '~end' first to use '~save' service.");
+    if (subscribing_)
+    {
+      ROS_ERROR("Images are already being saved.");
       return false;
     }
+
+    const size_t count = count_;
     should_unsubscribe_ = true;
     subscribe();
+
+    if (save_timeout_ > 0.0)
+    {
+      save_timeout_timer_ = nh_.createTimer(
+          ros::Duration(save_timeout_), &Callbacks::callbackTimeoutTimer, this,
+          /* oneshot= */true);
+    }
 
     if (wait_for_save_)
     {
       ros::Rate r(10);
-      while (ros::ok() && subscribing_) {
+      while (ros::ok() && subscribing_)
+      {
         ros::spinOnce();
         r.sleep();
+      }
+
+      if (count != count_)
+      {
+        ROS_ERROR("Timed out without saving image");
+        return false;
       }
     }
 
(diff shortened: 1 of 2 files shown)


image_view/src/nodes/image_saver.cpp
```diff
@@ -38,35 +38,127 @@
 #include <image_transport/image_transport.h>
 #include <camera_calibration_parsers/parse.h>
 #include <boost/format.hpp>
+#include <boost/thread.hpp>
```
JWhitleyWork: C++11 is fully supported in `melodic`. Please use `<thread>` instead.

image_view/src/nodes/image_saver.cpp
```diff
@@ -38,35 +38,127 @@
...
+      srv_end_ = nh_.advertiseService("end", &Callbacks::callbackEndSave, this);
+    }
   }
 
-  bool callbackStartSave(std_srvs::Trigger::Request &req,
-                         std_srvs::Trigger::Response &res)
+  void subscribe()
   {
-    ROS_INFO("Received start saving request");
+    boost::mutex::scoped_lock lock(mutex_);
```
JWhitleyWork: Please use `std::lock_guard` with a `std::mutex`.

image_view/src/nodes/image_saver.cpp
```diff
@@ -38,35 +38,127 @@
...
 
+    // Useful when CameraInfo is being published
+    sub_camera_ = it_.subscribeCamera(topic_, 1, &Callbacks::callbackWithCameraInfo, this);
+    // Useful when CameraInfo is not being published
+    sub_image_ = it_.subscribe(topic_, 1, &Callbacks::callbackWithoutCameraInfo, this);
+  }
+
+  void unsubscribe()
+  {
+    boost::mutex::scoped_lock lock(mutex_);
```
JWhitleyWork: Same.

JWhitleyWork (changes requested):
Please update launch files to match the parameter changes.

knorth55:
Thank you for your review.
I updated to follow your review.
However, I'm wondering if this change is released in Kinetic or not.
Kinetic is not EOL, and image_pipeline seems to not be released in Kinetic recently.
Do you have any plan to release in Kinetic?

knorth55:
Also, circle ci seems running test for kinetic, too, but is it a good idea to use C++11?
CI for this PR is failed on kinetic and passed on melodic because of this.

knorth55:
Sorry it was my misunderstanding.
We can use C++11 in kinetic, now I got it.
https://www.ros.org/reps/rep-0003.html

image_view/src/nodes/image_saver.cpp
knorth55: In order to build C++11, we need to set C++11 flag in CMakelists.txt.
I update this PR to add this change https://github.com/ros-perception/image_pipeline/pull/412/commits/cf0576a8436f2cfcd5c4f1343fe67dac7fac2d1a

image_view/src/nodes/image_saver.cpp
```diff
@@ -38,35 +38,128 @@
...
+  void callbackTimeoutTimer(const ros::TimerEvent&)
+  {
+    ROS_DEBUG("callback timeout timer");
+    unsubscribe();
+  }
+
+  bool callbackSave(std_srvs::Empty::Request &req,
+                    std_srvs::Empty::Response &res)
+  {
+    if (subscribing_)
```
JWhitleyWork, line 123: If the `std::lock_guard` above is to guard this variable, you should do it here as well. Locking the mutex only while writing doesn't really guard it because reading the variable here then becomes a non-atomic operation.

image_view/src/nodes/image_saver.cpp
```diff
@@ -38,35 +38,128 @@
...
+    {
+      save_timeout_timer_ = nh_.createTimer(
+          ros::Duration(save_timeout_), &Callbacks::callbackTimeoutTimer, this,
+          /* oneshot= */true);
+    }
+
+    if (wait_for_save_)
+    {
+      ros::Rate r(10);
+      while (ros::ok() && subscribing_)
```
JWhitleyWork, line 144: Same here. I recommend a `std::unique_lock` which can be `.lock()`'d and `.unlock()`'d.

JWhitleyWork (changes requested):
See comments.

JWhitleyWork:
Also, please update the launch files.

knorth55:
which launch file are you talking about?

image_view/src/nodes/image_saver.cpp
knorth55: Thanks, I updated.

JWhitleyWork:
Nevermind on the launch files. I guess there are no examples in `image_view`.

image_view/src/nodes/image_saver.cpp
JWhitleyWork: The `std::unique_lock` is created on line 122, which locks it automatically. It is then unlocked on line 128. However, it is not re-locked or unlocked around the read of `subscribing_` on line 144.

knorth55: This is because this while is waiting `subscribing_` to be updated by `subscribe()` function.

JWhitleyWork: Then this should use a `std::wait_condition` rather than a `while` loop which constantly polls the value. The current implementation is not thread-safe.

JWhitleyWork:
@knorth55 This has been hanging around for some time. Are you still able to work on it? If not, I can close the PR.

mikeferguson:
Closing due to lack of response

PR #417, opened 2019-05-10
Add a compressed_image_view app that subscribe to a sensor_msgs/CompressedImage directly

romainreignier:
Useful for node that directly publishes compressed image without image_transport

### DIFF
--- a/image_view/CMakeLists.txt
+++ b/image_view/CMakeLists.txt
@@ -63,6 +63,13 @@ target_link_libraries(image_view_exe ${catkin_LIBRARIES}
                                      ${Boost_LIBRARIES}
 )
 
+add_executable(compressed_image_view src/nodes/compressed_image_view.cpp)
+add_dependencies(compressed_image_view ${PROJECT_NAME}_gencfg)
+target_link_libraries(compressed_image_view ${catkin_LIBRARIES}
+                                            ${OpenCV_LIBRARIES}
+                                            ${Boost_LIBRARIES}
+)
+
 add_executable(disparity_view src/nodes/disparity_view.cpp)
 target_link_libraries(disparity_view ${catkin_LIBRARIES}
                                      ${OpenCV_LIBRARIES}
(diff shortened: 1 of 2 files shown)


SteveMacenski:
Question: Why not add this capability to the existing `image_view` node? It should just be a check if the word `compressed` `/namespace/camera/image/compressed` is around, then link the subscriber against that topic with the a compressed callback.

romainreignier:
@SteveMacenski Thanks for your quick reply.
Indeed, it is possible to add this directly in `image_view` and by-pass the `image_transport` subscriber.
I am ok with both solution, just tell me which one to choose.

SteveMacenski:
@romainreignier using the `Subscriber/CameraSubscriber` you can add a field of type ` TransportHints` which contains a `default_transport` which can be set to compressed. In this case, you'd just have to check if the topic ends with `/compressed` if so, make the hint contain the transport type appropriate and then register that callback. 

More info here: http://wiki.ros.org/image_transport

From how I read these docs (http://docs.ros.org/diamondback/api/image_transport/html/classimage__transport_1_1ImageTransport.html#ac871ee686e1d6767ec5c2eec9daf9399) you probably dont even need to use the fact that `/compressed` is the topic beyond the hints. The argument name `base_topic` makes me think that if you just tell it to use the hint of `compressed` that it will automatically deal with that other stuff for you. That should change this to about a 3 line change in the `main()` and potentially not even needing another callback (but probably)

romainreignier:
@SteveMacenski Sorry but I misunderstood your point.
My issue is to subscribe to a topic of type `sensor_msgs/CompressedImage` but not advertised by `image_transport` so the topic does not ends with `compressed`.

JWhitleyWork:
@romainreignier Is this still something you're interested in having merged? Please let me know or I can close the PR.

romainreignier:
I think that it would be nice to have this tool. But I understand if you don't want to maintain it because the use case does not seem too popular.

mikeferguson:
You should actually still be able to make image_view work for you by directly remapping the compressed topic name:
```
ros2 run image_view image_view --ros-args -r image/compressed:=your/other/topic/name -p image_transport:=compressed
```

mikeferguson:
This PR also targets melodic, which is EOL.

PR #418, opened 2019-05-13
Use SET_PRINCIPAL_POINT everywhere necessary

### LINKED ISSUE #324: Camera calibration introduces large change of aspect ratio
Camera calibration ignores the aspect ratio when it creates the projection matrix. This mainly affects cameras with a wide angle lens and a strong distortion. For some cameras, I got a difference of 30% in focal length for x/y in the projection_matrix. This translates to a rectified image with an incorrect aspect ratio. Since most computer vision algorithms (SIFT, KLT, etc.) cannot cope with such large changes in the aspect ratio, there should be an option to fix it.

The change of aspect ratio is introduced by cv2.getOptimalNewCameraMatrix. It seems it tries to find the largest rectangle that has no unknown/black areas, ignoring the aspect ratio of that rectangle.

Note this is not about the camera_matrix, which usually exhibits similar focal lengths in x/y. Also, the argument --fix-aspect-ratio allows to fix the aspect ratio in camera_matrix, but not in projection_matrix.

It would be interesting in what values other users get for the focal lengths in projection_matrix. Ideally, this could be fixed in OpenCV of course.

agdespopoulos:
Respect the --set-principal-point flag in camera_calibrator for another
function that accepts it.

Fixes #324.

### DIFF
--- a/camera_calibration/src/camera_calibration/calibrator.py
+++ b/camera_calibration/src/camera_calibration/calibrator.py
@@ -660,7 +660,10 @@ class MonoCalibrator(Calibrator):
         # NOTE: Prior to Electric, this code was broken such that we never actually saved the new
         # camera matrix. In effect, this enforced P = [K|0] for monocular cameras.
         # TODO: Verify that OpenCV #1199 gets applied (improved GetOptimalNewCameraMatrix)
-        ncm, _ = cv2.getOptimalNewCameraMatrix(self.intrinsics, self.distortion, self.size, a)
+        center_principle = (self.calib_flags & cv2.CALIB_FIX_PRINCIPAL_POINT) > 0
+        ncm, _ = cv2.getOptimalNewCameraMatrix(
+                self.intrinsics, self.distortion, self.size, a,
+                centerPrincipalPoint=center_principle)
         for j in range(3):
             for i in range(3):
                 self.P[j,i] = ncm[j, i]


camera_calibration/src/camera_calibration/calibrator.py
```diff
@@ -660,7 +660,10 @@ def set_alpha(self, a):
         # NOTE: Prior to Electric, this code was broken such that we never actually saved the new
         # camera matrix. In effect, this enforced P = [K|0] for monocular cameras.
         # TODO: Verify that OpenCV #1199 gets applied (improved GetOptimalNewCameraMatrix)
-        ncm, _ = cv2.getOptimalNewCameraMatrix(self.intrinsics, self.distortion, self.size, a)
+        center_principle = (self.calib_flags & cv2.CALIB_FIX_PRINCIPAL_POINT) > 0
```
SteveMacenski, line 663: isnt `cv2.CALIB_FIX_PRINCIPAL_POINT` a constant? What does adding this comparison do?

SteveMacenski (changes requested):

camera_calibration/src/camera_calibration/calibrator.py
agdespopoulos: Yes, it's a constant. `center_principle` is a boolean that indicates if `self.calib_flags` contains the `cv2.CALIB_FIX_PRINCIPAL_POINT` flag

SteveMacenski: I read this as `calib_flags and cv2.CALIB_...`, I'm not sure checking a `cv2` const makes sense - there's nothing here that compares if calib_flags has this cv2 constant contained with in it that I know of. 

Even if this is actually doing what you say it does, its very non-pythonic. This should be alot more clear as to what you're doing and I'll merge

JWhitleyWork:
@SteveMacenski I'm guessing this can be closed from lack of response.

SteveMacenski:
I suppose, yes

PR #423, opened 2019-06-25
Avoid crashing when the x or y offset is too large

lucasw:
Later should make it optional to create blank image where there is no source image to be had.

Also later need to respect the camera info roi of the incoming image, the crop coordinates should be relative to the uncropped incoming image, not the cropped one.

### DIFF
--- a/image_proc/src/nodelets/crop_decimate.cpp
+++ b/image_proc/src/nodelets/crop_decimate.cpp
@@ -195,8 +195,22 @@ void CropDecimateNodelet::imageCb(const sensor_msgs::ImageConstPtr& image_msg,
     config.height &= ~0x1;    
   }
 
+  // TODO(lucasw) should generate a black image where there isn't any source image,
+  // could be a dynamic reconfigure option
   int max_width = image_msg->width - config.x_offset;
+  if (max_width <= 0)
+  {
+    ROS_DEBUG_STREAM("x offset is outside the input image width: "
+        << image_msg->width << ", x offset: " << config.x_offset);
+    return;
+  }
   int max_height = image_msg->height - config.y_offset;
+  if (max_height <= 0)
+  {
+    ROS_DEBUG_STREAM("y offset is outside the input image height: "
+        << image_msg->height << ", y offset: " << config.y_offset);
+    return;
+  }
   int width = config.width;
   int height = config.height;
   if (width == 0 || width > max_width)
--- a/image_proc/src/nodelets/crop_decimate.cpp
+++ b/image_proc/src/nodelets/crop_decimate.cpp
@@ -195,6 +195,8 @@ void CropDecimateNodelet::imageCb(const sensor_msgs::ImageConstPtr& image_msg,
     config.height &= ~0x1;    
   }
 
+  // TODO(lucasw) need to incorporate the x_offset and y_offset within info_msg
+
   // TODO(lucasw) should generate a black image where there isn't any source image,
   // could be a dynamic reconfigure option
   int max_width = image_msg->width - config.x_offset;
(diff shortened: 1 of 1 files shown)


image_proc/src/nodelets/crop_decimate.cpp
```diff
@@ -195,8 +195,24 @@ void CropDecimateNodelet::imageCb(const sensor_msgs::ImageConstPtr& image_msg,
     config.height &= ~0x1;    
   }
 
+  // TODO(lucasw) need to incorporate the x_offset and y_offset within info_msg
+
+  // TODO(lucasw) should generate a black image where there isn't any source image,
+  // could be a dynamic reconfigure option
```
JWhitleyWork: Please remove these comments and create issues on this repo associated with these.

image_proc/src/nodelets/crop_decimate.cpp
```diff
@@ -195,8 +195,24 @@ void CropDecimateNodelet::imageCb(const sensor_msgs::ImageConstPtr& image_msg,
...
   int max_width = image_msg->width - config.x_offset;
+  if (max_width <= 0)
+  {
+    ROS_DEBUG_STREAM("x offset is outside the input image width: "
```
SteveMacenski: I think these should be warnings, not debugs. If this is happening there's something very wrong somewhere in the pipeline and it should be made aware to a user

SteveMacenski (changes requested):

image_proc/src/nodelets/crop_decimate.cpp
lucasw: The x offset can come straight from a dynamic reconfigure gui which doesn't respect the max width of the image- a flood of warnings isn't ideal when trying to dial in a number close to the edge of the true image width but somewhere in the middle of the slider range.  (Dynamic dynamic reconfigure would be nice there, set the max width and height and offsets based on the actual image instead of the cfg values)

SteveMacenski: Having a debug statement seems like it may make it harder to fix an issue when the image itself is the issue (changing cameras, widths are different, things are silently breaking) but that's a good point during the configuration process.

Maybe a `ROS_WARN_ONCE` or `THROTTLE`? Just something on screen to help someone see something's wrong, what do you think?

image_proc/src/nodelets/crop_decimate.cpp
lucasw: Done, #428 and #429

image_proc/src/nodelets/crop_decimate.cpp
lucasw: Throttling to once a second sounds fine, though even better may be to always warn when the max_width/height first goes negative, but then not warn again until it has gone positive and back to negative again- a couple of booleans would be used for that.

Maybe a zero size image should still be published instead of nothing at all?  Not sure if downstream nodes will handle those properly.

SteveMacenski: Throttling or your suggestion sounds fine. I would recommend to just keep it simple and throttle every 60 seconds. If it looks like its going to be a problem someone can open a ticket to reduce spam.

JWhitleyWork (approved):

SteveMacenski (approved):

SteveMacenski:
@JWhitleyAStuff I approved, but I would really like those warnings throttled. After merge I might just submit a new PR to do that

JWhitleyWork:
> @JWhitleyAStuff I approved, but I would really like those warnings throttled. After merge I might just submit a new PR to do that

Sorry for resolving the issue. I misread the code. I thought they had been. My bad.

SteveMacenski:
@JWhitleyAStuff if you're happy with it, merge away, I cant PR my updates until this is in

PR #431, opened 2019-07-17
Expand range for min_disparity and disparity_range.

mogumbo:
This provides a much greater limits for min_disparity and disparity_range in Disparity.cfg. I would have preferred to remove the limits completely, but that makes for ugly/useless slider bars in rqt_reconfigure (the text entry box besides the slider would remain usable).


SteveMacenski:
Works, did you test the extremes to make sure nothing breaks? (ei did you try -range, +range for them)

Confirm and I'll merge

mogumbo:
There are no error messages or crashes when exercising the full range of both sliders. As before, you can still produce useful point clouds or bad data depending on how you configure your camera and stereo_image_proc.

SteveMacenski:
image

PR #432, opened 2019-07-18
camera_calibration: Fix excessive CPU usage due to queue synchronization

valgur:
I was getting near 100% CPU usage with cameracalibrator.py even if no images (after then inital one) are being received and no actual processing is taking place.

This appeared to be caused by the queue synchronization code doing 
```python
while len(self.queue) == 0:
    time.sleep(0.1)
```
to wait for incoming images. This does not look too unreasonable, but replacing it with proper thread synchronization via locks fixed the excessive CPU usage.

Specifically, I replaced the deque with a slightly modified standard Queue that drops the oldest item instead of raising an exception when adding an item to an already full queue, just like deque. Queue also uses deque internally anyway, so in that sense the performance should be the same.

I also added a `--queue-size` parameter to allow adjustment of the max queue size. I found this useful when calibrating with images from a bag, where you want to use all available frames and set the playback rate as high as possible.

Relates to #112 and #121.

### DIFF
--- a/camera_calibration/nodes/cameracalibrator.py
+++ b/camera_calibration/nodes/cameracalibrator.py
@@ -68,6 +68,9 @@ def main():
     group.add_option("--no-service-check",
                      action="store_false", dest="service_check", default=True,
                      help="disable check for set_camera_info services at startup")
+    group.add_option("--queue-size",
+                     type="int", default=1,
+                     help="image queue size (default %default, set to 0 for unlimited)")
     parser.add_option_group(group)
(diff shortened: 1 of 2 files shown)


SteveMacenski (approved):
Looks good to me. I have no issues. Lets get a second approval and ship.

JWhitleyWork:
SHIP IT

JWhitleyWork (approved):

SteveMacenski:
(I do love some shipping graphics)

PR #341, opened 2018-05-19
distortion_coefficients size (cols) is not fixed, it depends on distortion mododel

k-okada:
discussed in #278, as discussed CameraInfo.msg, 
```
float64[] D
```
the size of D should not be fixed number (5)

c.f. http://docs.ros.org/api/sensor_msgs/html/msg/CameraInfo.html

NOTE : I haven't tested this code yet, someone?

PR #433, opened 2019-07-18
camera_calibration: Fix all-zero distortion coeffs returned for a rational_polynomial model

valgur:
Using ` --k-coefficients` greater than 3 with cameracalibrator.py currently returns a distortion coefficients array of all zeroes. I suppose this could be related to OpenCV version, which is 4.1.0 on my machine.

This is fixed by simply using the return values of `cv2.calibrateCamera()` instead of output variables to retrieve the distortion coefficients output.


camera_calibration/src/camera_calibration/calibrator.py
```diff
@@ -637,11 +637,14 @@ def cal_fromcorners(self, good):
...
         self.intrinsics[0,0] = 1.0
         self.intrinsics[1,1] = 1.0
-        cv2.calibrateCamera(
+        reproj_err, self.intrinsics, self.distortion, rvecs, tvecs = cv2.calibrateCamera(
                    opts, ipts,
                    self.size, self.intrinsics,
                    self.distortion,
                    flags = self.calib_flags)
+        # OpenCV returns > 8 coefficients (all zeros) when CALIB_RATIONAL_MODEL is set.
+        # These include, e.g. thin prism coefficients, which we are not interested in.
```
SteveMacenski: I see a self.distortion as input, output, and then remapped to [:8]. Lets clean this up to make it clear whats being stored or deleted. Only 1 thing should be stored as self.distortion.

camera_calibration/src/camera_calibration/calibrator.py
```diff
@@ -637,11 +637,14 @@ def cal_fromcorners(self, good):
         # If FIX_ASPECT_RATIO flag set, enforce focal lengths have 1/1 ratio
         self.intrinsics[0,0] = 1.0
         self.intrinsics[1,1] = 1.0
-        cv2.calibrateCamera(
+        reproj_err, self.intrinsics, self.distortion, rvecs, tvecs = cv2.calibrateCamera(
```
SteveMacenski: I’m slightly unclear why we expect the distortion params on return to be different from those added to the self.distortion implicitly. Is this the result of a bug in opencv? If this is true the API reads to me that those are optional parameters you can simply remove and just use the direct return copies.

SteveMacenski (changes requested):
some clarity needed + a link to something documenting the bug this is resolving would be great

valgur:
> some clarity needed + a link to something documenting the bug this is resolving would be great

I have not created an issue regarding this. Should I create one?

camera_calibration/src/camera_calibration/calibrator.py
valgur: As I understand it, the Python API of OpenCV does not intend the function arguments that are used for output in C++ to be used as such in Python and the return values should be used in their place instead. Considering this, I think the fact that the output variables returned something reasonable earlier was just a lucky coincidence.

Passing all-zero `self.distortion` as input is redundant, indeed, as it matches the default behavior. I'll remove it from the arguments.

camera_calibration/src/camera_calibration/calibrator.py
valgur: I updated the code to avoid using `self.intrinsics` and `self.distortion` as inputs for clarity.

valgur:
~~Might want to merge #341 as well, if this one gets approved.~~ That PR is buggy (tries to concatenate an int to str), so I included that fix here instead.

SteveMacenski:
>  I suppose this could be related to OpenCV version

I thought you had tracked down this is an OpenCV issue, if it is, you should definitely file a bug with them.

valgur:
I took a closer look at the `calibrateCamera()` behavior regarding `distCoeffs`.
The `distCoeffs` *returned* from the function are always correct. Using `distCoeffs` as an *input-output array* only outputs the correct values when the shape and dtype of the input matches the dtype and shape of the expected output from the function, though. Looks like the array gets re-allocated in C++ and the reference to the input numpy array discarded.

I suppose this could be considered an OpenCV bug, but since the more typical and reliable way of retrieving the output via a return value works, I'm not sure it's worth creating a bug report for this.

SteveMacenski:
Fair enough. This looks good to me, can you give me a "yes I have tested this sufficiently to my liking to be comfortable merging this" statement? I don't have the hardware at this moment on travel to evalulate.

Then merging is OK by me with a second approval from probably @JWhitleyAStuff

SteveMacenski (approved):

valgur:
Sure - yes, I would be comfortable merging this. The changes are quite straightforward and the unit tests pass on both my machine and the CI. I also checked the stereo calibration code for similar issues and did not find anything there.

SteveMacenski:
Ok, did you test this on an actual camera in real life? CI is good, but hardware is great.

valgur:
Yep, I have tested and used it on real camera data. Both `--k-coefficients` 3 and >3 are working as expected.

JWhitleyWork:
I'm honestly not familiar enough with what's going on in this PR to provide my stamp of approval so I'll leave this one up to @stevemacenski. Merge at your leisure.

SteveMacenski:
@JWhitleyAStuff even without being familiar with the thing happening under the hood, I think its clear to show that instead of an implicit-output, its having a hard output and for whatever reason in the Python wrapper for OpenCV that solves this issue (which a little different data representation in numpy) I dont see anything here that changes an algorithm or usage so I'm OK merging it if its been tested to work on hardware such that all the places with the new dimensions are correct

PR #437, opened 2019-07-23
camera_calibration: Make sure 'calibrate' button works even if not receiving images anymore

valgur:
Mouse events are currently only being checked and the buttons being enabled/disabled when the display gets redrawn on received images.
This change allows the calibration to be saved even when there are no more images in the queue, e.g. when a bag playback has finished.

### DIFF
--- a/camera_calibration/src/camera_calibration/camera_calibrator.py
+++ b/camera_calibration/src/camera_calibration/camera_calibrator.py
@@ -40,6 +40,7 @@ import rospy
 import sensor_msgs.msg
 import sensor_msgs.srv
 import threading
+import time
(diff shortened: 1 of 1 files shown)


SteveMacenski:
Sounds good, I'm going to need to test this to verify. I'm curious though, what's the button you're trying to mess with after its done?

valgur:
> what's the button you're trying to mess with after its done?

The "calibrate" and "save" ones. The "save" button only becomes enabled after calibration is complete, but does not currently do so if the image queue is empty.

SteveMacenski (approved):
@JWhitleyAStuff i approve. You?

JWhitleyWork (approved):
LGTM

PR #442, opened 2019-07-31
Add image_proc with GPU support

DavidTorresOcana:
* image_proc_tegra(https://github.com/DavidTorresOcana/image_pipeline/tree/indigo/image_proc_tegra): 
    - A camera rectification package for camera rectification with OpenCV >3 and CUDA support. Tested in Tegra cores. Faster than image_proc since it uses CUDA for rectification

   - It also implements a Nodelet for Fisheye cameras rectification, if calibrated with camera_calibration_fisheye(https://github.com/DavidTorresOcana/image_pipeline/tree/indigo/camera_calibration_fisheye). It uses OpenCV >3  with CUDA suport

### DIFF
--- a/image_proc_tegra/CMakeLists.txt
+++ b/image_proc_tegra/CMakeLists.txt
@@ -1,0 +1,42 @@
+cmake_minimum_required(VERSION 2.8.3)
+project(image_proc_tegra)
+set (CMAKE_CXX_STANDARD 11)
+
+find_package(CUDA REQUIRED)
+find_package(catkin REQUIRED COMPONENTS
+  cv_bridge
+  image_transport
+  image_geometry
+  OpenCV
+  sensor_msgs
+  std_msgs
+  roscpp
+  nodelet  
+)
+
+catkin_package(
+  CATKIN_DEPENDS image_geometry roscpp sensor_msgs
+  DEPENDS OpenCV
+  LIBRARIES ${PROJECT_NAME}
+)
+
+include_directories(SYSTEM 
+${catkin_INCLUDE_DIRS} 
+${OpenCV_INCLUDE_DIRS}
+${CUDA_INCLUDE_DIRS}
+)
+include_directories(include)
+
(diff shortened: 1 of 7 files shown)


SteveMacenski:
@DavidTorresOcana this looks fine to be if you can retarget to melodic, get CI to pass.

Why do you shutdown the subscriber for camera info? You should use a image transport subscriber so your CB gets the info that's synchronized

DavidTorresOcana:
> @DavidTorresOcana this looks fine to be if you can retarget to melodic, get CI to pass.
> 
> Why do you shutdown the subscriber for camera info? You should use a image transport subscriber so your CB gets the info that's synchronized

The reason of needing to employ GPU arise in embedded platforms, where limited CPU is available, hence you want to minimize computation.
That is why the info CB is shutdown

That being said, I see this package to target a very small niche and may be a pain to integrate in official releases. It depends on CUDA and OpenCV compiled with CUDA, which may not be very standard.
Hence, unless you think otherwise, we can close this.

SteveMacenski:
@DavidTorresOcana I have no concerns on merging it once you pass CI. I don't see any rosdep keys for cuda so you have a least a little bit to do.

I'd **strongly** suggest changing to use image transport, that way you can use compressed image streams which is typical for things in this package. It also synchronizes the camera info and image topics for you. It's not much overhead and its the typical way of doing this. It also allows for updating the camera parameters if they change

JWhitleyWork:
@DavidTorresOcana I would suggest changing the name of this to `_cuda` instead of `_tegra` since CUDA is applicable to more than just Tegra-based GPUs. I can also test it on a GTX 2060 once it's ready. Also, please rebase on `melodic` so you can get the new CI tests.

JWhitleyWork:
@DavidTorresOcana Ping for feedback. Also, please rebase on the current `melodic` branch to pick up a new CI pipeline.

DavidTorresOcana:
I am sorry but I am unable to maintain this branch and continue this PR.

Anybody interested is welcome to get the branch and continue the work.

Issue #403, opened 2019-04-19
Update image_pipeline to use new ros2 composition model


mjcarroll:
In `dashing`, we have updated how ROS2's components work.  This allows developers to write a `rclcpp::Node` that can be used either in standalone or composed into a larger executable, and defer that decision until runtime.  It is essentially a replacement for ROS1's `nodelet` concept, but without having to do special implementations.

I believe that `image_pipeline` would be a good candidate to take advantage of this new model, as it closely mirrors the way that `nodelets` were used in ROS1.

References:
* https://github.com/ros2/rclcpp/tree/master/rclcpp_components
* https://github.com/ros2/demos/tree/master/composition
* Documentation (WIP) https://github.com/ros2/ros2_documentation/pull/168
* Composable node launch actions: https://github.com/ros2/launch_ros/pull/9

JWhitleyWork:
While Dashing has "technically" been released and I really like this feature (been looking forward to it for a long time), I don't think it's well-supported enough yet (e.g. no official Docker images, `cv_bridge` not yet released, `image_geometry` not yet released, etc.) to target it in this repo. Long-term, though, this makes a lot of sense.

SteveMacenski:
I disagree, I heard William talk about this at an event last week and its particularly useful for this repo and other sensor pipeline metapackages. Nodelets don't exist in ROS2 and this gives you the zero copy in-process capabilities. 
None of the ROS2 releases are LTS and should be considered deprecated when a new one comes out every 6 months and there's demos working today.

SteveMacenski:
@mjcarroll are you imagining image_proc or all the image pipeline nodes?

JWhitleyWork:
I've been doing a lot of porting recently and my earlier statement no longer applies. I think it makes sense to do all of them.

SteveMacenski:
OK.

What are your porting plans / actions?

JWhitleyWork:
Luckily, the port isn't that complicated. We just need to make sure that:

1. Each node is a separate class which inherits from `rclcpp::Node` and takes an `rclcpp::NodeOptions` object.
2. An executable is generated for each `component` (this is just an option in the call to `rclcpp_components_register_node`).
3. Each node is registered as a `component` using the appropriate CMake macros.
4. Some example pipelines are provided either as launch files or binaries which combine several components.

SteveMacenski:
Got it, I thought you meant you were in the middle of some of these ports.

[...]

PR #443, opened 2019-08-04
ROS2 Image proc refactoring using components and added launch file

klintan:
Created a launch file for image_proc. I would love some input on how to get the parameter change callback to fire when parameters are set from a launch file ? 

Calling `ros2 run image_proc image_proc --camera_namespace /camera` runs everything fine. 
However running the launch file it seems everything starts ok (nodes etc) but the topics are not created, and my suspicion is that the callback is never fired (well since that is lambda that creates the publisher etc).

Maybe it needs to be done a different way to get it to work with the launch file?

edit: Ok tracked this down, will need to rewrite the `image_proc` package a bit, I'll provide a PR as soon as that is done, after that hopefully someone could review and merge this as well.


klintan:
Did some refactoring to be able to get it to work using components and ros launch file. 

The way it was build it needed a dynamic `ros2 param set` command for it to start subscription. I'm fine with that, however I haven't found a way to do this in the launch file ? (Setting the parameters on start as per usual will not work, you actually need the dynamic reconfigure action equivalent of `ros2 param set ...` 

I'm open for complete changes to my approach, since I don't think it might be the best one, so I would love some major feedback.

However, using this PR as of now is the only way to run it from a launch file.

edit: I would also like to refactor this to instead of a camera_namespace parameter, get this from the regular `__ns:=<namespace>` functionality, it messes up downstream stuff and launch files when it does not behaves like the standard. Let me know if anyone has any objections to this.

klintan:
This would also partially address this created issue https://github.com/ros-perception/image_pipeline/issues/403

JWhitleyWork:
Makes sense to me.

SteveMacenski:
This failed the build farm due to a missing dependency before compile time http://build.ros2.org/job/Cdev__image_pipeline__ubuntu_bionic_amd64/6/console

image_proc/src/image_proc.cpp
```diff
@@ -75,33 +54,21 @@ int main(int argc, char * argv[])
 
   rclcpp::init(argc, argv);
 
-  std::string camera_namespace;
-  if (rcutils_cli_option_exist(argv, argv + argc, "--camera_namespace")) {
```
SteveMacenski, line 79: Why did we remove this available to remap on commandline?

image_proc/src/image_proc.cpp
```diff
@@ -75,33 +54,21 @@ int main(int argc, char * argv[])
...
   auto debayer_node = std::make_shared<image_proc::DebayerNode>(options);
-  debayer_node->declare_parameter("camera_namespace", camera_namespace);
-  
 
   // Rectify nodelet, image_mono -> image_rect
   auto rectify_mono_node = std::make_shared<image_proc::RectifyNode>(options);
-  rectify_mono_node->declare_parameter("camera_namespace", camera_namespace);
-  rectify_mono_node->declare_parameter("image_mono", "/image_mono");
-  
+  rectify_mono_node->set_parameter(rclcpp::Parameter("image_topic", "/image_mono"));
```
SteveMacenski, line 67: Please revert this parameter ASAP, we should not be casually changing parameter names, this could make a huge issue across a lot of peoples' software after release

SteveMacenski:
~We really need to set up PR builders on this repo for ROS2~ wait, we have that, why didn't it run here?

JWhitleyWork:
> ~We really need to set up PR builders on this repo for ROS2~ wait, we have that, why didn't it run here?

It did. There are currently 2 PR builds happening: One on CircleCI (which passed) and one on the ROS Build Farm (which did not pass). However, the one on the Build Farm is targeted at Crystal and there are now Dashing features in this branch which will cause the Crystal build to always fail. The reason that the CircleCI job did not catch it is because it uses `ros-dashing-perception` as its base, which likely already had the dependency. I will create a new CI job which builds from a plain ROS image.

image_proc/src/image_proc.cpp
JWhitleyWork: I thought the standard was using ROS parameters rather than command-line options?

image_proc/src/image_proc.cpp
JWhitleyWork: This was my fault. Should have caught that.

SteveMacenski:
@klintan can you please resubmit the PR broken up more. You do substantially more here than just refactor to use components, you change key parameter names as well as remove namespacing commandline options (and then some initialization/setup associated with them).

I'd recommend it being broken into 2-3 PRs. Mega-PRs are always bound to create regressions and bite sizes pieces are easier to digest both for reviewing and for regression hunting down the line. I know this PR doesnt seem that big, but there's enough changes that I can't reasonably just pick up a camera and in 10 minutes ensure nothing is broken

image_proc/src/image_proc.cpp
klintan: That was my thinking as well, to normalize it to work like other ROS2-packages and use the `__ns:=/namespace` argument.

image_proc/src/image_proc.cpp
klintan: Ok will try to keep them.

klintan:
> @klintan can you please resubmit the PR broken up more. You do substantially more here than just refactor to use components, you change key parameter names as well as remove namespacing commandline options (and then some initialization/setup associated with them).
> 
> I'd recommend it being broken into 2-3 PRs. Mega-PRs are always bound to create regressions and bite sizes pieces are easier to digest both for reviewing and for regression hunting down the line. I know this PR doesnt seem that big, but there's enough changes that I can't reasonably just pick up a camera and in 10 minutes ensure nothing is broken

Ok I'll split it up, thanks for reviewing.

image_proc/src/image_proc.cpp
klintan: @SteveMacenski I fully understand we don't want to change too many parameters/arguments between the different version to make it easier for ppl. However before I go about changing stuff for this and put it additional broken up PRs, I'd just like to know the reason for having "camera_namespace parameter, and not using the `__ns:=` functionality instead ?

PR #453, opened 2019-09-10
Fixed double free crash, removed deprecation warnings.

luca-della-vedova:
Hi all,

There was a series of issues with this node, out of the box it just crashed because the parameters are not declared, output of command `ros2 run image_rotate image_rotate image:=/image_raw`:

`terminate called after throwing an instance of 'rclcpp::exceptions::ParameterNotDeclaredException'
  what():  parameter 'target_frame_id' cannot be set because it was not declared`

Fixed through changing the API to the latest declare_parameter (same as #452 ), nice side effect is that all the compile warnings are gone.
After this fix:

`double free or corruption (out)`

Which turns out is due to the construction of a temporary `rclcpp::Node::SharedPtr(this)` that is freed upon leaving the scope, fixed through passing a raw `*this`.

Works beautifully now (screenshot of a noise image I had on my PC):

image

### DIFF
--- a/image_rotate/src/image_rotate_node.cpp
+++ b/image_rotate/src/image_rotate_node.cpp
@@ -65,9 +65,9 @@ ImageRotateNode::ImageRotateNode()
   config_.input_frame_id = this->declare_parameter("input_frame_id", std::string(""));
   config_.use_camera_info = this->declare_parameter("use_camera_info", true);
   config_.max_angular_rate = this->declare_parameter("max_angular_rate",
-    static_cast<double>(10));
+      static_cast<double>(10));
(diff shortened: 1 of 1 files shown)


SteveMacenski (approved):
I love your polar bear in a snow storm!

SteveMacenski:
Get CI to pass on dashing and I’ll merge. Looks like uncrustify

luca-della-vedova:
> I love your polar bear in a snow storm!

That's was one of the first images I got while developing a low level camera driver, I like to believe it was kinda working.

Fixed uncrustify, now pass all tests!

PR #457, opened 2019-09-16
ROS2: added launch file and changes necessary for it to work

klintan:
From this PR https://github.com/ros-perception/image_pipeline/pull/443

- Added launch file using components
- Added necessary changes for it to work
- All parameters kept,
- Camera namespace cmd-line argument still available and same behavior.

@SteveMacenski and @JWhitleyAStuff I would love some feedback on this, and since you saw the last one I think it makes sense for you to review this again.


JWhitleyWork (approved):
I'm not familiar enough yet with composable nodes to definitively say much but it looks good as far as I'm concerned. I'll wait for @SteveMacenski to review before merging.

SteveMacenski:
There’s a little too much pending on the ros2 side here for me to merge anything. I’m working on allocating time post-roscon to pull away from my other work and refocus my attention here. The ROS2 branches have gotten out of wack with ROS1 & there’s been alot of non-scrutinized changes that need to be looked at again

SteveMacenski:
Things I know I’m not going to like when I review it
- any static variables. Namespace, amongst others should be runtime configurable via the launch configuration options
- debug main should be removed unless there’s a test using it
- are these composible nodes? I don’t recall that

image_proc/src/image_proc.cpp
```diff
@@ -89,19 +75,19 @@ int main(int argc, char * argv[])
 
   // Debayer nodelet, image_raw -> image_mono, image_color
   auto debayer_node = std::make_shared<image_proc::DebayerNode>(options);
-  debayer_node->declare_parameter("camera_namespace", camera_namespace);
+  debayer_node->set_parameter(rclcpp::Parameter("camera_namespace", camera_namespace));
```
SteveMacenski, line 78: Its not clear to me what your refactor buys us vs the old changes of declaring the parameters outside the node itself and here in the main()

SteveMacenski (changes requested):
More of a question, but if you've tested this it looks good to me.

There's the open question above about runtime configurations to change the namespace though. See `LaunchConfiguration`

image_proc/src/image_proc.cpp
klintan: I'm not sure how one would do it if we want to use it as a component. Before, trying to use it as a component, the callback was never triggered and the camera_namespace would not be declared.

image_proc/src/rectify.cpp
```diff
@@ -77,13 +77,29 @@ RectifyNode::RectifyNode(const rclcpp::NodeOptions& options)
       }
       return result;
     };
+
+  // Declare parameters
+  this->declare_parameter("camera_namespace", rclcpp::ParameterValue());
```
SteveMacenski, line 82: Can we set these as logical defaults? image_color, image_mono, camera_namespace being camera, etc. Right now if you basically dont use the launch file it'll never work

image_proc/src/rectify.cpp
```diff
@@ -68,7 +68,7 @@ RectifyNode::RectifyNode(const rclcpp::NodeOptions& options)
         if (parameter.get_name() == "image_color") {
           image_topic = camera_namespace_ + parameter.as_string();
           image_rect = camera_namespace_ + "/image_rect_color";
-          RCLCPP_INFO(get_logger(), "image_topic: %s, image_rect: %s", image_topic.c_str(), image_rect.c_str());
+          RCLCPP_INFO(get_logger(), "image_topic: %s, Publish to topic image_rect: %s", image_topic.c_str(), image_rect.c_str());
```
SteveMacenski, line 71: I think this should just go back to the way it was

SteveMacenski (changes requested):

image_proc/src/image_proc.cpp
SteveMacenski: mhm. So I'm trying to rectify 2 things, both the use in the lanuch file that you have so you can do commandline remapping and the use of the nodes as individual components. Right now if you don't use this main() for the library, you really cant use the components themselves because the declares don't set logical defaults and they don't try to get the parameter themselves. 

You can set arguments through the NodeOptions class http://docs.ros2.org/dashing/api/rclcpp/classrclcpp_1_1NodeOptions.html#ab0c61163fb466ceb980a001bdce2fb7a

JWhitleyWork:
@klintan Any updates? Also needs rebase.

klintan:
> @klintan Any updates? Also needs rebase.

Yes sorry I moved on to other things and sort of forgot about this one, I'll try to take a look this weekend and rebase + fix feedback. Might be a better way to do it now.

JWhitleyWork:
Ping @klintan

klintan:
Sorry for the delay.

I'll try to get to this during the holidays :) 

Thanks for reminding me @JWhitleyWork

pmusau17:
Ping @klintan

clalancette:
I'm going to close this one out since there has been no response in a long time.  Please feel free to reopen if you are planning on updating this PR.

PR #463, opened 2019-09-24
Try catch around cvtColor to avoid demosaicing src.empty() error

lucasw:
I've encountered a cvtColor opencv exception a few times, my camera driver is occasionally feeding bad images in but I'd rather the nodelets survive this than not handle the exception here.


SteveMacenski (approved):

SteveMacenski:
totally makes sense to me, I'll merge after CI

SteveMacenski:
Fastest PR -> merge I've seen in awhile :-) go circle CI

lucasw:
Can this be backported to kinetic?  I can make a new PR if that helps, it looks like the same change can be used as is.

PR #465, opened 2019-10-01
ROS2: Added resize component

klintan:
Created similar to the crop component, so the reviews could probably be consolidated somewhat.

### DIFF
--- a/image_proc/CMakeLists.txt
+++ b/image_proc/CMakeLists.txt
@@ -17,7 +17,7 @@ find_package(rclcpp_components REQUIRED)
 find_package(cv_bridge REQUIRED)
 find_package(image_transport REQUIRED)
 find_package(sensor_msgs REQUIRED)
-find_package(OpenCV REQUIRED)
+find_package(OpenCV 3 REQUIRED)
 find_package(image_geometry REQUIRED)
 
 include_directories(
@@ -94,16 +94,42 @@ ament_target_dependencies(image_proc
 )
 
 install(TARGETS
+  image_proc
+  DESTINATION lib/${PROJECT_NAME}
+)
+
+install(
+  TARGETS resize
+  EXPORT export_resize
+  ARCHIVE DESTINATION lib
+  LIBRARY DESTINATION lib
+  RUNTIME DESTINATION bin
+)
+install(
+  TARGETS debayer
+  EXPORT export_debayer
+  ARCHIVE DESTINATION lib
+  LIBRARY DESTINATION lib
+  RUNTIME DESTINATION bin
+)
+install(
+  TARGETS rectify
+  EXPORT export_rectify
+  ARCHIVE DESTINATION lib
+  LIBRARY DESTINATION lib
+  RUNTIME DESTINATION bin
+)
+
+ament_export_interfaces(
+  export_resize HAS_LIBRARY_TARGET
+  export_debayer HAS_LIBRARY_TARGET
+  export_rectify HAS_LIBRARY_TARGET
+)
+
+ament_export_libraries(
   resize
   debayer
   rectify
-  ARCHIVE DESTINATION lib
-  RUNTIME DESTINATION bin
-  LIBRARY DESTINATION lib
 )
-
-install(TARGETS
-        image_proc
-        DESTINATION lib/${PROJECT_NAME})
 
 ament_package()
--- a/image_proc/CMakeLists.txt
+++ b/image_proc/CMakeLists.txt
@@ -20,7 +20,20 @@ find_package(sensor_msgs REQUIRED)
 find_package(OpenCV REQUIRED)
 find_package(image_geometry REQUIRED)
 
-include_directories(include)
+include_directories(
+  include
+)
+
+set(dependencies
+  image_geometry
+  sensor_msgs
+  cv_bridge
+  class_loader
+  image_transport
+  rclcpp
+  rclcpp_components
+  rcutils
+)
 
 add_library(rectify SHARED
   src/rectify.cpp)
@@ -28,14 +41,8 @@ target_compile_definitions(rectify
   PRIVATE "COMPOSITION_BUILDING_DLL")
 
 ament_target_dependencies(rectify
-  "image_geometry"
-  "sensor_msgs"
-  "cv_bridge"
-  "class_loader"
-  "image_transport"
-  "rclcpp"
-  "rclcpp_components"
-  )
+  ${dependencies}
+)
   
 rclcpp_components_register_nodes(rectify "image_proc::RectifyNode")
 set(node_plugins "${node_plugins}image_proc::RectifyNode;$<TARGET_FILE:rectify>\n")
@@ -49,13 +56,8 @@ target_compile_definitions(debayer
   PRIVATE "COMPOSITION_BUILDING_DLL")
 
 ament_target_dependencies(debayer
-  "sensor_msgs"
-  "cv_bridge"
-  "class_loader"
-  "image_transport"
-  "rclcpp"
-  "rclcpp_components"
-  )
+  ${dependencies}
+)
 
 rclcpp_components_register_nodes(debayer "image_proc::DebayerNode")
 set(node_plugins "${node_plugins}image_proc::DebayerNode;$<TARGET_FILE:debayer>\n")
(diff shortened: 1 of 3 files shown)


image_proc/CMakeLists.txt
```diff
@@ -60,6 +60,25 @@ ament_target_dependencies(debayer
...
 set(node_plugins "${node_plugins}image_proc::DebayerNode;$<TARGET_FILE:debayer>\n")
 
+add_library(resize SHARED
+  src/resize.cpp
+  )
+
+target_compile_definitions(resize
+  PRIVATE "COMPOSITION_BUILDING_DLL")
+
+ament_target_dependencies(resize
```
SteveMacenski, line 72: I've pointed to this before, but please do 

`set(dependencies [list, of, dependencies])`

and then use `${dependencies}` as needed. I don't want several long lists of packages that can be easily forgotten to be updated

image_proc/src/resize.cpp
```diff
@@ -0,0 +1,179 @@
...
+        }
+      }
+
+      image_topic_ = camera_namespace_ + "/image";
+      camera_info_topic_ = camera_namespace_ + "/camera_info";
+      connectCb();
+
+      std::lock_guard<std::mutex> lock(connect_mutex_);
+      pub_image_ = image_transport::create_publisher(this, image_topic_ + "/resize");
+      pub_info_ = this->create_publisher<sensor_msgs::msg::CameraInfo>(camera_info_topic_ + "/resize", rclcpp::QoS(1));
```
SteveMacenski: Given this is a sensor stream, don't you want to use the default sensor QoS?

image_proc/src/resize.cpp
```diff
@@ -0,0 +1,179 @@
...
+#include <image_transport/image_transport.h>
+
+#include "resize.hpp"
+
+namespace image_proc
+{
+ResizeNode::ResizeNode(const rclcpp::NodeOptions & options) : rclcpp::Node("ResizeNode", options)
+{
+
+   auto parameter_change_cb =
```
SteveMacenski, line 49: This is... interesting. Did you pull this from somewhere? I like it, just curious

image_proc/src/resize.cpp
```diff
@@ -0,0 +1,179 @@
...
+};
+
+// Handles (un)subscribing when clients (un)subscribe
+void ResizeNode::connectCb()
+{
+  std::lock_guard<std::mutex> lock(connect_mutex_);
+  
+  if (!sub_image_)
+  {
+    std::cout << "subscribe sub_image_" << std::endl;
```
SteveMacenski: Use RCLCPP logging if not a debug article?

image_proc/src/resize.cpp
```diff
@@ -0,0 +1,179 @@
...
+    sub_image_ = image_transport::create_subscription(this, image_topic_, std::bind(&ResizeNode::imageCb, this, std::placeholders::_1), "raw");
+  }
+  
+  if (!sub_info_)
+  {
+    std::cout << "subscribe sub_info_" << std::endl;
+    sub_info_ = this->create_subscription<sensor_msgs::msg::CameraInfo>(camera_info_topic_, rclcpp::QoS(1), std::bind(&ResizeNode::infoCb, this, std::placeholders::_1));
```
SteveMacenski: Same with the sensor QoS

image_proc/src/resize.cpp
```diff
@@ -0,0 +1,179 @@
...
+{
+  sensor_msgs::msg::CameraInfo::SharedPtr dst_info_msg = info_msg;
+  
+  double scale_y;
+  double scale_x;
+  if (use_scale_)
+  {
+    scale_y = scale_height_;
+    scale_x = scale_width_;
+    dst_info_msg->height = static_cast<int>(info_msg->height * scale_height_);
```
SteveMacenski, line 116: I appreciate your care with casting

image_proc/src/resize.cpp
```diff
@@ -0,0 +1,179 @@
...
+}
+
+void ResizeNode::imageCb(const sensor_msgs::msg::Image::ConstSharedPtr& image_msg)
+{
+  cv_bridge::CvImagePtr cv_ptr;
+  try
+  {
+    cv_ptr = cv_bridge::toCvCopy(image_msg);
+  }
+  catch (cv_bridge::Exception& e)
```
SteveMacenski: `cv_bridge::Exception & e`

SteveMacenski (changes requested):
Looks good, a few small changes and I'm cool merging this

image_proc/src/resize.cpp
klintan: Yepp I'll just remove it, some debugging artifacts.

image_proc/src/resize.cpp
klintan: Thanks, but can't take credit for that, was in the ROS1 code. :)

image_proc/src/resize.cpp
klintan: This was in the original implementation of the image_proc I think @ahuizxc wrote the lambda function for parameter change. :)

klintan:
I fixed all (hopefully) comments, let me know if there is anything else that needs changing or looking over.

SteveMacenski (approved):
LGTM thanks for the quick turn around @JWhitleyAStuff thoughts?

image_proc/include/resize.hpp
```diff
@@ -0,0 +1,83 @@
...
+#include <string>
+#include <vector>
+#include <thread>
+#include <rclcpp/rclcpp.hpp>
+#include <image_transport/image_transport.h>
+
+#include <ament_index_cpp/get_resource.hpp>
+#include "rclcpp/rclcpp.hpp"
+#ifndef IMAGE_PROC_RESIZE_HPP
+#define IMAGE_PROC_RESIZE_HPP
```
JWhitleyWork: Can you please move the header guard above the `#include`s so they don't get included unnecessarily?

image_proc/CMakeLists.txt
```diff
@@ -49,20 +56,30 @@ target_compile_definitions(debayer
...
+  ${dependencies}
+)
+
+rclcpp_components_register_nodes(resize "image_proc::ResizeNode")
+set(node_plugins "${node_plugins}image_proc::ResizeNode;$<TARGET_FILE:resize>\n")
```
JWhitleyWork, line 77: To follow recommended best-practices for exporting components as shared libraries, the following should be added (see [this post](https://discourse.ros.org/t/ament-best-practice-for-sharing-libraries/3602) for more details):

```
ament_export_interfaces(
  export_resize HAS_LIBRARY_TARGET
  export_debayer HAS_LIBRARY_TARGET
  export_rectify HAS_LIBRARY_TARGET
)
ament_export_libraries(
  resize
  debayer
  rectify
)
```

And this would replace the `install` command below:

```
install(
  TARGETS resize
  EXPORT export_resize
  ARCHIVE DESTINATION lib
  LIBRARY DESTINATION lib
  RUNTIME DESTINATION bin
)
install(
  TARGETS debayer
  EXPORT export_debayer
  ARCHIVE DESTINATION lib
  LIBRARY DESTINATION lib
  RUNTIME DESTINATION bin
)
install(
  TARGETS rectify
  EXPORT export_rectify
  ARCHIVE DESTINATION lib
  LIBRARY DESTINATION lib
  RUNTIME DESTINATION bin
)
```

The advantage of exporting the targets/interfaces as above instead of how it's being done now is that downstream packages can link against the namespaced interface (e.g. `image_proc::ResizeNode`) instead of the library name (e.g. `resize`). It's also the recommended way going forward.

JWhitleyWork (changes requested):
See recommended changes.

I also think we should be applying all of the linters but that should be done in another PR.

SteveMacenski:
I also haven't personally tested this, so @klintan have you gone through all the major options and makes sure it all works as intended?

klintan:
> I also haven't personally tested this, so @klintan have you gone through all the major options and makes sure it all works as intended?

I'll fix the last comments and go through all of the options and test them again.

SteveMacenski:
@klintan any update?

klintan:
> @klintan any update?

Sorry I've been traveling so I haven't had much time. I'll try to get it done by tomorrow.

image_proc/CMakeLists.txt
klintan: Awesome thanks, hadn't seen that approach before.

klintan:
> @klintan any update?

Ok should be updated now and I tested it out using all of the different parameters, and everything seemed to work. 

I added `3` to the `find_packages(OpenCV)` let me know if you want me to remove that, if not I think this might be ready for a merge :)

SteveMacenski:
Toggling CI

SteveMacenski:
I'd remove the `3` unless theres a compelling reason for it

@JWhitleyAStuff does this satisfy your concerns?

SteveMacenski (approved):

JWhitleyWork (approved):

SteveMacenski:
@klintan please do another PR removing the `3`

PR #471, opened 2019-11-03
ROS2 port of crop decimate in image proc

SteveMacenski:
@klintan please take a look and note a few things. In particular the use of image transport camera specific publishers and subscribers. They let you get the camera info and message synchronized. This is important for RVIZ as well as future subscribers needing camera info and data together. I updated the resize component for this as well (on top of the sub counter)

@mjcarroll can you take a look? Can't well nit-pick my own PR to death. Edit: Josh came to the rescue. 

Wildly untested, but something's better than nothing. @klintan if you can test this the way you're testing your other stuff I'd appreciate the thumbs up on it working as you expect. Still... no... cameras...

### DIFF
--- a/image_proc/CMakeLists.txt
+++ b/image_proc/CMakeLists.txt
@@ -76,6 +76,20 @@ ament_target_dependencies(resize
 rclcpp_components_register_nodes(resize "image_proc::ResizeNode")
 set(node_plugins "${node_plugins}image_proc::ResizeNode;$<TARGET_FILE:resize>\n")
 
+add_library(crop_decimate SHARED
+  src/crop_decimate.cpp
+  )
+
+target_compile_definitions(crop_decimate
+  PRIVATE "COMPOSITION_BUILDING_DLL")
+
+ament_target_dependencies(crop_decimate
+  ${dependencies}
+)
+
+rclcpp_components_register_nodes(crop_decimate "image_proc::CropDecimateNode")
+set(node_plugins "${node_plugins}image_proc::CropDecimateNode;$<TARGET_FILE:crop_decimate>\n")
+
 add_executable(image_proc
                src/image_proc.cpp
 )
@@ -106,6 +120,13 @@ install(
   RUNTIME DESTINATION bin
 )
 install(
+  TARGETS crop_decimate
+  EXPORT export_resize
+  ARCHIVE DESTINATION lib
+  LIBRARY DESTINATION lib
+  RUNTIME DESTINATION bin
+)
+install(
   TARGETS debayer
   EXPORT export_debayer
   ARCHIVE DESTINATION lib
@@ -130,6 +151,7 @@ ament_export_libraries(
   resize
   debayer
   rectify
+  crop_decimate
 )
 
 ament_package()
--- a/image_proc/CMakeLists.txt
+++ b/image_proc/CMakeLists.txt
@@ -121,7 +121,7 @@ install(
 )
 install(
   TARGETS crop_decimate
-  EXPORT export_resize
+  EXPORT export_crop_decimate
   ARCHIVE DESTINATION lib
   LIBRARY DESTINATION lib
   RUNTIME DESTINATION bin
@@ -145,6 +145,7 @@ ament_export_interfaces(
   export_resize HAS_LIBRARY_TARGET
   export_debayer HAS_LIBRARY_TARGET
   export_rectify HAS_LIBRARY_TARGET
+  export_crop_decimate HAS_LIBRARY_TARGET
 )
 
 ament_export_libraries(
--- a/image_proc/include/crop_decimate.hpp
+++ b/image_proc/include/crop_decimate.hpp
@@ -1,0 +1,69 @@
+/*********************************************************************
+* Software License Agreement (BSD License)
+* 
+*  Copyright (c) 2008, 2019 Willow Garage, Inc., Steve Macenski
+*  All rights reserved.
+* 
+*  Redistribution and use in source and binary forms, with or without
+*  modification, are permitted provided that the following conditions
+*  are met:
+* 
+*   * Redistributions of source code must retain the above copyright
+*     notice, this list of conditions and the following disclaimer.
+*   * Redistributions in binary form must reproduce the above
+*     copyright notice, this list of conditions and the following
+*     disclaimer in the documentation and/or other materials provided
+*     with the distribution.
+*   * Neither the name of the Willow Garage nor the names of its
+*     contributors may be used to endorse or promote products derived
+*     from this software without specific prior written permission.
+* 
+*  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+*  "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+*  LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+*  FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+*  COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+*  INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+*  BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+*  LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+*  CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+*  LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+*  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+*  POSSIBILITY OF SUCH DAMAGE.
+*********************************************************************/
+
+#ifndef IMAGE_PROC_CROP_DECIMATE_H
+#define IMAGE_PROC_CROP_DECIMATE_H
+#include <thread>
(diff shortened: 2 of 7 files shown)


image_proc/CMakeLists.txt
```diff
@@ -105,6 +119,13 @@ install(
   LIBRARY DESTINATION lib
   RUNTIME DESTINATION bin
 )
+install(
+  TARGETS crop_decimate
+  EXPORT export_resize
+  ARCHIVE DESTINATION lib
+  LIBRARY DESTINATION lib
+  RUNTIME DESTINATION bin
+)
```
JWhitleyWork, line 128: For this to work, you have to add this to `ament_export_interfaces` below.

image_proc/include/crop_decimate.hpp
```diff
@@ -0,0 +1,79 @@
...
+*  BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+*  LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+*  CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+*  LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+*  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+*  POSSIBILITY OF SUCH DAMAGE.
+*********************************************************************/
+
+#ifndef IMAGE_PROC_CROP_DECIMATE_H
+#define IMAGE_PROC_CROP_DECIMATE_H
```
JWhitleyWork: These should be in the form `<folder_name>__<file_name>_HPP_`

image_proc/include/crop_decimate.hpp
```diff
@@ -0,0 +1,79 @@
...
+
+#include <rclcpp/rclcpp.hpp>
+#include <image_transport/image_transport.h>
+#include <cv_bridge/cv_bridge.h>
+#include <sensor_msgs/image_encodings.hpp>
+#include <opencv2/imgproc/imgproc.hpp>
+
+namespace image_proc {
+
+enum CropDecimateModes {
```
JWhitleyWork: Should be a scoped enum (e.g. `enum class CropDecimateModes`).

image_proc/include/crop_decimate.hpp
```diff
@@ -0,0 +1,79 @@
...
+  CropDecimate_NN = 0,
+  CropDecimate_Linear = 1,
+  CropDecimate_Cubic = 2,
+  CropDecimate_Area = 3,
+  CropDecimate_Lanczos4 = 4
+};
+
+using namespace cv_bridge;  // CvImage, toCvShare
```
JWhitleyWork: Instead of a generic `using namespace`, use `using namespace::Component` for each component you use.

image_proc/include/crop_decimate.hpp
```diff
@@ -0,0 +1,79 @@
...
+
+class CropDecimateNode : public rclcpp::Node
+{
+public:
+  CropDecimateNode(const rclcpp::NodeOptions&);
```
JWhitleyWork: Constructors with a single argument should be marked `explicit`.

image_proc/include/crop_decimate.hpp
```diff
@@ -0,0 +1,79 @@
...
+
+private:
+  image_transport::CameraSubscriber sub_;
+  image_transport::CameraPublisher pub_;
```
JWhitleyWork, line 67: Publishers/Subscribers (including those from `image_transport`) should be created as `std::shared_ptr<Type>`.

image_proc/include/resize.hpp
```diff
@@ -54,10 +54,8 @@ class ResizeNode : public rclcpp::Node
   ResizeNode(const rclcpp::NodeOptions &);
 protected:
   // ROS communication
-  image_transport::Publisher pub_image_;
-  rclcpp::Publisher<sensor_msgs::msg::CameraInfo>::SharedPtr pub_info_;
-  image_transport::Subscriber sub_image_;
-  rclcpp::Subscription<sensor_msgs::msg::CameraInfo>::SharedPtr sub_info_;
+  image_transport::CameraPublisher pub_image_;
+  image_transport::CameraSubscriber sub_image_;
```
JWhitleyWork, line 58: `std::shared_ptr`

image_proc/src/crop_decimate.cpp
```diff
@@ -0,0 +1,315 @@
...
+  // Apply further downsampling, if necessary
+  if (decimation_x > 1 || decimation_y > 1)
+  {
+    cv::Mat decimated;
+
+    if (interpolation_ == image_proc::CropDecimate_NN)
+    {
+      // Use optimized method instead of OpenCV's more general NN resize
+      int pixel_size = output.image.elemSize();
+      switch (pixel_size)
```
JWhitleyWork, line 239: Why not just call `decimate<pixel_size>` and let `decimate` log the error if it's an unsupported size? It'll save a lot of code.

image_proc/src/crop_decimate.cpp
```diff
@@ -0,0 +1,315 @@
...
+  // If no ROI specified, leave do_rectify as-is. If ROI specified, set do_rectify = true.
+  if (width != (int)image_msg->width || height != (int)image_msg->height)
+    out_info->roi.do_rectify = true;
+
+  if (!target_frame_id_.empty()) {
+    out_image->header.frame_id = target_frame_id_;
+    out_info->header.frame_id = target_frame_id_;
+  }
+
+  pub_.publish(out_image, out_info);
```
JWhitleyWork, line 306: For Dashing, it is recommended to publish the pointers to these if we aren't going to use `unique_ptr`s.

JWhitleyWork (changes requested):

JWhitleyWork:
We really need to implement linting...

image_proc/include/crop_decimate.hpp
SteveMacenski: So I generally agree, I was trying to keep things consistent against neighboring files before a full-scale refactoring

image_proc/include/crop_decimate.hpp
SteveMacenski: can do

SteveMacenski:
@JWhitleyAStuff agreed on linting, its not there now and I'd like to not place it in scope for this particular PR. When I ran my linting tools the pages and pages of issues that came back was just too overwhelming to deal with at this particular moment in time

JWhitleyWork:
@stevemacenski I'm willing to do the linting. I actually kind of enjoy it. I'll create an issue to track it and get started.

image_proc/src/crop_decimate.cpp
SteveMacenski: This was just directly ported from ROS1, but from looking at it, not all the options are ordered numeric, so its a way of enforcing the use of specific values. I agree it could be done other ways.

Even the fact the number `N` was templated seems unwise to me vs. just another input.

image_proc/include/resize.hpp
SteveMacenski: Why?

image_proc/include/crop_decimate.hpp
SteveMacenski: Why?

JWhitleyWork: I'm actually not positive - let me check. I just know that all of the examples and the default functions from `Node` always return them. However, this isn't associated with the node so I'm not sure. I'll get back to you.

SteveMacenski:
>  I actually kind of enjoy it.

My god. I have plenty of enjoyable work for you across the ecosystem :laughing:

image_proc/src/crop_decimate.cpp
SteveMacenski: Those are both edit: shared pointers

SteveMacenski:
Added items that I understood your intent on. Look over ones I asked questions on - in particular the shared ptrs on sub/pubs which I painstakingly spent time on in Navigation to remove excess pointers.

image_proc/src/crop_decimate.cpp
JWhitleyWork: See deprecation of `shared_ptr` publish signatures here: https://github.com/ros2/rclcpp/pull/709 Discussion about preferring `unique_ptr`: https://github.com/ros2/demos/pull/327 (since it works whether you're using intra-process or not and is advantageous when you are).

image_proc/src/crop_decimate.cpp
JWhitleyWork: I'm OK with leaving it for later, though.

image_proc/src/crop_decimate.cpp
SteveMacenski: Sure, and that was part of my hot patch, but this isn't a rclcpp publisher, its a image transport publisher that wants a shared pointer specifically unless I give it a stack object

https://github.com/ros-perception/image_common/blob/dashing/image_transport/include/image_transport/camera_publisher.h#L114

JWhitleyWork: You can use refs, but, IMO, `unique_ptr` + `std::move` is the most efficient and flexible.

JWhitleyWork: HUH. OK, nevermind then. Seems an odd choice given the way that the core is going, but whatevs.

image_proc/include/crop_decimate.hpp
SteveMacenski: Agreed, but look at the image transport API - they return a base object so I'd be wrapping it in the a pointer for (no? little? unclear?) reason

https://github.com/ros-perception/image_common/blob/dashing/image_transport/include/image_transport/image_transport.h#L59

JWhitleyWork (approved):
LGTM!

JWhitleyWork:
@SteveMacenski - Try merging `maint/ci_reduce_concurrency` and see if it fixes CI.

SteveMacenski:
@JWhitleyAStuff  deleting the maint branch since its been pulled in here

Issue #468, opened 2019-11-03
update ROS2 image_proc launch file & example component pipelines


SteveMacenski:

SteveMacenski:
Issues with node/nodelet directories, or `_node` files that wrap the composible node. Some nodes still named "nodelet". Launch files need to be ported to ROS2, as well as any yaml configuration files floating around.

SteveMacenski:
Adding linter and uncrustify as well

PR #474, opened 2019-11-04
ROS2: Lint/Uncrustify image_proc

JWhitleyWork:
Completes part of #468. Does not update launch files or examples yet.


JWhitleyWork:
FYI: The `edge_aware` algorithm used in `debayer.cpp` needs testing. I replaced the macros in that file with `constexpr` functions. However, it also looks like the version of OpenCV provided with Ubuntu Bionic (3.2.0) provides this function natively too (see https://docs.opencv.org/3.2.0/d7/d1b/group__imgproc__misc.html#gga4e0972be5de079fed4e3a10e24ef5ef0af945cf163b1b5d01b69feabfe10d62bc), though it does not provide the weighted version.

CONTRIBUTING.md
```diff
@@ -0,0 +1,3 @@
+Any contribution that you make to this repository will
+be under the BSD license 2.0, as dictated by that
```
SteveMacenski, line 2: You list BSD2 but link is BSD3. 

Also I might ask that if we’re going to be specific, we allow for Apache 2.0. If I can’t contribute back in Apache, then I’m not able to help maintain during my work time, and this is true for Amazon and other larger open-source companies.

SteveMacenski (commented):
I see the updates but not CI looking for these things. I know some of the package’s will run tests for linting but most will not. 

You’ve done some great stuff here and want to make sure it doesn’t get messed up moving forward

SteveMacenski:
After going through this so deeply, what are your thoughts? 

What tools are missing from this project we could make for ROS2 (or someone else)? Any stuff have big TODOs that need fixing? 

I’m working on some work estimates to propose for internal people to contribute back to ROS but I have to be very specific. Adding new OpenCV functionality to the stereo/image_proc, finishing the port, and adding tests are on my list

JWhitleyWork:
> I see the updates but not CI looking for these things. I know some of the package’s will run tests for linting but most will not.

From the "Run Tests" step of CI: `206 tests, 0 errors, 0 failures, 0 skipped` These include all of the linter tests.

CONTRIBUTING.md
JWhitleyWork: All of the files in this repo have been under BSD from the beginning. I'm fine with new contributions being under Apache 2.0 but, if we want to change the license for any of the existing code, we would need to contact the original developers to get approval. We could also dual-license but we'll need to be careful about how we handle that. For now, I'll change the CONTRIBUTING.md to point to Apache.

Regarding the BSD 2/3 - The original or "simplified" BSD license is the "2-clause" license and the "extended" BSD license is the "3-clause," sometimes also called BSD 2.0.

JWhitleyWork: According to http://www.apache.org/legal/resolved.html#category-a, BSD 3-clause code can be included in an Apache 2.0-licensed project. How would you like to handle this?

SteveMacenski:
But the ros packages won’t run the linter by default, there were many packages like you found with linter errors that the testing didnt find because they weren’t being run.

JWhitleyWork:
@SteveMacenski - To have the linters run, you must have the following in `package.xml`:

```
<test_depend>ament_lint_auto</test_depend>
<test_depend>ament_lint_common</test_depend>
```

and the following in CMakeLists.txt:

```
if(BUILD_TESTING)
  find_package(ament_lint_auto REQUIRED)
  ament_lint_auto_find_test_dependencies()
endif()
```

That last macro call (`ament_lint_auto_find_test_dependencies()`) is where the magic happens. It looks at the `<test_depend>`s in `package.xml` and creates a list of test dependencies to be run (in this case, `ament_lint_common` which includes the following:

```
  <exec_depend>ament_cmake_copyright</exec_depend>
  <exec_depend>ament_cmake_cppcheck</exec_depend>
  <exec_depend>ament_cmake_cpplint</exec_depend>
  <exec_depend>ament_cmake_flake8</exec_depend>
  <exec_depend>ament_cmake_lint_cmake</exec_depend>
  <exec_depend>ament_cmake_pep257</exec_depend>
  <exec_depend>ament_cmake_uncrustify</exec_depend>
  <exec_depend>ament_cmake_xmllint</exec_depend>
```

It then runs all of those dependencies.

Then, when `colcon test` is run, it runs that macro.

SteveMacenski:
Sorry, let me try that again. 

I'm saying I dont think all the packages have that in the CMakeLists, or else CI would have been failing for months. 

I also just realized this is **only** for image_proc, which you added. So my bad :-)

SteveMacenski (approved):

Issue #221, opened 2016-08-24
Unable to cleanly exit image_view


chrisspen:
Is there some sort of trick for closing image_view? Clicking the standard "x" close button in the window titlebar does nothing. Sending it ctrl+c in the console does nothing. The window doesn't seem to respond to any standard exit keys like Esc or Q. I find I almost always have to use `kill` to close it.

chaonan99:
I have the same problem. I have to restrain my impulse on clicking "X" and only use ctrl+c to quit. Otherwise (if click "X" first) nothing but killing the window would work.

JWhitleyWork:
Should be closed by #481 in ROS1 and #475 in ROS2.

PR #475, opened 2019-11-06
ROS2: Port image_view

JWhitleyWork:
Port is far from complete but I wanted to create a WIP PR to let others know that it is being worked on.


SteveMacenski:
I'm sure as you've seen in the github queue, there are an awful lot of tickets complaining about image viewer not letting them save, freezing, etc. I don't want to _make_ you deal with these in the port, but if you happen to be testing by running stuff and you _do_ fix them, that would be great

JWhitleyWork:
I'll see what I can do and back-port anything I find.

SteveMacenski:
LMK when you want me to take a look

JWhitleyWork:
@SteveMacenski - Could you possibly take a look at the build failure that's happening right now? I know there's something happening in the linking phase but I'm not sure why it is failing.

SteveMacenski:
@JWhitleyAStuff try fixing 
```
/root/project/src/image_pipeline/image_view/src/disparity_view_node.cpp: In constructor ‘image_view::DisparityViewNode::DisparityViewNode(const rclcpp::NodeOptions&)’:

/root/project/src/image_pipeline/image_view/src/disparity_view_node.cpp:69:8: warning: unused variable ‘autosize’ [-Wunused-variable]

   bool autosize = this->declare_parameter("autosize", false);

```

First, then I'll take a look. CI sometimes has a habit of not failing well so if you had warnings or something odd, downstream weird things happen.

if its still happening, ping me again

JWhitleyWork:
> @JWhitleyAStuff try fixing
> 
> ```
> /root/project/src/image_pipeline/image_view/src/disparity_view_node.cpp: In constructor ‘image_view::DisparityViewNode::DisparityViewNode(const rclcpp::NodeOptions&)’:
> 
> /root/project/src/image_pipeline/image_view/src/disparity_view_node.cpp:69:8: warning: unused variable ‘autosize’ [-Wunused-variable]
> 
>    bool autosize = this->declare_parameter("autosize", false);
> ```
> 
> First, then I'll take a look. CI sometimes has a habit of not failing well so if you had warnings or something odd, downstream weird things happen.
> 
> if its still happening, ping me again

I fixed this warning and I'm still getting the same two linker errors:

`libimage_view_nodes.so: undefined reference to `vtable for image_view::DisparityViewNode'`

and

`libimage_view_nodes.so: undefined reference to `vtable for image_view::DisparityViewNode'`

JWhitleyWork:
@SteveMacenski - I figured it out. I had a function declared in `disparity_view_node.hpp` that was not declared in `disparity_view_node.cpp`. It's been fixed. Now I just need to test it and fix all the stuff that's broken!

SteveMacenski:
Ah ok! 

Errors are usually meaningful, but when I see them in CI and not locally my logic-sensor turns off and my wild-speculation-sensor turns on

image_view/src/extract_images.cpp
```diff
@@ -34,121 +34,137 @@
...
+      filename_format_(""), count_(0), _time(this->now())
   {
-    std::string topic = nh.resolveName("image");
-    ros::NodeHandle local_nh("~");
+    auto topic = rclcpp::expand_topic_or_service_name("image", this->get_name(), this->get_namespace());
 
-    std::string format_string;
-    local_nh.param("filename_format", format_string, std::string("frame%04i.jpg"));
+    if (topic == "/image") {
+      RCLCPP_WARN(
```
SteveMacenski: I think this is a debug

image_view/src/image_saver.cpp
```diff
@@ -0,0 +1,284 @@
...
+    // if (request_start_end) {
+    start_srv_ = this->create_service<std_srvs::srv::Trigger>(
+        "start", std::bind(&ImageSaverNode::callbackStartSave, this, _1, _2, _3));
+    end_srv_ = this->create_service<std_srvs::srv::Trigger>(
+        "end", std::bind(&ImageSaverNode::callbackEndSave, this, _1, _2, _3));
+    // }
+  }
+
+  bool service(
+    const std::shared_ptr<rmw_request_id_t> request_header,
```
SteveMacenski: You can also do `/*request_header*/` to avoid the `(void)request_header` but either works.

image_view/src/image_saver.cpp
```diff
@@ -0,0 +1,284 @@
...
+    (void)request_header;
+    (void)request;
+    RCLCPP_INFO(this->get_logger(), "Received end saving request");
+    end_time_ = this->now();
+
+    response->success = true;
+    return true;
+  }
+
+  void callbackWithoutCameraInfo(const sensor_msgs::msg::Image::ConstSharedPtr& image_msg)
```
SteveMacenski: `&` I'm surprised that links, I thought it was only ConstSharedPtr was the profile (but if keeping you're missing a space).

image_view/src/image_view_node.cpp
```diff
@@ -0,0 +1,254 @@
...
+void ImageViewNode::windowThread()
+{
+  cv::namedWindow(window_name_, autosize_ ? cv::WND_PROP_AUTOSIZE : 0);
+  cv::setMouseCallback(window_name_, &ImageViewNode::mouseCb, this);
+
+  if(!autosize_) {
+    int width = this->declare_parameter("width", -1);
+    int height = this->declare_parameter("height", -1);
+
+    if (width > -1 && height > -1) {
```
SteveMacenski: You can also declare and get, then check if its not been declared, but this is also OK

SteveMacenski (commented):
No real substantive issues

JWhitleyWork:
@SteveMacenski - Thanks for the feedback. I'm still cleaning up a lot of stuff so it's not quite ready for a review just yet. I'll hit you back when it's working and I've linted everything. Thanks!

image_view/src/extract_images.cpp
JWhitleyWork: I don't think it should be. It's to let the user know that they may have intended to remap the `image` topic but they didn't. Either way, this works properly now.

image_view/src/image_saver.cpp
JWhitleyWork: Done! Thanks.

image_view/src/image_saver.cpp
JWhitleyWork: Fixed.

JWhitleyWork:
@SteveMacenski - OK, I think I got the majority of it working. `image_view` works as expected now. Definitely found some things that need to be cleaned up in `melodic`. Anyway, would you mind taking a look over the code now? Please ignore formatting for the moment - I'll add linting and static analysis in a separate MR since it'll be much harder to review with all of those formatting changes.

I tested `image_view`, and `image_saver`. The rest, I don't have the hardware to test.

Also, FYI, I used this driver and, other than having a really low default FPS, it works fine: https://github.com/klintan/ros2_usb_camera

SteveMacenski:
I saw when i was cleaning up the filesystem someone added a usb capture to the publisher node. Maybe that should be pulled out again and in a new executable (like that)

image_view/src/image_saver.cpp
```diff
@@ -0,0 +1,284 @@
...
+      RCLCPP_WARN(
+        this->get_logger(), "'request_start_end' is true, so overwriting 'save_all_image' as true");
+    }
+
+    save_srv_ = this->create_service<std_srvs::srv::Empty>(
+      "save", std::bind(&ImageSaverNode::service, this, _1, _2, _3));
+
+    // FIXME(unkown): This does not make services appear
+    // if (request_start_end) {
+    start_srv_ = this->create_service<std_srvs::srv::Trigger>(
```
SteveMacenski: Does this work while processing an image? I’d think with a single threaded executor this wouldn’t get called

image_view/src/image_saver.cpp
```diff
@@ -0,0 +1,284 @@
...
+      camera_calibration_parsers::writeCalibration(filename, "camera", *info);
+    }
+
+    count_++;
+  }
+};
+
+}  // namespace image_view
+
+int main(int argc, char** argv)
```
SteveMacenski: Main in new file and register this as a component

SteveMacenski (commented):
what do you still need boost for?

SteveMacenski:
LGTM. Any of the tickets you fixed, can you comment on their tickets saying resolved in ROS2? Keep track of the ROS2-needs

image_view/src/image_saver.cpp
JWhitleyWork: Would you prefer I do this for all of the stand-alone tools? I considered it but I didn't think they would be that useful in a pipeline (though I could see this one).

SteveMacenski: I'd say "sure, doesn't hurt".

JWhitleyWork: All nodes are now components and have a separate single-node executable.

image_view/src/image_saver.cpp
JWhitleyWork: *While* an image is being processed, this will not be handled but it will go into the callback queue which will get handled on the next `spin()`. The queue is asynchronous but the processing of callbacks is sequential unless you use a MultiThreadedExecutor (in which case you have to worry more about synchronization). Even so, it wouldn't take effect until the next image is received and that image callback handled.

JWhitleyWork:
> what do you still need boost for?

https://stackoverflow.com/questions/23412978/c11-equivalent-to-boost-format

Essentially, there is no Standard Library, direct replacement for `boost::Format`. We could replace it with `std::sprintf` or an `ostringstream` but we'd have to change the format of the input parameter to match those methods or parse the meaning ourselves (since the user can input whatever they want for the "format" parameter). How do you want to handle this?

SteveMacenski:
That's OK for me. Just curious. 

Are you good for a review now?

JWhitleyWork:
@SteveMacenski - Yeah, code should be ready for review, minus formatting (will be my next PR). When you say "That's OK for me", do you mean keeping Boost or using one of the other formatting methods?

SteveMacenski:
Keeping Boost. No need to make you do everything all at once ;-) If you're so inclined to do something non-boost, power to you, but I'm not pushing it.

image_view/include/image_view/disparity_view_node.hpp
```diff
@@ -0,0 +1,35 @@
+#ifndef IMAGE_VIEW__DISPARITY_VIEW_NODE_HPP_
```
SteveMacenski, line 15: For all files you made, can you add copyright headers? This removed ambiguity

image_view/include/image_view/disparity_view_node.hpp
```diff
@@ -0,0 +1,35 @@
...
+  static unsigned char colormap[];
+
+  std::string window_name_;
+  rclcpp::Subscription<stereo_msgs::msg::DisparityImage>::SharedPtr sub_;
+  cv::Mat_<cv::Vec3b> disparity_color_;
+  bool initialized;
+
+  void imageCb(const stereo_msgs::msg::DisparityImage::SharedPtr msg);
+
+public:
```
SteveMacenski: Just out of convention, can you place the constructor and destructor first in the headers?

image_view/src/disparity_view_node.cpp
```diff
@@ -31,127 +31,92 @@
...
-  if (shutdown_on_close)
-    g_signal_connect(widget, "destroy", G_CALLBACK(destroyNode), NULL);
-  else
-    g_signal_connect(widget, "destroy", G_CALLBACK(destroyNodelet), &sub_);
-#endif
-  // Start the OpenCV window thread so we don't have to waitKey() somewhere
-  startWindowThread();
-#endif
+  sub_ = this->create_subscription<stereo_msgs::msg::DisparityImage>(
+    topic, rclcpp::QoS(10), std::bind(&DisparityViewNode::imageCb, this, std::placeholders::_1));
```
SteveMacenski, line 86: Curious - why 10?

image_view/src/image_saver_node.cpp
```diff
@@ -0,0 +1,258 @@
...
+
+  if (request_start_end && !save_all_image) {
+    RCLCPP_WARN(
+      this->get_logger(), "'request_start_end' is true, so overwriting 'save_all_image' as true");
+  }
+
+  save_srv_ = this->create_service<std_srvs::srv::Empty>(
+    "save", std::bind(&ImageSaverNode::service, this, _1, _2, _3));
+
+  // FIXME(unkown): This does not make services appear
```
SteveMacenski: Is this still a fix need?

image_view/src/image_view.cpp
```diff
@@ -0,0 +1,67 @@
...
+
+#include <rclcpp/rclcpp.hpp>
+
+#include <memory>
+
+#include "image_view/image_view_node.hpp"
+
+/*
+TODO(jwhitleyastuff): Add ROS2 dynamic reconfigure stuff
+void reconfigureCb(image_view::ImageViewConfig &config, uint32_t level)
```
SteveMacenski: So dynamic reconfigure is never coming back. We have dynamic parameters, where RQT reconfigure can just tag new parameter values and you can register parameter callback on change https://github.com/ros-planning/navigation2/pull/1370

image_view/src/video_recorder.cpp
```diff
@@ -0,0 +1,40 @@
...
+  using image_view::VideoRecorderNode;
+
+  rclcpp::init(argc, argv);
+
+  rclcpp::NodeOptions options;
+  auto vr_node = std::make_shared<VideoRecorderNode>(options);
+
+  rclcpp::spin(vr_node);
+
+  std::cout << "\nVideo saved as " << vr_node->filename << std::endl;
```
SteveMacenski: Is this necessary? shouldn't it be in the node destructor vs in the `main()`

SteveMacenski (changes requested):
Small things - mostly good

image_view/src/video_recorder.cpp
JWhitleyWork: Moved to destructor.

image_view/src/disparity_view_node.cpp
JWhitleyWork: 10 is the default queue size before providing an `rclcpp::QoS()` was required. It just explicitly uses the default.

image_view/include/image_view/disparity_view_node.hpp
JWhitleyWork: Done.

image_view/include/image_view/disparity_view_node.hpp
JWhitleyWork: Done. For those that already had BSD headers, I added a second one for Apache 2.0. For those that had Apache 2.0 headers, I combined them with a new one. For those that had no headers (new files), I just added the Apache 2.0 header.

image_view/src/image_view.cpp
JWhitleyWork: Yeah, I forgot about this. I'll work on adding them.

image_view/src/image_saver_node.cpp
JWhitleyWork: Instead of trying to "fix" this, I just removed the `request_start_end` parameter. Having topics or services appear or not appear based on parameters is bad form. I just have both services available all the time and people can choose how to use the node (whether they use the `save` service or the `start` and `end` services).

image_view/src/image_view.cpp
JWhitleyWork: Done.

SteveMacenski (approved):

JWhitleyWork:
@SteveMacenski - Sorry for the late entry. I was going back over your comment about there being several tickets regarding `image_view` and I realized that #221 reports a very annoying issue, which I finally fixed. I'll fix it in `melodic` too but there is some clean-up needed first (there is a separate implementation of `image_view` as a node vs a nodelet).

Issue #360, opened 2018-09-30
Port Image_pipeline on ROS2


yechun1:
Image_pioeline porting status: (updated on 7/8/2019)

- [x] [depth_image_proc] - **Merged (PR: #362)**
- [x] [image_publisher] - **Merged(PR: #366 )**
- [x] [image_rotate] -  **Merged(PR:https://github.com/ros-perception/image_pipeline/pull/385)**
- [x] [image_proc] - **Merged(PR: ~#369~  https://github.com/ros-perception/image_pipeline/pull/426)**
- [x] [image_view] - **Merged(PR: ~#475~ #360)**
- [x] [camera_calibration] - **in review (PR: ~#376~ #447)**
- [x] [stereo_image_proc] - no owner, depends on image_proc **in review (PR: #486)**

yechun1:
Could image_pipeline's owner/maintainer please help to create ros2 branch for [ros-perception/image_pipeline](https://github.com/ros-perception/image_pipeline)?

yechun1:
@mjcarroll thanks for help to create ros2 branch, first PR to enable ros2 depth_image_proc is ready on #362.

jacobperron:
Any update on `stereo_image_proc`?

SteveMacenski:
I can tell you @JWhitleyAStuff and I aren't thinking about it. I had alot of the ROS2 work in this repo scheduled post-ROSCon as I'm a little preoccupied with Navigation2 and slam toolbox. 

@jacobperron is that something you're motivated to look at?

yechun1:
I tried to port stereo_image_proc in the beginning of the year but blocked by image_proc which is not ready then. And now the image_proc has been ported and stereo_image_proc porting could be continue, but very sorry that I have no resource on this right now. 
It would be glad if some one could have a time to work on it. Below is previously patch(not fully completed) just for reference.

port ros2 stereo_image_proc: https://github.com/yechun1/image_pipeline/commit/e9993bed6220d28488fa73516eb95110d68f804e
on (https://github.com/yechun1/image_pipeline/tree/ros2_stereo_image_proc)

klintan:
@yechun1 one version of image_proc is actually already merged, unfortunately it is not possible to run it as a component yet, but I'll put in some PRs for that. However I think it might still be considered "ready/done" for image_pipeline. (at least as a initial release candidate) 

edit: and the PR you are referencing for image_proc is merged and closed :)

jacobperron:
[...]

PR #484, opened 2019-12-11
Port image processor library to ROS 2

jacobperron:
I think this was overlooked in during the original ROS 2 port (#426).

Summary of changes:
* Add back processor.cpp (00bc038)
* Port processor.cpp for ROS 2 (3ce82e7)
    * Added library back to CMakeLists.txt
    * Made changes for library to compile
    * Added new dependency to rcutils for logging
    * Linted to satisfy tests


jacobperron:
I'll add that I noticed that the library was missing when I started porting `stereo_image_proc` (https://github.com/ros-perception/image_pipeline/issues/360), which uses it.

SteveMacenski (approved):

PR #486, opened 2019-12-13
Port stereo_image_proc to ROS 2

jacobperron:
Connects to #360. 

I think the majority of the port is complete. Opening this PR for visibility/early feedback. I've removed OpenCV 2 support in an attempt to simplify the code.

In addition to the port, I've added some simple launch tests that can be expanded upon in the future.
Test data taken from OpenCV ([from here](https://github.com/opencv/opencv_extra/tree/dc0c6c1ee2cd142f936e7f957bbc595df2ce17e8/testdata/gpu/stereobm)).

Here's what's missing:

- Monitoring the status of connected publishers to dynamically subscribe/unsubscribe as needed

https://github.com/ros-perception/image_pipeline/blob/a0a361be0eb6cfac94d18ebc53f4a4f906dfe0e1/stereo_image_proc/src/stereo_image_proc/disparity_node.cpp#L128-L132

This feature doesn't exist in ROS 2, so I've left TODOs in the code (similar to what was done in other image_pipeline packages).

- Reconfigurable callbacks for parameters. This should be achievable since parameters are configurable by default. I'll proceed replacing the use of dynamic reconfigure using the rclcpp API. I can add it to this PR or after it is merged, whatever comes first.

- I should also double-check QoS compatibility with other parts of image_pipeline (mentioned in https://github.com/ros-perception/image_pipeline/issues/360#issuecomment-564571781).

- Windows support. E.g add missing visibility flags, cmake commands, etc.

---

I've been building/testing with ROS 2 Eloquent and the latest ROS 2 source code.


jacobperron:
`rclcpp_components_register_node` was [recently backported to Dashing](https://github.com/ros2/rclcpp/pull/935). It should be available in the [next patch release](https://discourse.ros.org/t/preparing-for-dashing-sync-and-patch-release-2019-12-12/11791) and then hopefully CI will be happy.

stereo_image_proc/launch/stereo_image_proc.launch.py
```diff
@@ -0,0 +1,64 @@
...
+
+from launch import LaunchDescription
+from launch.actions import DeclareLaunchArgument
+from launch.substitutions import LaunchConfiguration
+from launch_ros.actions import ComposableNodeContainer
+from launch_ros.descriptions import ComposableNode
+
+
+def generate_launch_description():
+    # TODO(jacobperron): Include image_proc launch file when it exists
```
SteveMacenski, line 41: Something blocking this?

stereo_image_proc/src/stereo_image_proc/disparity_node.cpp
```diff
@@ -0,0 +1,240 @@
...
+  auto disp_msg = std::make_shared<stereo_msgs::msg::DisparityImage>();
+  disp_msg->header = l_info_msg->header;
+  disp_msg->image.header = l_info_msg->header;
+
+  // Compute window of (potentially) valid disparities
+  int border = block_matcher_.getCorrelationWindowSize() / 2;
+  int left = block_matcher_.getDisparityRange() + block_matcher_.getMinDisparity() + border - 1;
+  int wtf;
+  if (block_matcher_.getMinDisparity() >= 0) {
+    wtf = border + block_matcher_.getMinDisparity();
```
SteveMacenski, line 168: What does `wtf` represent?  I think that could be named clearer

stereo_image_proc/src/stereo_image_proc/point_cloud_node.cpp
```diff
@@ -0,0 +1,268 @@
...
+    sensor_msgs::msg::CameraInfo,
+    sensor_msgs::msg::CameraInfo,
+    stereo_msgs::msg::DisparityImage>;
+  using ExactSync = message_filters::Synchronizer<ExactPolicy>;
+  using ApproximateSync = message_filters::Synchronizer<ApproximatePolicy>;
+  std::shared_ptr<ExactSync> exact_sync_;
+  std::shared_ptr<ApproximateSync> approximate_sync_;
+
+  // Publications
+  // TODO(jacobperron): Uncomment when we can be notified of subscriber status
```
SteveMacenski: Can't you check with http://docs.ros2.org/dashing/api/rclcpp/classrclcpp_1_1Node.html#a7aa2ed22450d2f3300ea5b879acd73b8?

SteveMacenski (changes requested):
LGTM, just a couple comments

stereo_image_proc/launch/stereo_image_proc.launch.py
jacobperron: Not that I'm aware of. I think we should be able to port [this launch file](https://github.com/ros-perception/image_pipeline/blob/568bb54404b740295a505234b08e23db5ddc22a3/image_proc/launch/image_proc.launchhttps://github.com/ros-perception/image_pipeline/blob/568bb54404b740295a505234b08e23db5ddc22a3/image_proc/launch/image_proc.launch), I just didn't want to add it to this PR.

stereo_image_proc/src/stereo_image_proc/disparity_node.cpp
jacobperron: I wish I knew, haha. I didn't touch the logic that was here since ROS 1. Any suggestions are appreciated.

stereo_image_proc/src/stereo_image_proc/point_cloud_node.cpp
jacobperron: We can poll for subscriptions, but I think we'd prefer this to be event driven.
I wasn't aware of the graph event API, I'll see if we can use it.

stereo_image_proc/launch/stereo_image_proc.launch.py
SteveMacenski: OK, just wondering

stereo_image_proc/src/stereo_image_proc/point_cloud_node.cpp
SteveMacenski: Ah, well what I think would accomplish the exact same thing would be to have the callback just check if any of the output topics have any subscribers, else just return without doing any work.

The `connectCb()` is nice so you don't have to process any of the callbacks that would just be exiting, but right now, Dashing/Eloquent have an issue with unsubscribing and resubscribing to a topic causing a crash (we have some navigation / rclcpp tickets around), so I'd rather do it in a non-crashing way. And its a way we can do right now with that API call. Some small performance hit, but the period of time a pipeline is A) up B) gets a message and C) doesn't have any users is pretty short

jacobperron: Sure, that sounds reasonable. For posterity, could you reference an issue regarding problems with resubscribing? Thanks!

SteveMacenski: https://github.com/ros-planning/navigation2/issues/1363

https://github.com/ros2/rmw_fastrtps/issues/341

jacobperron: See 3e40a3d

jacobperron:
Note, I also did some minor refactoring a25372d

SteveMacenski (approved):
Any other changes you'd like to make?

We should wait until we pass CI though,

stereo_image_proc/src/stereo_image_proc/disparity_node.cpp
SteveMacenski: I... don't even know what that is doing (or why there isn't one of those for the height). I'm going to just back away slowly before I go too far down a rabbithole

jacobperron:
I want to add the reconfigurable parameters, and figure out what `wtf` is.. but I can make changes in separate PRs.

It'd be nice if CI was green though before this one goes in. I figured since the latest dashing sync it should be working... :thinking:

JWhitleyWork:
@jacobperron / @SteveMacenski - They have to manually re-build the Docker containers with the updated packages for CI to pass.

SteveMacenski:
Looks like it was updated 3 days ago https://hub.docker.com/_/ros?tab=tags

SteveMacenski:
Potentially do it the more nasty way to get CI to build?

JWhitleyWork:
@SteveMacenski - You're right. I checked the SHA hash of the image it pulled and CI must be using a cached copy somewhere. I don't know that there is anyway to update this from our end.

jacobperron:
I don't think the image has been updated yet. If I pull `ros:dashing-ros-base` locally, it doesn't contain the latest version of [ros_base](https://github.com/ros2/variants/tree/dashing/ros_base) (0.7.3).

SteveMacenski:
https://circleci.com/docs/2.0/caching/#clearing-cache

@jacobperron  try adding a cache variable we can increment, ei `CACHE_VERSION=v1`

Edit: ah, ok

jacobperron:
Also, I just noticed that my tests aren't compatible with Dashing, so I'll update those in the meantime.

jacobperron:
@SteveMacenski I've updated the tests so they're backwards compatible with Dashing (ecce41b). PTAL.

jacobperron:
TL;DR It doesn't seem like the Docker image will be updated any time soon. Not until the underlying Ubuntu image gets rebuilt: https://github.com/osrf/docker_images/issues/112

It's a bit annoying, but I can refactor the CMakeLists.txt to see if we can make CI happy.

jacobperron:
See https://github.com/ros-perception/image_pipeline/pull/487 for a proposed fix for failing CI.

SteveMacenski:
Merged.

jacobperron:
Rebased :crossed_fingers:

JWhitleyWork (approved):

PR #490, opened 2020-01-07
Add reconfigurable parameters to disparity node

jacobperron:
Follow-up to #486 

The parameters are declared with descriptors containing constraints. A callback is used to update the matcher when a parameter is changed.

Also removed the old dynamic reconfigure config file.

The parameter names and defaults are the same as ROS 1, with the following exceptions:

- `fullDP` was renamed to `full_dp`
- `disp12MaxDiff` was renamed to `disp12_max_diff`

both renamed for consistency with other parameter names.


jacobperron:
This change uncovered a bug. See #491 for a fix (it should resolve the CI failure).

SteveMacenski:
I can definitely live with this, but another option would be to have lambdas for the registered callbacks when you setup the on change parameter stuff. That would let you register unique lambdas for each to effect that parameter's client rather than the long `if else else else` statement checking on what the parameter is.

If you prefer this way, I don't have an issue with it but I wanted to check that you intentionally chose this way for some reason you prefer. I generally avoid lambdas, but that particular use-case is compelling.

jacobperron:
@SteveMacenski I'm not sure I follow what you're suggesting. Any callback we register with [`set_on_parameters_set_callback`](https://github.com/ros2/rclcpp/blob/01d03c5d4e5d9b141240ca8c9077fdbc381d8c28/rclcpp/include/rclcpp/node.hpp#L843) or [`add_on_set_parameters_callback`](https://github.com/ros2/rclcpp/blob/01d03c5d4e5d9b141240ca8c9077fdbc381d8c28/rclcpp/include/rclcpp/node.hpp#L798) is passed a vector of parameters that have changed. We could maintain a map of parameter names to lambdas that would remove the need for the if-else block, but I'm not sure that it would be more readable or save lines of code.

jacobperron:
Rebased on `ros2` now that #491 is merged.

jacobperron:
Also, I opted to not use `add_on_set_parameters_callback` since it is not available in Dashing.

SteveMacenski:
Ok that’s fine, just wanted to highlight the option

SteveMacenski (approved):

JWhitleyWork (approved):
LGTM

PR #493, opened 2020-01-09
Add trigger srv to recorder

lorenz-halt:
I added service to start and stop the recording.
Further a parameter "save_all_video" (compare: image_saver) has to be set to false to enable the services.

The atomic bool requires c++11.

### DIFF
--- a/image_view/src/nodes/video_recorder.cpp
+++ b/image_view/src/nodes/video_recorder.cpp
@@ -23,6 +23,8 @@
 #include <cv_bridge/cv_bridge.h>
 #include <image_transport/image_transport.h>
 #include <camera_calibration_parsers/parse.h>
+#include "std_srvs/Trigger.h"
+#include <atomic>
 #if CV_MAJOR_VERSION == 3
 #include <opencv2/videoio.hpp>
 #endif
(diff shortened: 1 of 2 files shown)


image_view/src/nodes/video_recorder.cpp
```diff
@@ -35,76 +37,112 @@ std::string encoding;
...
+        {
+          cv_bridge::CvtColorForDisplayOptions options;
+          options.do_dynamic_scaling = use_dynamic_range;
+          options.min_image_value = min_depth_range;
+          options.max_image_value = max_depth_range;
+          options.colormap = colormap;
+          const cv::Mat image = cv_bridge::cvtColorForDisplay(cv_bridge::toCvShare(image_msg), encoding, options)->image;
+          if (!image.empty()) {
+            outputVideo << image;
+            ROS_INFO_STREAM("Recording frame " << g_count << "\x1b[1F");
```
JWhitleyWork: Can you please make this `ROS_DEBUG`? There's no need to spam the terminal every frame.

image_view/src/nodes/video_recorder.cpp
```diff
@@ -35,76 +37,112 @@ std::string encoding;
...
+            g_count++;
+            g_last_wrote_time = image_msg->header.stamp;
+          } else {
+              ROS_WARN("Frame skipped, no data!");
```
JWhitleyWork: This can be `ROS_WARN_ONCE` or `ROS_WARN_THROTTLE` to avoid spamming.

image_view/src/nodes/video_recorder.cpp
```diff
@@ -35,76 +37,112 @@ std::string encoding;
...
-    } catch(cv_bridge::Exception)
-    {
-        ROS_ERROR("Unable to convert %s image to %s", image_msg->encoding.c_str(), encoding.c_str());
-        return;
-    }
+bool start_cb(std_srvs::Trigger::Request  &req,
+         std_srvs::Trigger::Response &res)
+{
+    filename = stamp_filename(filename_orig);
+    ROS_INFO("STARTING");
```
JWhitleyWork: Prefer "Starting recording" (more context is better).

image_view/src/nodes/video_recorder.cpp
```diff
@@ -35,76 +37,112 @@ std::string encoding;
...
+    recording = true;
+    return true;
+}
+
+bool stop_cb(std_srvs::Trigger::Request  &req,
+         std_srvs::Trigger::Response &res)
+{
+    ROS_INFO("Stoping");
```
JWhitleyWork: Same - "Stopping recording" (also spelling).

JWhitleyWork (changes requested):

JWhitleyWork:
Just a few questions:

1. Why target this at `kinetic` instead of `melodic` or `noetic`? It's much preferred to target newer releases and then back-port if necessary.
2. Has this been tested with real hardware?

lorenz-halt:
I tested it with real hardware (but on kinetic). I have no doubt about melodic/noetic, but I probably cannot test it on real hardware. I will see what I can do

JWhitleyWork:
Very late ping to @ipa-lth for testing / re-targeting to `melodic` or `noetic`.

mikeferguson:
No response in over a year, and targeting a very dead branch

PR #504, opened 2020-03-18
[image_view] Add dynamic reconfigure to image_nodelet.cpp in melodic

708yamaguchi:
I added dynamic reconfigure to image_nodelet.cpp in melodic, which is removed in #479 .
We can select `do_dynamic_scaling`, `colormap`, `min_image_value` and `max_image_value`.

The converted image is published as `~/output` topic.

### DIFF
--- a/image_view/cfg/ImageView.cfg
+++ b/image_view/cfg/ImageView.cfg
@@ -21,7 +21,7 @@ edit_method_colormap = gen.enum([
     gen.const("HOT", int_t, 11, "COLORMAP_HOT"),
 ], "colormap")
 
-gen.add('do_dynamic_scaling', bool_t, 0, 'Do dynamic scaling about pixel values or not', False)
+gen.add('do_dynamic_scaling', bool_t, 0, 'Do dynamic scaling about pixel values or not', True)
 gen.add('colormap', int_t, 0, "colormap", -1, -1, 11, edit_method=edit_method_colormap);
 gen.add('min_image_value', double_t, 0, "Minimum image value for scaling depth/float image.", default=0, min=0);
 gen.add('max_image_value', double_t, 0, "Maximum image value for scaling depth/float image.", default=0, min=0);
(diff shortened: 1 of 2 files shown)


image_view/cfg/ImageView.cfg
```diff
@@ -21,7 +21,7 @@ edit_method_colormap = gen.enum([
     gen.const("HOT", int_t, 11, "COLORMAP_HOT"),
 ], "colormap")
 
-gen.add('do_dynamic_scaling', bool_t, 0, 'Do dynamic scaling about pixel values or not', False)
+gen.add('do_dynamic_scaling', bool_t, 0, 'Do dynamic scaling about pixel values or not', True)
```
SteveMacenski: Please turn back to false, don't change default behavior please

image_view/src/nodelets/image_nodelet.cpp
```diff
@@ -167,23 +180,57 @@ void ImageNodelet::onInit()
...
+  colormap_ = config.colormap;
+  min_image_value_ = config.min_image_value;
+  max_image_value_ = config.max_image_value;
 }
 
 void ImageNodelet::imageCb(const sensor_msgs::ImageConstPtr& msg)
 {
   // We want to scale floating point images so that they display nicely
-  bool do_dynamic_scaling = (msg->encoding.find("F") != std::string::npos);
+  bool do_dynamic_scaling = do_dynamic_scaling_ & (msg->encoding.find("F") != std::string::npos);
```
SteveMacenski: Do you intend to do a bitwise operation, or do you actually mean `&&`?

SteveMacenski: If the message is asking for scaling because its a float image, shouldn't we always do it? Maybe the logic should be that it does scaling for floats, but then it checks the `do_dynamic_scaling_` for non-float images to do scalling on those?

SteveMacenski (changes requested):

708yamaguchi:
Thank you for your kind review.

I understand that `do_dynamic_scaling` in imageCb() should be:
 - (for float images) always true.
 - (for non-float images) true only when `do_dynamic_scaling_` is true.

I pushed an additional commit based on your advice.

SteveMacenski (approved):
@JWhitleyWork take a look.

JWhitleyWork (approved):
Thank you for adding this functionality back to the node. LGTM.

PR #508, opened 2020-03-29
[image_view] enable to select to create image window

708yamaguchi:
In this pull request, I add `gui` param, which enables to select to create image window.

This feature is needed to reuse images colorized by `image_view` in other programs.

This is one of the features added in https://github.com/ros-perception/image_pipeline/pull/241


JWhitleyWork:
@708yamaguchi The purpose of the `image_view` nodelet is to provide a GUI window, not to process images. If the nodelet is also performing some useful image processing functions, then those should be split into another nodelet and a common library should be used for both. This is not the correct way to add this functionality.

PR #511, opened 2020-04-02
image_view: add missing depends to package.xml

scpeters:
I noticed some packages missing from the `image_view` package.xml file when building a large workspace from source. These packages have [find_package calls in the CMakeLists.txt](https://github.com/ros-perception/image_pipeline/blob/ros2/image_view/CMakeLists.txt#L18-L19), so I assume they should be listed in `<depend>` tags as well.

### DIFF
--- a/image_view/package.xml
+++ b/image_view/package.xml
@@ -22,6 +22,7 @@
   <depend>image_transport</depend>
   <depend>message_filters</depend>
   <depend>rclcpp</depend>
+  <depend>rclcpp_components</depend>
   <depend>sensor_msgs</depend>
   <depend>std_srvs</depend>
   <depend>stereo_msgs</depend>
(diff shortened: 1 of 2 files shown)


image_view/package.xml
```diff
@@ -22,6 +22,8 @@
   <depend>image_transport</depend>
   <depend>message_filters</depend>
   <depend>rclcpp</depend>
+  <depend>rclcpp_components</depend>
+  <depend>rosidl_default_generators</depend>
```
JWhitleyWork: `rosidl_default_generators` isn't necessary as this package does not create any of its own messages. Would you mind removing this here and from CMakeLists.txt?

scpeters: I reverted the addition of this line from package.xml and squashed into 8630335 and removed the find_package call in c83b533

JWhitleyWork:
Thanks!

Issue #515, opened 2020-04-13
ROS2 image_proc not creating any topics


caelinsutch:
My Launch File: 

```
import launch
import os
from ament_index_python.packages import get_package_share_directory
from launch_ros.actions import ComposableNodeContainer
from launch_ros.descriptions import ComposableNode
from launch_ros.actions import Node

def generate_launch_description():

    # Takes in gst-launch from gscam.py, which is a t for Rocos and a UDP src for ROS
    udpConfig = 'udpsrc port=5000 ! application/x-rtp,encoding-name=JPEG,payload=26 ! rtpjpegdepay ! jpegdec ! videoconvert'
    # Takes in the straight src from sensor 1
    vanillaConfig = "nvarguscamerasrc sensor-id=1 ! 'video/x-raw(memory:NVMM),width=1280,height=720,framerate=15/1, format=NV12, framerate=15/1' ! nvvidconv flip-method=2 ! video/x-raw, format=BGRx ! videoconvert"
    config_dir = os.path.join(get_package_share_directory('br-perception'), 'config')

    rightCameraConfig = 'file://' + os.path.join(config_dir, "ost_right.yaml")
    leftCameraConfig = 'file://' + os.path.join(config_dir, "ost_left.yaml")

    rightComposableNode = ComposableNode(
        package='gscam',
        node_plugin='gscam::GSCamNode',
        node_name='gscam_right',
        remappings=[
            ('/image_raw', '/stereo/right/image_raw'),
            ('/camera_info', '/stereo/right/camera_info')
        ],
        parameters=[
            os.path.join(get_package_share_directory('br-perception'), "config", "params.yaml"),
            {'camera_name': 'right'},
            {'gscam_config': udpConfig},
            {'camera_info_url': rightCameraConfig},
        ],
        extra_arguments=[{'use_intra_process_comms': True}],
    )
    leftComposableNode = ComposableNode(
        package='gscam',
        node_plugin='gscam::GSCamNode',
        node_name='gscam_left',
        remappings=[
            ('/image_raw', '/stereo/left/image_raw'),
            ('/camera_info', '/stereo/right/camera_info')
        ],
        parameters=[
            os.path.join(get_package_share_directory('br-perception'), "config", "params.yaml"),
            {'camera_name': 'left'},
            {'gscam_config': vanillaConfig},
            {'camera_info_url': leftCameraConfig},
        ],
        extra_arguments=[{'use_intra_process_comms': True}],
    )

[...]

PR #519, opened 2020-04-16
Removed namespaces as parameters and added launch file example in image_proc

caelinsutch:
For #515

### DIFF
--- a/image_proc/launch/image_proc.launch.py
+++ b/image_proc/launch/image_proc.launch.py
@@ -60,10 +60,24 @@ def generate_launch_description():
                 # Remap subscribers and publishers
                 remappings=[
                     # Subscriber remap
-                    ('image', '/image_raw'),
+                    ('image', 'image_mono'),
                     ('camera_info', 'camera_info'),
                     # Publisher remap
-                    ('/image_rect', '/image_rect_color')
+                    ('image_rect', 'imgage_rect')
+                ],
+            ),
+            # Example of rectifying an image
+            ComposableNode(
+                package='image_proc',
+                node_plugin='image_proc::RectifyNode',
+                node_name='rectify_color_node',
+                # Remap subscribers and publishers
+                remappings=[
+                    # Subscriber remap
+                    ('image', 'image_color'),
+                    ('camera_info', 'camera_info'),
+                    # Publisher remap
+                    ('image_rect', 'image_rect_color')
                 ],
             )],
         output='screen'
--- a/image_proc/launch/image_proc.launch.py
+++ b/image_proc/launch/image_proc.launch.py
@@ -42,9 +42,6 @@ NAMESPACE = "/camera"
 
 
 def generate_launch_description():
-    """
-    Generate a launch description for image processing nodes.
-    """
     ld = LaunchDescription()
 
     # Load composable container
--- a/image_proc/launch/image_proc.launch.py
+++ b/image_proc/launch/image_proc.launch.py
@@ -40,7 +40,7 @@ from launch_ros import actions, get_default_launch_description
 from launch_ros.actions import ComposableNodeContainer
 from launch_ros.descriptions import ComposableNode
 
-# Set namespace for all nodes
+# Set namespace for all nodes to /camera (for readability purposes)
 NAMESPACE = "/camera"
 
 def generate_launch_description():
@@ -83,8 +83,11 @@ def generate_launch_description():
                 node_name='rectify_color_node',
                 node_namespace=NAMESPACE,
                 parameters=[
+                    # Set the camera namespace
+                    # Namespace is where the camera_info is pulled and what the input and outputs are set relative too
                     {"camera_namespace": NAMESPACE},
-                    {"image_color": "/image_color"}
+                    # Where the raw image is coming from relative to the namespace
+                    {"image_color": "/image_color"},
                 ]
             )],
         output='screen'
--- a/image_proc/launch/image_proc.launch.py
+++ b/image_proc/launch/image_proc.launch.py
@@ -30,18 +30,16 @@
 # ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 # POSSIBILITY OF SUCH DAMAGE.
 
-import os
 import sys
 
-import launch
 from launch import LaunchDescription, LaunchIntrospector, LaunchService
 from launch_ros import actions, get_default_launch_description
 
-from launch_ros.actions import ComposableNodeContainer
 from launch_ros.descriptions import ComposableNode
 
 # Set namespace for all nodes to /camera (for readability purposes)
 NAMESPACE = "/camera"
+
 
 def generate_launch_description():
     """
@@ -61,34 +59,21 @@ def generate_launch_description():
                 node_plugin='image_proc::DebayerNode',
                 node_name='debayer_node',
                 node_namespace=NAMESPACE,
-                parameters=[
-                    {"camera_namespace": NAMESPACE}
-                ]
             ),
-            # Example of rectifying a mono image
+            # Example of rectifying an image
             ComposableNode(
                 package='image_proc',
                 node_plugin='image_proc::RectifyNode',
                 node_name='rectify_mono_node',
                 node_namespace=NAMESPACE,
-                parameters=[
-                    {"camera_namespace": NAMESPACE},
-                    {"image_mono": "/image_mono"}
-                ]
-            ),
-            # Example of rectifying a color image
-            ComposableNode(
-                package='image_proc',
-                node_plugin='image_proc::RectifyNode',
-                node_name='rectify_color_node',
-                node_namespace=NAMESPACE,
-                parameters=[
-                    # Set the camera namespace
-                    # Namespace is where the camera_info is pulled and what the input and outputs are set relative too
-                    {"camera_namespace": NAMESPACE},
-                    # Where the raw image is coming from relative to the namespace
-                    {"image_color": "/image_color"},
-                ]
+                # Remap subscribers and publishers
+                remappings=[
+                    # Subscriber remap
+                    ('/image', NAMESPACE + '/image'),
+                    ('/camera_info', NAMESPACE + '/camera_info'),
+                    # Publisher remap
+                    ('/image_rect', NAMESPACE + '/image_rect')
+                ],
             )],
         output='screen'
     )
--- a/image_proc/launch/image_proc.launch.py
+++ b/image_proc/launch/image_proc.launch.py
@@ -44,7 +44,6 @@ def generate_launch_description():
     # Load composable container
     image_processing = actions.ComposableNodeContainer(
         node_name="image_proc_container",
-        node_namespace=NAMESPACE,
         package='rclcpp_components',
         node_executable='component_container',
         composable_node_descriptions=[
@@ -52,14 +51,12 @@ def generate_launch_description():
                 package='image_proc',
                 node_plugin='image_proc::DebayerNode',
                 node_name='debayer_node',
-                node_namespace='/camera',
             ),
             # Example of rectifying an image
             ComposableNode(
                 package='image_proc',
                 node_plugin='image_proc::RectifyNode',
                 node_name='rectify_mono_node',
-                node_namespace='/camera',
                 # Remap subscribers and publishers
                 remappings=[
                     # Subscriber remap
--- a/image_proc/launch/image_proc.launch.py
+++ b/image_proc/launch/image_proc.launch.py
@@ -30,8 +30,6 @@
 # ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 # POSSIBILITY OF SUCH DAMAGE.
 
-""" A simple launch file for the nmea_serial_driver node. """
-
 import os
 import sys
 
@@ -42,6 +40,7 @@ from launch_ros import actions, get_default_launch_description
 from launch_ros.actions import ComposableNodeContainer
 from launch_ros.descriptions import ComposableNode
 
+# Set namespace for all nodes
 NAMESPACE = "/camera"
 
 def generate_launch_description():
@@ -50,6 +49,7 @@ def generate_launch_description():
     """
     ld = LaunchDescription()
 
+    # Load composable container
     image_processing = actions.ComposableNodeContainer(
         node_name="image_proc_container",
         node_namespace=NAMESPACE,
@@ -65,6 +65,7 @@ def generate_launch_description():
                     {"camera_namespace": NAMESPACE}
                 ]
             ),
+            # Example of rectifying a mono image
             ComposableNode(
                 package='image_proc',
                 node_plugin='image_proc::RectifyNode',
@@ -75,6 +76,7 @@ def generate_launch_description():
                     {"image_mono": "/image_mono"}
                 ]
             ),
+            # Example of rectifying a color image
             ComposableNode(
                 package='image_proc',
                 node_plugin='image_proc::RectifyNode',
--- a/image_proc/launch/image_proc.launch.py
+++ b/image_proc/launch/image_proc.launch.py
@@ -30,10 +30,8 @@
 # ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 # POSSIBILITY OF SUCH DAMAGE.
 
-import sys
-
-from launch import LaunchDescription, LaunchIntrospector, LaunchService
-from launch_ros import actions, get_default_launch_description
+from launch import LaunchDescription
+from launch_ros import actions
 
 from launch_ros.descriptions import ComposableNode
 
@@ -61,6 +59,8 @@ def generate_launch_description():
                 remappings=[
                     # Subscriber remap
                     ('image', 'image_mono'),
+                    ('camera_info', 'camera_info'),
+                    ('image_rect', 'image_rect')
                 ],
             ),
             # Example of rectifying an image
--- a/image_proc/launch/image_proc.launch.py
+++ b/image_proc/launch/image_proc.launch.py
@@ -61,9 +61,6 @@ def generate_launch_description():
                 remappings=[
                     # Subscriber remap
                     ('image', 'image_mono'),
-                    ('camera_info', 'camera_info'),
-                    # Publisher remap
-                    ('image_rect', 'image_rect')
                 ],
             ),
             # Example of rectifying an image
@@ -75,7 +72,6 @@ def generate_launch_description():
                 remappings=[
                     # Subscriber remap
                     ('image', 'image_color'),
-                    ('camera_info', 'camera_info'),
                     # Publisher remap
                     ('image_rect', 'image_rect_color')
                 ],
@@ -86,26 +82,3 @@ def generate_launch_description():
     ld.add_action(image_processing)
 
     return ld
-
-
-def main(argv):
-    ld = generate_launch_description()
-
-    print('Starting introspection of launch description...')
-    print('')
-
-    print(LaunchIntrospector().format_launch_description(ld))
-
-    print('')
-    print('Starting launch of launch description...')
-    print('')
-
-    ls = LaunchService()
-    ls.include_launch_description(get_default_launch_description())
-    ls.include_launch_description(ld)
-
-    return ls.run()
-
-
-if __name__ == '__main__':
-    main(sys.argv)
--- a/image_proc/launch/image_proc.launch.py
+++ b/image_proc/launch/image_proc.launch.py
@@ -37,9 +37,6 @@ from launch_ros import actions, get_default_launch_description
 
 from launch_ros.descriptions import ComposableNode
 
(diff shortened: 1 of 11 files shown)


image_proc/src/resize.cpp
```diff
@@ -54,16 +54,19 @@ ResizeNode::ResizeNode(const rclcpp::NodeOptions & options)
...
           RCLCPP_INFO(get_logger(), "camera_namespace: %s ", camera_namespace_.c_str());
-          break;
+        } else if (parameter.get_name() == "camera_topic") {
+            camera_topic_ = parameter.as_string();
+            RCLCPP_INFO(get_logger(), "camera_topic: %s ", camera_topic_.c_str());
         }
       }
 
-      image_topic_ = camera_namespace_ + "/image";
+      image_topic_ = camera_namespace_ + image_topic_;
```
SteveMacenski: I think you're missing a `/`?

image_proc/src/resize.cpp
```diff
@@ -54,16 +54,19 @@ ResizeNode::ResizeNode(const rclcpp::NodeOptions & options)
...
       camera_info_topic_ = camera_namespace_ + "/camera_info";
```
SteveMacenski: You still have camera info hardcoded, these should be consistent.

@caelinsutch why not use remapping like we typically use in image pipeline? You can leave the `namespace/image` and then remap the `namespace/image` -> `namespace2/blah_image` rather than as a parameter. I think that's what you see in the existing launch file.

image_proc/src/resize.cpp
```diff
@@ -54,16 +54,19 @@ ResizeNode::ResizeNode(const rclcpp::NodeOptions & options)
     {
       auto result = rcl_interfaces::msg::SetParametersResult();
       result.successful = true;
+      image_topic_ = "/camera";
```
SteveMacenski: You shouldn't hardcode the `/` since its going to be gotten by a parameter. Parameters convention is not to include `/`.

image_proc/launch/image_proc.launch.py
```diff
@@ -30,36 +30,87 @@
...
+
+    ls = LaunchService()
+    ls.include_launch_description(get_default_launch_description())
+    ls.include_launch_description(ld)
+
+    return ls.run()
+
+
+if __name__ == '__main__':
+    main(sys.argv)
```
SteveMacenski: Add extra line at end of files please

image_proc/launch/image_proc.launch.py
```diff
@@ -30,36 +30,87 @@
 # ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 # POSSIBILITY OF SUCH DAMAGE.
 
-from launch import LaunchDescription
+""" A simple launch file for the nmea_serial_driver node. """
```
SteveMacenski: `nmea_serial_driver`?

SteveMacenski (changes requested):
It seems like you really want a remap rather than parameters. 

e.g. `image` is default and remap `image` -> `namespace/image_mono`. This would remove the namespace and the additional parameter you added. @klinton any reason not to do that?

image_proc/src/resize.cpp
caelinsutch: I was referencing the image_proc node's rectify.cpp that uses parameters for input and output, I'll change it to remaps.

image_proc/src/resize.cpp
```diff
@@ -107,9 +107,6 @@ void ResizeNode::imageCb(
   sensor_msgs::msg::Image::ConstSharedPtr image_msg,
   sensor_msgs::msg::CameraInfo::ConstSharedPtr info_msg)
 {
-  if (pub_image_.getNumSubscribers() < 1) {
```
SteveMacenski: Why remove this? This can be good to save on traffic if no one's actually listening.

image_proc/src/resize.cpp
```diff
@@ -63,7 +63,7 @@ ResizeNode::ResizeNode(const rclcpp::NodeOptions & options)
         }
       }
 
-      image_topic_ = camera_namespace_ + "/image";
+      image_topic_ = camera_namespace_ + "/camera";
```
SteveMacenski: Why change the default to `/camera`? Isn't `image` more standard?

image_proc/launch/image_proc.launch.py
```diff
@@ -30,36 +30,89 @@
...
-            ],
-        ),
-    ])
+    """
+    Generate a launch description for image processing nodes.
+    """
+    ld = LaunchDescription()
+
+    # Load composable container
+    image_processing = actions.ComposableNodeContainer(
```
SteveMacenski, line 43: I'm not trying to be difficult, but I don't understand. You remove the parameter, OK, but I don't see any remaps here to actually enable things to change those default values. 

It seems to be that what you want is no parameters (like in the original launch file here) and use remappings of the topics. It seems like the use of `camera_namespace` is strange. If you can remap the topic, you can include the namespace there, so there's no value in having a namespace parameter. 

How do your changes make this work where the original did not?

SteveMacenski (changes requested):

image_proc/src/resize.cpp
caelinsutch: As mentioned in the other issue, there's a bug in the getNumSubscribers method that makes it always return 0. I fixed this here, your choice if you'd like me to put it back in

image_proc/src/resize.cpp
caelinsutch: This was a mistake on my end, I'll be fixing this

image_proc/launch/image_proc.launch.py
caelinsutch: I get what you're saying, I suppose I'm not sure of the value of this change, though I think an example with a remap would be valuable, which I'll be changing this too. What made my code work (on another project) was the remapping changes, which I didn't implement in this launch file. 

The `camera_namespace` keeps things a little more readable for me at least, and when used in remaps, prevents repetitive use of the same string for the namespace. I'll push an example tonight.

All in all, I don't mind the criticism, this is my first open source PR, so I'm still learning the ropes :), thanks for being accommodating!

image_proc/src/resize.cpp
SteveMacenski: Got it, can you please comment it out instead? that will be a placeholder to come back and address this issue.

image_proc/launch/image_proc.launch.py
SteveMacenski: @caelinsutch You can always make a launch file that has the namespace your pre-pend to the new remap target if you like (e.g. `{image, NAMESPACE + /remapped_name}`). But for the general case doing something like `{image, front_cam/image}` makes alot more sense to have remap abilities rather than independent namespace and image topic parameters. 

I think the right answer is to delete all the namespace parameters in image_proc components and then replace them with remaps like in the original launch file here. I think we added the namespaces here during a port before remap was available.

SteveMacenski:
Also, make sure to take a look at the CI builder, it looks like you have linting problems. You can reproduce locally using `ament_flake8` command.

image_proc/launch/image_proc.launch.py
caelinsutch: So would you like me to do that in this PR, or make a new where I delete namespace params in the image_proc node

SteveMacenski: We should do it in this PR, I'm not sure this works as is without it. 

I think what should be done:
- Remove namespace parameter (then all pub/subs should be just "image" and "camera_info")
- Launch file has remaps for "image" and "camera_info" as needed for namespacing or changing to `_mono` `_color` for use in the pipeline

Agree?

caelinsutch: Sounds good, I'll likely write that up and write a PR for it tomorrow :)

caelinsutch: Updated with removed namespace for image_proc and added launch file remaps

SteveMacenski:
I think you need to update the other image_proc items to use remap and not the camera namespace parameter. Otherwise, looks good.

caelinsutch:
I'm not sure why the CI test is failing too, it says it has a problem w making but it makes fine on two of my machines :/. I'll update the other files in image_proc to use remap

SteveMacenski:
https://github.com/ros-perception/image_pipeline/blob/ros2/.circleci/config.yml#L22

Try removing the parallel workers from the build and test.

SteveMacenski (approved):
Only thing I need to glance at is making sure all the namespaces are gone, but it looks like you got most/all of them.

SteveMacenski:
Ugh, try `colcon build --parallel-workers 1`. Sorry, dropping the tag actually defaults to the number of cores (which was >2). My bad. 

Otherwise, looks good. Just needs to pass CI and verify that this works.

SteveMacenski:
Cool, worked now. Failed from code lint errors

caelinsutch:
@SteveMacenski How do I fix the uncrustify errors? I got rid of all the cpplint ones

SteveMacenski:
Read the output of uncrustify and fix whatever issues they brought up. Not sure what else to say :smile:

caelinsutch:
```
    --- src/crop_non_zero.cpp
    +++ src/crop_non_zero.cpp.uncrustify
    @@ -50,2 +50,2 @@
    -                                                std::bind(&CropNonZeroNode::imageCb,
    -                                                    this, std::placeholders::_1), "raw");
    +      std::bind(&CropNonZeroNode::imageCb,
    +      this, std::placeholders::_1), "raw");
 ```

Not sure what this means or how to fix it... The current code format looks nothing like this, I changed these lines to match the `-` lines (I think the old style?) but its been giving me the same errors. My current code doesn't look like the `+` lines at all, not sure where it's pulling these from

**Comment by JWhitleyWork** [comment]
@caelinsutch The `+` lines are the changes it wants you to make and the `-` lines are the originals that it wants you to remove. It's throwing this error because the `uncrustify` rules want continuation lines to only be indented one level deeper than the line above them (rather than indented all the way over to match the opening parentheses).

File: image_proc/launch/image_proc.launch.py [hunk 9]
```diff
@@ -30,36 +30,74 @@
...
+        node_name="image_proc_container",
+        node_namespace=NAMESPACE,
+        package='rclcpp_components',
+        node_executable='component_container',
+        composable_node_descriptions=[
+            ComposableNode(
+                package='image_proc',
+                node_plugin='image_proc::DebayerNode',
+                node_name='debayer_node',
+                node_namespace=NAMESPACE,
```
SteveMacenski: Can you verify for me that putting the `node_namespace` as `NAMESPACE` doesn't also namespace the topics? 

E.g. `('/image', NAMESPACE + '/image')` is that necessary with the `node_namespace=NAMESPACE`?

caelinsutch: I'll check that out tomorrow morning PST and get back to you then :)

SteveMacenski: If it turns out that its redundant, remove the node namespace and leave the remaps.

JWhitleyWork: If the topic names do not have a slash (which they shouldn't), then the `node_namespace` argument will namespace the topics.

SteveMacenski: It looks like many (if not all) have leading `/`s.

JWhitleyWork: That puts them in the root namespace which is bad juju. From http://wiki.ros.org/Names:

>Names that start with a "/" are global -- they are considered fully resolved. Global names should be avoided as much as possible as they limit code portability.

The "correct" way to handle topic names is to use un-prefixed names and let namespacing handle the full path resolution.

SteveMacenski: Yeah, agreed. I think for this demo the right answer is to remove the `/` and then do the remaps without the node_namespace set.

caelinsutch: I removed the namespace and changed all the topics to relative

image_proc/launch/image_proc.launch.py
```diff
@@ -30,36 +30,68 @@
...
+            ComposableNode(
+                package='image_proc',
+                node_plugin='image_proc::DebayerNode',
+                node_name='debayer_node',
+            ),
+            # Example of rectifying an image
+            ComposableNode(
+                package='image_proc',
+                node_plugin='image_proc::RectifyNode',
+                node_name='rectify_mono_node',
```
SteveMacenski, line 57: `rectify_mono_node` but then you have the topic remap to `image_rect_color`. Looking at the other launch file, it looks like it has 2 rectify nodes: one for mono and the other for color.

image_proc/launch/image_proc.launch.py
```diff
@@ -30,36 +30,68 @@
...
+                # Remap subscribers and publishers
+                remappings=[
+                    # Subscriber remap
+                    ('image', '/image_raw'),
```
SteveMacenski: You need to remove all leading `/` from both sides of the remap for all remaps.

image_proc/src/crop_non_zero.cpp
```diff
@@ -44,44 +44,11 @@ namespace enc = sensor_msgs::image_encodings;
...
-
-// Handles (un)subscribing when clients (un)subscribe
-void CropNonZeroNode::connectCb()
-{
-  image_sub_topic_ = camera_namespace_ + "/image_raw";
-
-  RCLCPP_INFO(this->get_logger(), "subscribe: %s", image_sub_topic_.c_str());
-  sub_raw_ = image_transport::create_subscription(this, image_sub_topic_,
-      std::bind(&CropNonZeroNode::imageCb, this, std::placeholders::_1), "raw");
+  pub_ = image_transport::create_publisher(this, "/image");
```
SteveMacenski: This is actually what @JWhitleyWork meant, you're remapping these topics so these need to not have the `/` in front of them so that they could be namespaced if someone wanted. I see 10 instances of the `/` in the diff here and below. Just remove the leading `/` with no other alterations.

SteveMacenski (changes requested):
Playing wack-a-hole a little here. 
- Remove all `/` from all topics everywhere (see comments)
- Naming of the nodes mono/color. 

After that, good to go.

image_proc/src/crop_non_zero.cpp
caelinsutch: Ah I see, thank's for the clarification!

image_proc/src/resize.cpp
```diff
@@ -49,66 +49,28 @@ namespace image_proc
...
-      pub_image_ = image_transport::create_camera_publisher(this, image_topic_ + "/resize");
-
-      return result;
-    };
-
-  this->declare_parameter("camera_namespace");
+  // Create image pub
+  pub_image_ = image_transport::create_camera_publisher(this, "/resize");
+  // Create image sub
+  sub_image_ = image_transport::create_camera_subscription(this, "/image",
```
SteveMacenski: Missed 2 here: image and resize

image_proc/launch/image_proc.launch.py
```diff
@@ -30,36 +30,82 @@
...
+                package='image_proc',
+                node_plugin='image_proc::RectifyNode',
+                node_name='rectify_mono_node',
+                # Remap subscribers and publishers
+                remappings=[
+                    # Subscriber remap
+                    ('image', 'image_mono'),
+                    ('camera_info', 'camera_info'),
+                    # Publisher remap
+                    ('image_rect', 'imgage_rect')
```
SteveMacenski:
```suggestion
                    ('image_rect', 'image_rect')
```

SteveMacenski (changes requested):
2 more `/` needing removed and a type just needs to accept my change

SteveMacenski (approved):
LGTM @JWhitleyWork any objections?

image_proc/launch/image_proc.launch.py
```diff
@@ -30,36 +30,82 @@
...
+                package='image_proc',
+                node_plugin='image_proc::RectifyNode',
+                node_name='rectify_mono_node',
+                # Remap subscribers and publishers
+                remappings=[
+                    # Subscriber remap
+                    ('image', 'image_mono'),
+                    ('camera_info', 'camera_info'),
+                    # Publisher remap
+                    ('image_rect', 'image_rect')
```
JWhitleyWork, line 63: These two remaps aren't actually remapping anything so they can be removed.

image_proc/launch/image_proc.launch.py
```diff
@@ -30,36 +30,82 @@
...
+            # Example of rectifying an image
+            ComposableNode(
+                package='image_proc',
+                node_plugin='image_proc::RectifyNode',
+                node_name='rectify_color_node',
+                # Remap subscribers and publishers
+                remappings=[
+                    # Subscriber remap
+                    ('image', 'image_color'),
+                    ('camera_info', 'camera_info'),
```
JWhitleyWork: This one can also be removed.

image_proc/launch/image_proc.launch.py
```diff
@@ -30,36 +30,82 @@
...
+
+    ls = LaunchService()
+    ls.include_launch_description(get_default_launch_description())
+    ls.include_launch_description(ld)
+
+    return ls.run()
+
+
+if __name__ == '__main__':
+    main(sys.argv)
```
JWhitleyWork: You don't need the main function here. With launch files from Dashing forward, you just need to create the `generate_launch_description` function.

JWhitleyWork (changes requested):
See comments on launch file. Otherwise looks good.

JWhitleyWork (approved):
Thanks for making those changes. Once you fix the test failures I'm good to merge.

image_proc/launch/image_proc.launch.py
SteveMacenski: @JWhitleyWork since this is a demo, I think its useful to show all 3 of them even if right now they don't do anything. Lets someone easily update this for their applications.

JWhitleyWork: Either way is fine with me. I'm not picky on this one - I just thought they were unnecessary. @caelinsutch, would you mind adding them back? Sorry about the back-and-forth.

SteveMacenski:
@caelinsutch 
```
  <<< failure message
    -- run_test.py: invoking following command in '/root/project/src/image_pipeline/image_proc':
     - /opt/ros/dashing/bin/ament_flake8 --xunit-file /root/project/build/image_proc/test_results/image_proc/flake8.xunit.xml
    
    ./launch/image_proc.launch.py:33:1: F401 'sys' imported but unused
    import sys
    ^
    
    ./launch/image_proc.launch.py:35:1: F401 'launch.LaunchIntrospector' imported but unused
    from launch import LaunchDescription, LaunchIntrospector, LaunchService
    ^
    
    ./launch/image_proc.launch.py:35:1: F401 'launch.LaunchService' imported but unused
    from launch import LaunchDescription, LaunchIntrospector, LaunchService
    ^
    
    ./launch/image_proc.launch.py:36:1: F401 'launch_ros.get_default_launch_description' imported but unused
    from launch_ros import actions, get_default_launch_description
    ^
    
    4     F401 'sys' imported but unused
    
    1 files checked
    4 errors
    
    'F'-type errors: 4
    
    Checked files:
    
    * ./launch/image_proc.launch.py
    -- run_test.py: return code 1
    -- run_test.py: verify result file '/root/project/build/image_proc/test_results/image_proc/flake8.xunit.xml'
  >>>
build/image_proc/test_results/image_proc/flake8.xunit.xml: 4 tests, 0 errors, 4 failures, 0 skipped
- image_proc.flake8 F401 (./launch/image_proc.launch.py:33:1)
  <<< failure message
    'sys' imported but unused:
    import sys
  >>>
- image_proc.flake8 F401 (./launch/image_proc.launch.py:35:1)
  <<< failure message
    'launch.LaunchIntrospector' imported but unused:
    from launch import LaunchDescription, LaunchIntrospector, LaunchService
  >>>
- image_proc.flake8 F401 (./launch/image_proc.launch.py:35:1)
  <<< failure message
    'launch.LaunchService' imported but unused:
    from launch import LaunchDescription, LaunchIntrospector, LaunchService
  >>>
- image_proc.flake8 F401 (./launch/image_proc.launch.py:36:1)
  <<< failure message
    'launch_ros.get_default_launch_description' imported but unused:
    from launch_ros import actions, get_default_launch_description
  >>>
```

Run ament_flake8 locally to find these issues. You can also re-add the remaps to be complete as well, I'd appreciate that.

JWhitleyWork:
> Run ament_flake8 locally to find these issues.

I recommend running `colcon test --packages-select image_proc && colcon test-result --verbose` to make sure it catches any and all test/linter failures.

SteveMacenski:
Yes, but for just running the 1 test to quickly trouble shoot, you don't want to wait for all the tests to run to get feedback :-)

SteveMacenski:
Alright, I'm happy with it. @JWhitleyWork merge if you're happy with it. 

Sorry for such a painful experience @caelinsutch.

PR #521, opened 2020-04-20
Add parameter to use system default QoS for image subscriptions in stereo_image_proc

jacobperron:
In some applications, it is desired to have a "reliable" stereo pipeline where we ensure that stereo image messages are always received (ie. no dropped messages). This change gives the user that option.

We are still defaulting to *best effort* reliability for subscriptions because it is the least restrictive when it comes to publisher/subscription QoS compatibility.

### DIFF
--- a/stereo_image_proc/launch/stereo_image_proc.launch.py
+++ b/stereo_image_proc/launch/stereo_image_proc.launch.py
@@ -46,9 +46,8 @@ def generate_launch_description():
                         'the left and right cameras do not produce exactly synced timestamps.'
         ),
         DeclareLaunchArgument(
-            name='image_sub_reliability', default_value='best_effort',
-            description='Reliability QoS for the stereo image subscriptions in the disparity '
-                        "and point cloud nodes. Can be 'best_effort' or 'reliable'."
+            name='use_system_default_qos', default_value='False',
+            description='Use the RMW QoS settings for the publishers and subscriptions.'
         ),
         ComposableNodeContainer(
             package='rclcpp_components', node_executable='component_container',
@@ -59,7 +58,7 @@ def generate_launch_description():
                     node_plugin='stereo_image_proc::DisparityNode',
                     parameters=[{
                         'approximate_sync': LaunchConfiguration('approximate_sync'),
-                        'image_sub_reliability': LaunchConfiguration('image_sub_reliability'),
+                        'use_system_default_qos': LaunchConfiguration('use_system_default_qos'),
                     }]
                 ),
                 ComposableNode(
@@ -67,7 +66,7 @@ def generate_launch_description():
                     node_plugin='stereo_image_proc::PointCloudNode',
                     parameters=[{
                         'approximate_sync': LaunchConfiguration('approximate_sync'),
-                        'image_sub_reliability': LaunchConfiguration('image_sub_reliability'),
+                        'use_system_default_qos': LaunchConfiguration('use_system_default_qos'),
                     }]
                 ),
             ],
--- a/stereo_image_proc/launch/stereo_image_proc.launch.py
+++ b/stereo_image_proc/launch/stereo_image_proc.launch.py
@@ -47,7 +47,7 @@ def generate_launch_description():
         ),
         DeclareLaunchArgument(
             name='use_system_default_qos', default_value='False',
-            description='Use the RMW QoS settings for the publishers and subscriptions.'
+            description='Use the RMW QoS settings for the image and camera info subscriptions.'
         ),
         ComposableNodeContainer(
             package='rclcpp_components', node_executable='component_container',
--- a/stereo_image_proc/launch/stereo_image_proc.launch.py
+++ b/stereo_image_proc/launch/stereo_image_proc.launch.py
@@ -45,6 +45,11 @@ def generate_launch_description():
             description='Whether to use approximate synchronization of topics. Set to true if '
                         'the left and right cameras do not produce exactly synced timestamps.'
         ),
+        DeclareLaunchArgument(
+            name='image_sub_reliability', default_value='best_effort',
+            description='Reliability QoS for the stereo image subscriptions in the disparity '
+                        "and point cloud nodes. Can be 'best_effort' or 'reliable'."
+        ),
         ComposableNodeContainer(
             package='rclcpp_components', node_executable='component_container',
             node_name='stereo_image_proc_container', node_namespace='',
@@ -52,12 +57,18 @@ def generate_launch_description():
                 ComposableNode(
                     package='stereo_image_proc',
                     node_plugin='stereo_image_proc::DisparityNode',
-                    parameters=[{'approximate_sync': LaunchConfiguration('approximate_sync')}]
+                    parameters=[{
+                        'approximate_sync': LaunchConfiguration('approximate_sync'),
+                        'image_sub_reliability': LaunchConfiguration('image_sub_reliability'),
+                    }]
                 ),
                 ComposableNode(
                     package='stereo_image_proc',
                     node_plugin='stereo_image_proc::PointCloudNode',
-                    parameters=[{'approximate_sync': LaunchConfiguration('approximate_sync')}]
+                    parameters=[{
+                        'approximate_sync': LaunchConfiguration('approximate_sync'),
+                        'image_sub_reliability': LaunchConfiguration('image_sub_reliability'),
+                    }]
                 ),
             ],
         ),
--- a/stereo_image_proc/src/stereo_image_proc/disparity_node.cpp
+++ b/stereo_image_proc/src/stereo_image_proc/disparity_node.cpp
@@ -153,7 +153,7 @@ DisparityNode::DisparityNode(const rclcpp::NodeOptions & options)
   // Declare/read parameters
   int queue_size = this->declare_parameter("queue_size", 5);
   bool approx = this->declare_parameter("approximate_sync", false);
-  bool use_system_default_qos = this->declare_parameter("use_system_default_qos", false);
+  this->declare_parameter("use_system_default_qos", false);
 
   // Synchronize callbacks
   if (approx) {
@@ -259,10 +259,7 @@ DisparityNode::DisparityNode(const rclcpp::NodeOptions & options)
   this->declare_parameters("", double_params);
   this->declare_parameters("", bool_params);
 
-  const auto disparity_pub_qos = use_system_default_qos ?
-    rclcpp::SystemDefaultsQoS() : rclcpp::QoS(1);
-  pub_disparity_ = create_publisher<stereo_msgs::msg::DisparityImage>(
-    "disparity", disparity_pub_qos);
+  pub_disparity_ = create_publisher<stereo_msgs::msg::DisparityImage>("disparity", 1);
 
   // TODO(jacobperron): Replace this with a graph event.
   //                    Only subscribe if there's a subscription listening to our publisher.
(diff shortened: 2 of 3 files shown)


jacobperron:
I don't know what's up with CI. It looks like the build didn't make it to stereo_image_proc.

SteveMacenski:
> c++: internal compiler error: Killed (program cc1plus)

That's typically an error I see when I'm trying to compile complex code on an embedded platform because its out of RAM with no swap set. I re-ran it. I suspect its just a one time CI blip from Circle's side. 

edit: failed again. Try removing the 2 worker threads [here](https://github.com/ros-perception/image_pipeline/blob/ros2/.circleci/config.yml#L22). Though [this](https://github.com/ros-perception/image_pipeline/pull/519#pullrequestreview-395877770) PR is still building fine off dashing, but I don't see any changes you've made that would really push this over the edge. 

I'm curious on your opinion on how to handle this generally. Throughout navigation, image pipeline, pcl, etc we have parameters to set what would be "typically" sensor QoS to standard QoS. I wonder if its possible to standardize this kind of if/else configuration or if its possible to do a runtime remap of the QoS settings for a publisher like we can with the topic names. e.g. `remap_qos(topic="thing" to="StandardQoS")` or something. If you think this is also a common issue that's coming up, I can file a ticket for this under ... launch? rclcpp?

stereo_image_proc/launch/stereo_image_proc.launch.py
```diff
@@ -45,19 +45,30 @@ def generate_launch_description():
             description='Whether to use approximate synchronization of topics. Set to true if '
                         'the left and right cameras do not produce exactly synced timestamps.'
         ),
+        DeclareLaunchArgument(
+            name='image_sub_reliability', default_value='best_effort',
```
SteveMacenski: How do you feel about the parameter name being `use_system_default_qos` of type bool and the description says that when false, uses sensor data qos? Else, uses system default QoS.

SteveMacenski (commented):
I'm fine with it as is, though I think the parameter name and the logic could be cleaned up. We have the QoS set to sensor by default and if the param `use_system_default_qos` is true, then we just override it as the system default QoS. 

This is in line with the [ouster drivers](https://github.com/SteveMacenski/ros2_ouster_drivers#ros-interfaces) and seems to be a more general solution for typically sensor data that you may want to override to reliable. I'm open to thoughts though. I'm not sure I care the specific mechanics as long as there's something reasonable we can generalize to ROS2 packages processing sensor data.

jacobperron:
> I'm curious on your opinion on how to handle this generally. Throughout navigation, image pipeline, pcl, etc we have parameters to set what would be "typically" sensor QoS to standard QoS. I wonder if its possible to standardize this kind of if/else configuration or if its possible to do a runtime remap of the QoS settings for a publisher like we can with the topic names. e.g. remap_qos(topic="thing" to="StandardQoS") or something. If you think this is also a common issue that's coming up, I can file a ticket for this under ... launch? rclcpp?

There are reasons to not allow configuring the QoS generally (it could break a node authors assumptions), however, IMO there should be a standard mechanism in ROS 2 that authors can opt-in to, allowing certain QoS settings to be configured. I think deciding what QoS should be configurable and how we configure it deserves more thought and so I was planning to create a design doc to propose and discuss ideas.  I likely won't get to it until post-Foxy, but if you're so inclined you are welcome to start brainstorming and a design doc.

SteveMacenski:
> allowing certain QoS settings to be configured

That's another interesting thought: giving ranges of valid QoS that can be configured. I'd hope though that there's some standard way of representing those parameters concisely and not fill up yaml files with QoS. That's a more general solution and I think that's the way to probably go. From my experience, the only time I've exposed these is for swapping between sensor and default QoS for pretty much the same reason you submitted this PR, hence a focus on that use-case. 

Anyhow, conversation for another time and place. Do you have an objection to the `use_system_default_qos` suggestion? I think it would simplify the code and description.

stereo_image_proc/launch/stereo_image_proc.launch.py
jacobperron: This can work, though I'll point out it's possible that the system default has the same settings as SensorDataQoS since it just means we use whatever is configured in the underlying RMW. It means if users wants a "reliable" pipeline, they should set `use_system_default_qos` to `true` and ensure their RMW is configured appropriately.

That being said, I'm fine changing the parameter for conciseness as it works for me.

SteveMacenski: to clarify that: `use system default = true` I think is reliable for all RMWs, is that incorrect? Then `use system default = false` (default settings) I would expect to use sensor data qos. The parameter description then  can be more clear about the use of default vs sensor unreliable connections.

jacobperron: > use system default = true I think is reliable for all RMWs, is that incorrect?

Perhaps it is reliable for the default configuration of RMWs shipped with ROS 2 at the moment, but I don't think it is always true. The setting can change outside the control of the stereo_image_proc nodes, which I think is fine since presumably the integrator is configuring the RMW.

According to the [design doc](http://design.ros2.org/articles/qos.html):

> To honor the QoS settings of the system, developers can use the rmw_qos_profile_system_default QoS profile which delegates the responsibility of the QoS machinery to the underlying DDS vendor. This allows developers to deploy ROS 2 applications and use DDS vendor tools to configure the QoS settings.

jacobperron: If we go this route, we should just avoid giving the incorrect notion that the QoS will be reliable if `rmw_qos_profile_system_default` is used.

SteveMacenski: OK, well I wouldn't necessarily expect that to change if that's how it is now (default behavior very rarely changes in large things like that). I know the Cyclone and Fast RTPS both have reliable as default from the users of the ouster drivers who requested that be added. 

I'm OK with another profile or system default that has an explicit call for reliable. I think the latter would  resolve our differences.

jacobperron: I think changing the option to be "use system default" is good (it makes the node more configurable than just exposing the reliability setting). I think while we're at it, we should also toggle the profiles for the publishers as well.

SteveMacenski: Sure, makes sense

jacobperron: Updated, PTAL aea2740

stereo_image_proc/src/stereo_image_proc/disparity_node.cpp
```diff
@@ -258,7 +259,10 @@ DisparityNode::DisparityNode(const rclcpp::NodeOptions & options)
   this->declare_parameters("", double_params);
   this->declare_parameters("", bool_params);
 
-  pub_disparity_ = create_publisher<stereo_msgs::msg::DisparityImage>("disparity", 1);
+  const auto disparity_pub_qos = use_system_default_qos ?
+    rclcpp::SystemDefaultsQoS() : rclcpp::QoS(1);
```
SteveMacenski: Why ` rclcpp::QoS(1)` and not `rclcpp::SensorDataQoS()`? I see the existing code was just the `1`, but if we're exposing this option, shouldn't it be `SensorDataQoS`?

SteveMacenski (commented):
One last question and otherwise good to go

stereo_image_proc/src/stereo_image_proc/disparity_node.cpp
SteveMacenski: Ci is working now too, oddly

jacobperron: `rclcpp::QoS(1)` is equivalent to just passing `1`, which is using the default QoS with depth = 1. This actually has different settings than the `SensorDataQoS()`.

SteveMacenski: I understand that, I'm asking for that change to be made to `SensorDataQoS` to complement the subscribers for consistency (that we should have had before)

jacobperron: Ah, sorry, misunderstood.

In the original port, I consciously left the disparity pub/sub pair (and the point cloud publisher) reliable by default as it was closest to the ROS 1 behaviour. But, seeing we've already changed the image subscribers to best effort, I don't mind changing all pubs and subs to use best effort by default if that's your preference.

jacobperron: Well, actually, I think setting the publishers reliability to *reliable* is the most flexible, since it is compatible with both *reliable* and *best effort* subscriptions.

SteveMacenski: I'm thinking about in a pipeline, if you have this as step 7 / 10, you'd want to have the same input and outputs for each (e.g. if you wanted to subscribe to unreliable SensorDataQoS, you'd expect the param to also mess with the publisher so that they're all in line). 

With that said, I understand your comment and I think that's a valid point as well. I'm not really sure then what to do here. The 3 options we've mentioned:
- Publisher just being QoS(1) [before PR]
- Publisher having a param for the same as Subscriber for SensorData vs SystemDefaults [my original suggestion]
- Publisher having a param but with QoS(1) as the alternative [what's there now].

I would think for sensor processing the assumption would be defaulting to a sensor data QoS, but I suppose the question would be: in a pipeline of processing data, at what point should we switch from unreliable to reliable (or should we at all) between stages? Only first stage is unreliable and rest are reliable seems to be a reasonable choice, but so does all being unreliable so you don't lag behind if one stage has an issue doesn't try to retransmit.

SteveMacenski: I'll say that I think any decision is better than what's there before this PR. I don't think we need to go too overboard in deciding this. Lets come up with the best for now and there should be an image-pipeline-full analysis of QoS and making it all consistent later.

jacobperron: >  I suppose the question would be: in a pipeline of processing data, at what point should we switch from unreliable to reliable (or should we at all) between stages?

I think this is the question. And I'd posit the answer is different per application; we're not going to find an ideal configuration for everyone. By adding an option to fall back to the RMW configuration (system default), we give the user full control over how they'd like to configure the pipeline.

For default behaviour, I think making the nodes in the pipeline the most flexible (most compatible with other QoS settings) is a reasonable choice. For reliability, this means defaulting all subscriptions to best effort and all publishers to reliable. I think this is the current state of the PR.

SteveMacenski: I'm not sure I agree since the default config or Qos(1) are incredibly similar. I think that's a misuse of the variable for `use_system_default_qos` since the opposite for a subscriber is a sensordataQoS and the publisher is essentially the same thing.

I'd merge this if we just removed the use of `use_system_default_qos` for the publisher entirely and just defaulted to system default for that (for now). We can leave that bear alone for another time.

jacobperron: Maybe we have a different view of `use_system_default_qos`. To me it means the user is responsible for configuring the QoS directly with the RMW. I don't think it's safe to assume the ROS default settings (`QoS(1)`) will be similar to the RMW default.

I've removed the use of the flag for the publisher for now (9827cbe).

If we change the publishers to default to SensorDataQoS in the future, I suggest releasing the change into a new ROS distro because it will break out-of-the-box experience for users. E.g. point clouds in RViz will not render without changes to the RViz point cloud display QoS configuration.

SteveMacenski (approved):

SteveMacenski:
Waiting on CI

stereo_image_proc/src/stereo_image_proc/disparity_node.cpp
SteveMacenski: Yeah, fair enough. We're still in porting mode at the moment. We're nearing the point where we can do a release but its currently unreleased in any ROS2 distribution as of date.

PR #524, opened 2020-04-27
Add parameter to avoid points2 padding

ivanpauno:
This PR adds a parameter to `stereo_image_proc` `point_cloud_node`, which allows to send a point cloud without the standard PCL padding.

The motivation of this is that the out of the box performance of all the available `rmw` implementations with large data is really bad, even when using the `loopback` interface (any rmw imp) or shared memory transport (in `Connext`).

e.g.: To send a point cloud generated from a 640x480 image, the resulting size (with padding) is 640x480x32B = 9830400B = 9.4MB, and messages are lost even when the publisher and subscription are reliable.

It's possible to make it work with some tweaking of the receiver buffer size (increased to 8MB) when only using UDPv4 transport (tried with `cyclonedds` and `connext`). This requires to also tweak kernel `rmem_max`.
When using shared memory (only tried with `connext`), the receiver buffer size has to also be tweaked (to 16MB) to make the communication work.

When not using padding, the image is half the size, and the buffers sizes to make the communication work can also be halved.

Besides the current problems in the different DDS vendors we're using, I imagine that for many scenarios where bandwidth is limited, reducing the pointcloud size to the half might be a good option, even if it's needed to repack the point-cloud with padding before processing.

P.S.: The padding is required by `PCL`, to be able to use `SIMD`/etc.

### DIFF
--- a/stereo_image_proc/src/stereo_image_proc/point_cloud_node.cpp
+++ b/stereo_image_proc/src/stereo_image_proc/point_cloud_node.cpp
@@ -106,7 +106,7 @@ PointCloudNode::PointCloudNode(const rclcpp::NodeOptions & options)
   // can improve performance in some cases or not.
   descriptor.description =
     "This parameter avoids using alignment padding in the generated point cloud."
-    "That allows reducing bandwidth requirements, as the point cloud size is halved."
+    "This reduces bandwidth requirements, as the point cloud size is halved."
     "Using point clouds without alignment padding might degrade performance for some algorithms.";
   this->declare_parameter("avoid_point_cloud_padding", false);
 
--- a/stereo_image_proc/src/stereo_image_proc/point_cloud_node.cpp
+++ b/stereo_image_proc/src/stereo_image_proc/point_cloud_node.cpp
@@ -101,6 +101,7 @@ PointCloudNode::PointCloudNode(const rclcpp::NodeOptions & options)
   int queue_size = this->declare_parameter("queue_size", 5);
   bool approx = this->declare_parameter("approximate_sync", false);
   this->declare_parameter("use_system_default_qos", false);
+  this->declare_parameter("avoid_point_cloud_padding", false);
 
   // Synchronize callbacks
   if (approx) {
@@ -184,7 +185,23 @@ void PointCloudNode::imageCb(
   points_msg->is_dense = false;  // there may be invalid points
 
   sensor_msgs::PointCloud2Modifier pcd_modifier(*points_msg);
-  pcd_modifier.setPointCloud2FieldsByString(2, "xyz", "rgb");
+
+  if (!this->get_parameter("avoid_point_cloud_padding").as_bool()) {
+    // Data will be packed as (DC=don't care, each item is a float):
+    //   x, y, z, DC, rgb, DC, DC, DC
+    // Resulting step size: 32 bytes
+    pcd_modifier.setPointCloud2FieldsByString(2, "xyz", "rgb");
+  } else {
+    // Data will be packed as:
+    //   x, y, z, rgb
+    // Resulting step size: 16 bytes
+    pcd_modifier.setPointCloud2Fields(
+      4,
+      "x", 1, sensor_msgs::msg::PointField::FLOAT32,
+      "y", 1, sensor_msgs::msg::PointField::FLOAT32,
+      "z", 1, sensor_msgs::msg::PointField::FLOAT32,
+      "rgb", 1, sensor_msgs::msg::PointField::FLOAT32);
+  }
 
   sensor_msgs::PointCloud2Iterator<float> iter_x(*points_msg, "x");
   sensor_msgs::PointCloud2Iterator<float> iter_y(*points_msg, "y");
--- a/stereo_image_proc/src/stereo_image_proc/point_cloud_node.cpp
+++ b/stereo_image_proc/src/stereo_image_proc/point_cloud_node.cpp
@@ -108,7 +108,7 @@ PointCloudNode::PointCloudNode(const rclcpp::NodeOptions & options)
     "This parameter avoids using alignment padding in the generated point cloud."
     "This reduces bandwidth requirements, as the point cloud size is halved."
     "Using point clouds without alignment padding might degrade performance for some algorithms.";
-  this->declare_parameter("avoid_point_cloud_padding", false);
+  this->declare_parameter("avoid_point_cloud_padding", false, descriptor);
 
   // Synchronize callbacks
   if (approx) {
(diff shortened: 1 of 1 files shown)


ivanpauno:
@clalancette @jacobperron FYI

stereo_image_proc/src/stereo_image_proc/point_cloud_node.cpp
```diff
@@ -101,6 +101,7 @@ PointCloudNode::PointCloudNode(const rclcpp::NodeOptions & options)
   int queue_size = this->declare_parameter("queue_size", 5);
   bool approx = this->declare_parameter("approximate_sync", false);
   this->declare_parameter("use_system_default_qos", false);
+  this->declare_parameter("avoid_point_cloud_padding", false);
```
jacobperron: It would be nice to document why you might not want to avoid padding. Maybe with a `ParameterDescriptor` argument?

I'm not sure if we any solid documentation for ROS 2 usage of `image_pipeline`, yet.

SteveMacenski:
Should PCL perception and other ROS tools for pointcloud processing be changed to support this as well? 

Its unclear to me how this change propagates across the pointcloud processing toolset to have an impact. Will everything be able to continue to process correctly with this change?

ivanpauno:
> Will everything be able to continue to process correctly with this change?

I'm mostly sure that yes, but I haven't tried that yet.
Up to now, I've only tried how performance changes when using this parameter with a simple subscriber that doesn't do processing.

I think that the option can be added in others tool progressively, as people find the need.

SteveMacenski:
Sorry, what I'm (failing, my bad) to ask is if this change is still compatible with other tools that do not yet have this update? E.g. if my subscriber expects the pointcloud to extract using the sensor_msgs headers or through a tf2 conversion, will they work because the encoding is different?

ivanpauno:
> Sorry, what I'm (failing, my bad) to ask is if this change is still compatible with other tools that do not yet have this update?

I'm not 100% sure, but I think that things should work out of the box without making any changes in other tools.
The [fields](https://github.com/ros2/common_interfaces/blob/213f5da57182766989b49f18b0879dcef16e71b4/sensor_msgs/msg/PointCloud2.msg#L19) member of the pointcloud specifies the layout of it, and it's being filled correctly.
So, if the other tools are written correctly,  they should either work or show an error message saying that the point cloud is not using the expected layout.

> if my subscriber expects the pointcloud to extract using the sensor_msgs headers or through a tf2 conversion, will they work because the encoding is different?

I think it should work, I haven't tried though (looking at the tf2 [code](https://github.com/ros2/geometry2/blob/2e4c23caf9f17104e42976c5f8236a34496e03f0/tf2_sensor_msgs/include/tf2_sensor_msgs/tf2_sensor_msgs.h#L63-L84) for example, it should work).

----

A little caveat: Though I think changes aren't needed in other packages, in the ones doing processing not having padding can hit performance. In those cases, it might worth it to repack the  received point cloud again, this time using padding.

SteveMacenski:
> I'm not 100% sure, but I think that things should work out of the box without making any changes in other tools.

Can you verify this with some pcl_ros node? I want to be 100% sure we're not adding an option that effectively can't be used with the rest of the ecosystem. I agree it looks like it should work; trust but verify. If it breaks a couple things that weren't properly implemented, that's one thing, but just making sure that typical tools work OK. 

> in the ones doing processing not having padding can hit performance. In those cases, it might worth it to repack the received point cloud again, this time using padding.

Why would the removing the padding impact performance negatively?

ivanpauno:
> Can you verify this with some pcl_ros node? I want to be 100% sure we're not adding an option that effectively can't be used with the rest of the ecosystem. I agree it looks like it should work; trust but verify. If it breaks a couple things that weren't properly implemented, that's one thing, but just making sure that typical tools work OK.

I will check.

> Why would the removing the padding impact performance negatively?

IIUC, the reason of the padding is that "vectorized" instructions can be used (SIMD/etc).
I guess that when `PCL` finds a `pcl::PCLPointCloud2` with the correct layout, it can use the optimized algorithm.
If for using the optimized algorithm the point cloud has to be converted to a `pcl::PointCloud<pcl::PointXYZRGB>` using [1](https://github.com/ros-perception/perception_pcl/blob/36eb607be7d2f946174acee4662577ed05d0164e/pcl_conversions/include/pcl_conversions/pcl_conversions.h#L551-L557) and [2](http://docs.pointclouds.org/1.7.0/conversions_8h_source.html#l00167) (which might actually be the case), I fail to see why we're using padding in ros messages at all.

I will also investigate this last point (i.e.: if using the padding in ROS  messages allows to use faster algorithms without doing unnecessary conversions).

ivanpauno:
@SteveMacenski I've tested how this interact with other tools (a tf2 transform, and using pcl conversions), and it does work well.

I'm still not sure what's the point of using padding in `sensor_msgs::msg::PointCloud2`.
It was added a long time ago (https://github.com/ros/common_msgs/commit/0fd47de601cf9e52ce76a10a19dde914f20d8693), and then imported to `ros2`.

stereo_image_proc/src/stereo_image_proc/point_cloud_node.cpp
ivanpauno: Done

stereo_image_proc/src/stereo_image_proc/point_cloud_node.cpp
```diff
@@ -101,6 +101,13 @@ PointCloudNode::PointCloudNode(const rclcpp::NodeOptions & options)
   int queue_size = this->declare_parameter("queue_size", 5);
   bool approx = this->declare_parameter("approximate_sync", false);
   this->declare_parameter("use_system_default_qos", false);
+  rcl_interfaces::msg::ParameterDescriptor descriptor;
+  // TODO(ivanpauno): Confirm if using point cloud padding in `sensor_msgs::msg::PointCloud2`
+  // can improve performance in some cases or not.
+  descriptor.description =
+    "This parameter avoids using alignment padding in the generated point cloud."
+    "That allows reducing bandwidth requirements, as the point cloud size is halved."
```
jacobperron:
```suggestion
    "This reduces bandwidth requirements, as the point cloud size is halved."
```

jacobperron (approved):
One naming suggestion, otherwise LGTM

stereo_image_proc/src/stereo_image_proc/point_cloud_node.cpp
ivanpauno: Done

SteveMacenski (approved):

SteveMacenski:
OK good to know. Please fix CI and we can merge this guy

@ivanpauno did you notice any performance hits that you suggested might occur? Just to document for posterity

jacobperron:
@SteveMacenski It looks like the CI failure is unrelated, introduced in https://github.com/ros-perception/image_pipeline/pull/523

SteveMacenski:
@jacobperron that didn't land that way, there's no TODO in master, please rebase.

SteveMacenski:
Nice, good to go!

ivanpauno:
> @ivanpauno did you notice any performance hits that you suggested might occur? Just to document for posterity

No, but what I have tested doesn't do any "heavy" point cloud processing.
If a performance hit is noticed in the future I will let you know.

PR #538, opened 2020-05-22
[image_view] allow filename_format without format specifier

708yamaguchi:
In this pull request, I allow `filename_format` rosparam without format specifier.

For example, the default value of `filename_format` is "frame%04i.jpg"
With this pull request, "frame.jpg" is allowed.

### DIFF
--- a/image_view/src/nodelets/image_nodelet.cpp
+++ b/image_view/src/nodelets/image_nodelet.cpp
@@ -260,7 +260,15 @@ void ImageNodelet::mouseCb(int event, int x, int y, int flags, void* param)
     return;
   }
 
-  std::string filename = (this_->filename_format_ % this_->count_).str();
+  std::string filename;
+  try
+  {
+    filename = (this_->filename_format_ % this_->count_).str();
+  }
+  catch (const boost::io::too_many_args&)
+  {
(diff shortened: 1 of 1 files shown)


JWhitleyWork:
@708yamaguch This seems like a very niche use-case. It seems rare that you would only want to save _the last_ image which is published to a topic and I feel like the change you're suggesting here would have that unintended consequence if a user didn't understand it (this would be the consequence because it would simply overwrite the file with that name again and again as frames are received). Can you explain your use-case a little more?

708yamaguchi:
Thank you very much for your review.

> Can you explain your use-case a little more?

I created this pull request simply because it took me time to realize that I needed a format specifier when overwriting the file name.

> This seems like a very niche use-case. It seems rare that you would only want to save the last image which is published to a topic and I feel like the change you're suggesting here would have that unintended consequence if a user didn't understand it

I think you're right, this program is likely to cause unintended consequence.

On second thought, I'm starting to feel that it would be better to output an warn message that is easier to understand when the format specifier is invalid.
I moved to https://github.com/ros-perception/image_pipeline/pull/586.

Thanks,

JWhitleyWork:
Thank you! Closing.

PR #544, opened 2020-06-12
make parameters work in depth_image_proc

mikeferguson:


depth_image_proc/src/crop_foremost.cpp
```diff
@@ -66,7 +66,8 @@ class CropForemostNode : public rclcpp::Node
 CropForemostNode::CropForemostNode(const rclcpp::NodeOptions & options)
 : Node("CropForemostNode", options)
 {
-  this->get_parameter("distance", distance_);
+  // TODO(ros2) What is a good default for this?
```
SteveMacenski: ROS1 had no default (and uninitialized, so 0, probably). We could declare without default so it crashes if not provided in a configuration file.

SteveMacenski (approved):

SteveMacenski:
Let me know what you want to do about the distance parameter (if anything)

SteveMacenski:
```
t/project/build/depth_image_proc/test_results/depth_image_proc/uncrustify.xunit.xml'
 
  >>>
 
build/depth_image_proc/test_results/depth_image_proc/cpplint.xunit.xml: 12 tests, 0 errors, 1 failure, 0 skipped
 
- depth_image_proc.cpplint whitespace/line_length [2] (/root/project/src/image_pipeline/depth_image_proc/src/disparity.cpp:89)
 
  <<< failure message
 
    Lines should be <= 100 characters long
 
  >>>
 
build/depth_image_proc/test_results/depth_image_proc/uncrustify.xunit.xml: 12 tests, 0 errors, 1 failure, 0 skipped
 
- depth_image_proc.uncrustify src/disparity.cpp
 
  <<< failure message
 
    Diff with 6 lines
 
  >>>
 ```

Linters again

depth_image_proc/src/crop_foremost.cpp
mikeferguson: Oh, there is a call with no default-- I hadn't noticed that. I'll update to that.

mikeferguson: Actually, there isn't a call to declare_parameters without a default. So i think for now, we leave this as-is

SteveMacenski: @mikeferguson yes there is http://docs.ros2.org/dashing/api/rclcpp/classrclcpp_1_1Node.html. `default_value=rclcpp::ParameterValue()` having the parameter value structure empty is an empty non-default argument that will throw an exception when you try to `get` it. An example is https://github.com/ros-planning/navigation2/blob/master/nav2_bt_navigator/src/bt_navigator.cpp#L62 where I specifically made it so that if you don't provide a behavior tree, it will throw an exception and bring down the process as a fatal failure.

mikeferguson: ok - that wasn't compiling for me cause I still had the template parameter in there. updated and pushed.

SteveMacenski:
Thanks for fixing that up :-)

Issue #191, opened 2016-04-15
add ArUco and ChArUco board calibration as an option


ezavesky:
OpenCV3 has introduced a free-to-use powerful AR marker [called ArUco](http://www.uco.es/investiga/grupos/ava/node/26). Also introduced in this code set is the ability to calculate a pose or calibration set [along with a tutorial](http://docs.opencv.org/3.1.0/df/d4a/tutorial_charuco_detection.html).  The current board calibration works well enough, but in future generations (like when CV3 is more standard), there might be large community benefit to get this working properly.

If the time comes up, I might consider it in the future, but hopes are that this issue might tease out other volunteers in the mean time.

(some inline links in case the opencv or author page go away in the future)
`detect_board_charuco.cpp` inside the module samples folder of opencv 3

![charucoboard example](https://cloud.githubusercontent.com/assets/12480297/14561976/9edf9054-02df-11e6-95fc-ab544d18ad39.jpg)

vrabaud:
+1
Well, OpenCV3 is the default in Kinetic so have fun :)

ezavesky:
That's great news.  We're using one of those environments still based on `indigo`; do you know if that branch's version of these modules play as well with OCV3 or would some back-pulls be required?

vrabaud:
It's easy to write code that works with OpenCV2 and 3, cf http://wiki.ros.org/opencv3 but I won't have time for that. OpenCV3 is packaged in Indigo but not the default, so you will have to recompile your whole ROS stacks to use OpenCV3.

JStech:
This issue looks stale, but I'd like to have ChArUco board support. I'll see if I can get it working.

SteveMacenski:
@JStech sounds good! Its mostly stale because no one did it. If it works and it was done relatively cleanly, no issue merging it.

ezavesky:
Thanks folks, sorry that I'm not in that project space to help any more.  That said it still looks like a slick feature for calibration. +1 @JStech

JWhitleyWork:
Looks like @JStech took care of adding this feature to melodic and noetic. Thanks for that! We'll handle updating ROS2 branches with this feature as part of #449.

PR #548, opened 2020-06-16
Add ChArUco board target

JStech:
This PR adds the option of using a ChArUco board for calibration, as requested in #191. It proved more complicated than I expected, based on how the application is structured.

A ChArUco calibration can be performed with the image below as follows:
`rosrun camera_calibration cameracalibrator.py --pattern charuco --size 5x7 --marker_size 0.0172 --square 0.0279 --aruco_dict 5x5_250`
charuco 5x7

### DIFF
--- a/camera_calibration/nodes/cameracalibrator.py
+++ b/camera_calibration/nodes/cameracalibrator.py
@@ -53,7 +53,7 @@ def main():
                         "You must specify one or more chessboards as pairs of --size and --square options.")
     group.add_option("-p", "--pattern",
                      type="string", default="chessboard",
-                     help="calibration pattern to detect - 'chessboard', 'circles', 'acircles'")
+                     help="calibration pattern to detect - 'chessboard', 'circles', 'acircles', 'charuco'")
     group.add_option("-s", "--size",
                      action="append", default=[],
                      help="chessboard size as NxM, counting interior corners (e.g. a standard chessboard is 7x7)")
@@ -98,7 +98,7 @@ def main():
     group.add_option("--fisheye-k-coefficients",
                      type="int", default=4, metavar="NUM_COEFFS",
                      help="for fisheye, number of radial distortion coefficients to use fixing to zero the rest (up to 4, default %default)")
-    
+
     group.add_option("--fisheye-check-conditions",
                      action="store_true", default=False,
                      help="for fisheye, the functions will check validity of condition number")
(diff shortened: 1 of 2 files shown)


camera_calibration/nodes/cameracalibrator.py
```diff
@@ -53,13 +53,22 @@ def main():
...
+                     help="calibration pattern to detect - 'chessboard', 'circles', 'acircles', 'charuco'\n" +
+                     "  if 'charuco' is used, a --marker_size and --aruco_dict argument must be supplied\n" +
+                     "  with each --size and --square argument")
     group.add_option("-s", "--size",
                      action="append", default=[],
                      help="chessboard size as NxM, counting interior corners (e.g. a standard chessboard is 7x7)")
     group.add_option("-q", "--square",
                      action="append", default=[],
                      help="chessboard square size in meters")
+    group.add_option("-m", "--marker_size",
```
SteveMacenski: If this is just for charuco, can you make these new flags `--charuco-marker-size` or something so that its clear its only for that option?

camera_calibration/nodes/cameracalibrator.py
```diff
@@ -113,17 +122,24 @@ def main():
 
     options, args = parser.parse_args()
 
-    if len(options.size) != len(options.square):
+    if (len(options.size) != len(options.square)) or (options.pattern == "charuco" and (
```
SteveMacenski: Please break this logic out into a new function `or self.CharucoValid()` or something so that its not so convoluted

SteveMacenski (changes requested):

camera_calibration/nodes/cameracalibrator.py
JStech: :+1:

camera_calibration/nodes/cameracalibrator.py
JStech: :+1:

camera_calibration/nodes/cameracalibrator.py
```diff
@@ -113,17 +138,22 @@ def main():
 
     options, args = parser.parse_args()
 
-    if len(options.size) != len(options.square):
+    if (len(options.size) != len(options.square)):
```
SteveMacenski, line 141:
```suggestion
    if len(options.size) != len(options.square):
```

SteveMacenski (approved):

SteveMacenski:
@JStech can you retarget this to Noetic (Or a new PR for noetic) since that's the newest distribution? 

A ROS2 port of this would be nice as well to keep everything in sync

JStech:
Noetic PR: #549 

I'll look at a ROS2 port next.

PR #555, opened 2020-06-28
Fix image_publisher installation

sgvandijk:
The `libimage_publisher.so` library currently gets wrongly installed into `install/image_publisher/bin` when compiling from source, and into `/opt/ros/foxy/bin` when installing with Apt. This PR updates the CMake install commands for it to be as they would be when generated by `ros2 pkg create`, so that the library gets installed correctly at `install/image_publisher/lib/libimage_publisher.so`.

### DIFF
--- a/image_publisher/CMakeLists.txt
+++ b/image_publisher/CMakeLists.txt
@@ -59,9 +59,6 @@ install(TARGETS image_publisher_node
 install(TARGETS image_publisher
         DESTINATION lib/${PROJECT_NAME}
 )
-install(TARGETS image_publisher
-        DESTINATION bin
-)
 install(DIRECTORY launch
   DESTINATION share/${PROJECT_NAME}/
 )


image_publisher/CMakeLists.txt
```diff
@@ -44,24 +44,28 @@ ament_target_dependencies(image_publisher
...
 )
-install(TARGETS image_publisher
-        DESTINATION lib/${PROJECT_NAME}
-)
-install(TARGETS image_publisher
-        DESTINATION bin
+
+install(
+  TARGETS image_publisher
+  EXPORT export_${PROJECT_NAME}
```
JWhitleyWork: The `EXPORT` portion shouldn't be necessary. It is usually paired with `ament_export_targets()` (previously `ament_export_interfaces()`) but that isn't required here.

image_publisher/CMakeLists.txt
```diff
@@ -44,24 +44,28 @@ ament_target_dependencies(image_publisher
...
+  ARCHIVE DESTINATION lib
+  LIBRARY DESTINATION lib
+  RUNTIME DESTINATION bin
 )
+
+install(TARGETS image_publisher_node
+  EXPORT export_${PROJECT_NAME}
```
JWhitleyWork: Same. Shouldn't be required.

JWhitleyWork (changes requested):
Instead of the changes you propose (which go outside the scope of your described problem), all that should be required is to remove the redundant `install` command from lines 62-64 and maybe to change line 57 to install `image_publisher_node` to `bin` instead of `lib/${PROJECT_NAME}`.

sgvandijk:
You are right, I've reduced it to just what is necessary to have it work.

maximus5684 (approved):
LGTM! Thanks!

JWhitleyWork (approved):
LGTMT! Thanks, other me!

PR #562, opened 2020-07-09
Use newer 'add_on_set_parameters_callback' API

jacobperron:
The old API was deprecated in Foxy and since removed in https://github.com/ros2/rclcpp/pull/1199.

PR #563, opened 2020-07-09
Fix bad quotes in image_proc launch file

jacobperron:
This fixes a flake8 error.


JWhitleyWork (approved):

JWhitleyWork:
Before I merge this one, can you rebase on the `ros2` branch now that #562 is in? I want to see if we can get the Foxy job on the build farm to pass.

jacobperron:
Done. But if I'm not mistaken, I think the camera_calibration test failure is due to https://github.com/ros-perception/vision_opencv/issues/339 (I can reproduce locally).

PR #564, opened 2020-07-15
Opencv 3 compatibility

clalancette:
This PR reinstates OpenCV 3 compatibility.

While Foxy only supports Ubuntu 20.04 (and hence OpenCV 4), we still strive to maintain Ubuntu 18.04 (which has OpenCV 3).  In this case, it is trivial to keep keep image_pipeline working with OpenCV 3, so reintroduce compatibility with it.

Along with this, remove the dependency on GTK3.  It fails on Ubuntu 18.04, but it actually doesn't matter since there is no usage of the GTK API anywhere in the code.

### DIFF
--- a/image_proc/src/crop_decimate.cpp
+++ b/image_proc/src/crop_decimate.cpp
@@ -46,7 +46,13 @@ void debayer2x2toBGR(
   int R, int G1, int G2, int B)
 {
   typedef cv::Vec<T, 3> DstPixel;  // 8- or 16-bit BGR
-  dst.create(src.rows / 2, src.cols / 2, cv::traits::Type<DstPixel>::value);
+#if CV_VERSION_MAJOR >= 3 && CV_VERSION_MINOR >= 2
+  // Use OpenCV 3 API
+  dst.create(src.rows / 2, src.cols / 2, cv::DataType<DstPixel>::type);
+#else
+  // Assume OpenCV 4
+  dst.create(src.rows / 2, src.cols / 2, cv::traits::DataType<DstPixel>::type);
+#endif
 
   int src_row_step = src.step1();
   int dst_row_step = dst.step1();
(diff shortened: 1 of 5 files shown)


image_proc/src/crop_decimate.cpp
```diff
@@ -46,7 +46,13 @@ void debayer2x2toBGR(
   int R, int G1, int G2, int B)
 {
   typedef cv::Vec<T, 3> DstPixel;  // 8- or 16-bit BGR
-  dst.create(src.rows / 2, src.cols / 2, cv::traits::Type<DstPixel>::value);
+#if CV_VERSION_MAJOR >= 3 && CV_VERSION_MINOR >= 2
```
dirk-thomas: This condition looks wrong. Version 4.0 / 4.1 would evaluate to `FALSE` and 4.2 would evaluate to `TRUE`.

Since on a CMake level we already assert 3.2 or higher I would swap the blocks and just check for `CV_VERSION_MAJOR >= 4`.

SteveMacenski (approved):

dirk-thomas (approved):

dirk-thomas (approved):
The build failure on the Jenkins builds is the same as already present on the target branch.

clalancette:
CI failed because camera_calibration test failed.  It looks like a fix for that is here: https://github.com/ros-perception/vision_opencv/pull/342 .  But its not released yet.  I'll see about getting a release done for it.

Suffice it to say, this PR didn't cause that particular failure, so I think this is good to go.

JWhitleyWork (approved):
LGTM too.

clalancette:
Thanks everyone!

JWhitleyWork:
@clalancette Are you doing the new release with this change or do you want me to?

**Edit:** Same question for Dashing.

clalancette:
> @clalancette Are you doing the new release with this change or do you want me to?
> 
> **Edit:** Same question for Dashing.

It would be great if you could.  Thanks.

JWhitleyWork:
Foxy: https://github.com/ros/rosdistro/pull/25958
Dashing: https://github.com/ros/rosdistro/pull/25959

clalancette:
Thank you!

PR #569, opened 2020-07-29
[Dashing] Add rclcpp and rclcpp_components dependencies to package.xml.

nuclearsandwich:
I noticed that these are listed in CMakeLists.txt but not in package.xml
and this is causing a build failure for the binary releases on
build.ros2.org:

http://build.ros2.org/view/Dbin_ubhf_uBhf/job/Dbin_uB64__depth_image_proc__ubuntu_bionic_amd64__binary/

### DIFF
--- a/depth_image_proc/package.xml
+++ b/depth_image_proc/package.xml
@@ -29,6 +29,8 @@
   <build_depend>image_transport</build_depend>
   <build_depend>message_filters</build_depend>
   <build_depend>class_loader</build_depend>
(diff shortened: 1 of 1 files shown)


JWhitleyWork (approved):
Sorry about the build failures. Thanks for the fix!

nuclearsandwich:
> Sorry about the build failures. Thanks for the fix!

No worries. Dependency issues on release are just one of those things and it's always nice when a naive PR is the right answer.

Issue #571, opened 2020-07-29
Update ROS2 Packages to Use Ament Auto


JWhitleyWork:
The ROS Build Farm is reporting that the packages in `image_pipeline` are unstable because warnings are being generated for Foxy and Rolling that some CMake macros being used are deprecated. Switching to using the `ament_auto` macros will keep this from happening since those macros should be updated with each release.

JWhitleyWork:
Closed by #573.

PR #570, opened 2020-07-29
Add rclcpp and rclcpp_components dependencies to package.xml. (#569)

JWhitleyWork:
Forward-port of #569 to Foxy.

PR #573, opened 2020-07-29
[Foxy] Use ament_auto Macros

JWhitleyWork:
This PR applies the macros from `ament_cmake_auto` to all packages in this repository as well as fixing some other minor quirks (updating email addresses, package version mismatches, minor build warnings, etc.). Part of #571 but doesn't close it because this should be done for Dashing too.

### DIFF
--- a/camera_calibration/package.xml
+++ b/camera_calibration/package.xml
@@ -11,7 +11,7 @@
   <author>Patrick Mihelich</author>
   <maintainer email="vincent.rabaud@gmail.com">Vincent Rabaud</maintainer>
   <maintainer email="stevenmacenski@gmail.com">Steven Macenski</maintainer>
-  <maintainer email="software@autonomoustuff.com">Autonomoustuff team</maintainer>
+  <maintainer email="whitleysoftwareservices@gmail.com">Joshua Whitley</maintainer>
   
   <license>BSD</license>
   <url>http://www.ros.org/wiki/camera_calibration</url>
--- a/camera_calibration/package.xml
+++ b/camera_calibration/package.xml
@@ -12,7 +12,7 @@
   <maintainer email="stevenmacenski@gmail.com">Steven Macenski</maintainer>
   <maintainer email="whitleysoftwareservices@gmail.com">Joshua Whitley</maintainer>
   <license>BSD</license>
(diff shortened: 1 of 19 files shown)


JWhitleyWork:
Calling @nuclearsandwich for review. :)

SteveMacenski:
Are some of these new? I haven't seen `ament_auto_find_build_dependencies` and others used before. Is there a reason for that? I found in source and they appear to be old. What sparked this? 

On distro stuff - we should pull out of Rolling if things are going to continue to fail, its spamming my inbox

JWhitleyWork:
@SteveMacenski `ament_cmake_auto` has been around since Crystal and we use it in Autoware.AUTO extensively. Using `ament_auto_find_build_dependencies()` keeps from having to keep the dependency lists in `package.xml` and `CMakeLists.txt` in-sync. Using `ament_auto_package()` automatically exports an `include` folder if one is found and exports dependencies and interfaces automatically. Recently (see #570) there have been issues with package.xml/CMakeLists.txt dependencies and this helps to avoid that and many other issues.

Re: Rolling, this MR should fix the failures and instabilities there too.

camera_calibration/package.xml
```diff
@@ -32,4 +34,4 @@
   <export>
     <build_type>ament_python</build_type>
   </export>
-</package>
\ No newline at end of file
+</package>
```
nuclearsandwich, line 37: Trailing newlines are the newlines of love and peace :+1:

nuclearsandwich (approved):
I'm not the most familiar with the ament_auto utilities but as far as I can tell this looks correct and based on the CMakeLists.txt reduction it enabled I clearly ought to (hehe ought-to, *auto*) become more familiar with them.

I am curious was there a package manifest format 3 feature that you wanted which caused the package.xml format bump or was it just to enable the xmllint testing?

camera_calibration/package.xml
JWhitleyWork: I agree! :heart:

JWhitleyWork:
> I am curious was there a package manifest format 3 feature that you wanted which caused the package.xml format bump or was it just to enable the xmllint testing?

Mostly just preferrence. At least a couple of the packages were already updated to use format 3 so I figured, if there are no backward-compatibility issues, why not go with the latest?

PR #440, opened 2019-07-31
Add Fisheye calibration tool

DavidTorresOcana:
This PR introduces changes to be able to calibrate fisheye cameras and aim to Fix #146

This PR is related with ros/common_msgs#151 , ros-perception/vision_opencv#306 and ros-perception/image_common#146

Main changes it introduces are:
 * Upgrade current calibration tool to:
     - Allow user select between pinhole and fisheye model
     - Calibrate Monocular and Stereo with fisheye model

PR #580, opened 2020-08-02
Fix calibration yaml formatting

DavidTorresOcana:
This PR fixes formatting of calibration tool yaml.

camera_model -> distortion_model

This issue was introduced in #440

### DIFF
--- a/camera_calibration/src/camera_calibration/calibrator.py
+++ b/camera_calibration/src/camera_calibration/calibrator.py
@@ -654,7 +654,7 @@ class Calibrator(object):
             "  rows: 3",
             "  cols: 3",
             "  data: " + format_mat(k, 5),
-            "camera_model: " + dist_model,
+            "distortion_model: " + dist_model,
             "distortion_coefficients:",
             "  rows: 1",
             "  cols: %d" % d.size,


DavidTorresOcana:
It would be nice if anybody could merge this PR onto Noetic and ROS2

JWhitleyWork:
Yeah, looks like it was a copy/paste error. Thanks for the fix!

parameciumvance:
Why is this fix not updated to melodic release?
I mean, when apt install ros-melodic-camera-calibration, I get version 1.15.0 rather than 1.15.1

JWhitleyWork:
@parameciumvance Thanks for bringing this up. There is definitely some funny stuff going on with the [melodic rosdistro entry](https://github.com/ros/rosdistro/blob/d4ae2b9be465fe11e7e9958d0ee4ba73b6832e84/melodic/distribution.yaml#L4672). Looking into fixing it now.

JWhitleyWork:
Something must have gone wrong when I tried to make the Bloom PR for the 1.15.1 release and I apparently didn't fix it. This has now been corrected in a new 1.15.2 release (see [this PR](https://github.com/ros/rosdistro/pull/33848)) which also fixes the `indigo` references. Once this PR is merged and the next Melodic package sync happens, the issue should be resolved.

PR #586, opened 2020-08-19
[image_view] Warn when filename_format is invalid

708yamaguchi:
From https://github.com/ros-perception/image_pipeline/pull/538

Currently, when the `filename_format` rosparam is invalid (e.g. format specifier is not included) and we try to save `image_view`'s image by right-click, `image_view` terminate with the error output like below:
```
terminate called after throwing an instance of 'boost::exception_detail::clone_impl<boost::exception_
detail::error_info_injector<boost::io::too_many_args> >'
  what():  boost::too_many_args: format-string referred to fewer arguments than were passed
```
With this pull request, warn message is like below:
```
[ WARN] [1597853504.654056636]: Couldn't save image, filename_format is invalid.
```

### DIFF
--- a/image_view/src/nodelets/image_nodelet.cpp
+++ b/image_view/src/nodelets/image_nodelet.cpp
@@ -260,7 +260,16 @@ void ImageNodelet::mouseCb(int event, int x, int y, int flags, void* param)
     return;
   }
 
-  std::string filename = (this_->filename_format_ % this_->count_).str();
+  std::string filename;
+  try
+  {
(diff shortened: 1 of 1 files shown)


JWhitleyWork:
I like it! Thanks for the contribution. Would you mind porting this to Noetic as well?

708yamaguchi:
Thank you for merging!
I created noetic version PR at https://github.com/ros-perception/image_pipeline/pull/587

PR #587, opened 2020-08-19
[image_view] Warn when filename_format is invalid

708yamaguchi:
This pull request is noetic version of https://github.com/ros-perception/image_pipeline/pull/586

I am sorry, but I have not been able to test with noetic, because I do not have noetic environment.
cc @JWhitleyWork


JWhitleyWork:
Thanks for making this! This is an incredibly minor change so as long as CI passes, I'm fine.

PR #581, opened 2020-08-02
Update fisheye distortion model definition

DavidTorresOcana:
Update fisheye distortion model definition.

Update calibration tool distortion model for fisheye camera to be compatible with ros-perception/vision_opencv#306 and ros-perception/image_common#146 in as indicated in ros/common_msgs#151.

PR #592, opened 2020-08-23
Update fisheye distortion model definition

DavidTorresOcana:
Update fisheye distortion model definition.
Replaces #581 retargeting to `noetic`

Update calibration tool distortion model for fisheye camera to be compatible with ros-perception/vision_opencv#306 and ros-perception/image_common#146 in as indicated in ros/common_msgs#151.

### DIFF
--- a/camera_calibration/src/camera_calibration/calibrator.py
+++ b/camera_calibration/src/camera_calibration/calibrator.py
@@ -295,7 +295,7 @@ def _get_dist_model(dist_params, cam_model):
         else:
             dist_model = "plumb_bob"
     elif CAMERA_MODEL.FISHEYE == cam_model:
-        dist_model = "fisheye"
+        dist_model = "equidistant"
     else:
         dist_model = "unknown"
     return dist_model


mintar:
+1

This is the correct thing to do. See [my comment on common_msgs#151](https://github.com/ros/common_msgs/pull/151#issuecomment-659981501) for the full explanation.

JWhitleyWork:
@DavidTorresOcana @mintar Is this PR dependent upon the others? It looks like it builds without it but we're talking about a text string that probably doesn't have a definition backing it until at least ros-perception/vision_opencv#306 is merged, right?

DavidTorresOcana:
@JWhitleyWork  This PR does not depends on ros-perception/vision_opencv#306 nor ros-perception/vision_opencv#358

It simply renames the distortion model the tool gives to the distortion model already defined in [ros/common_msgs](https://github.com/ros/common_msgs/blob/noetic-devel/sensor_msgs/include/sensor_msgs/distortion_models.h#L47) so that they match. So those two need to be the same. 

See this [PR comments](https://github.com/ros/common_msgs/pull/151#issuecomment-659981501)

JWhitleyWork:
Thanks for the clarification. LGTM!

JWhitleyWork (approved):

PR #593, opened 2020-08-23
make crop_decimate work

mikeferguson:
in ROS1, there were two different nodehandles,
which put these in different namespaces. Right
now this publishes the same topic it subscribes
to. The style of naming also allows the
camera_info topics to be properly remapped

### DIFF
--- a/image_proc/src/crop_decimate.cpp
+++ b/image_proc/src/crop_decimate.cpp
@@ -119,9 +119,9 @@ CropDecimateNode::CropDecimateNode(const rclcpp::NodeOptions & options)
   int interpolation = this->declare_parameter("interpolation", 0);
   interpolation_ = static_cast<CropDecimateModes>(interpolation);
 
-  pub_ = image_transport::create_camera_publisher(this, "image_raw");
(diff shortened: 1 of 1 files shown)


image_proc/src/crop_decimate.cpp
```diff
@@ -119,9 +119,9 @@ CropDecimateNode::CropDecimateNode(const rclcpp::NodeOptions & options)
   int interpolation = this->declare_parameter("interpolation", 0);
   interpolation_ = static_cast<CropDecimateModes>(interpolation);
 
-  pub_ = image_transport::create_camera_publisher(this, "image_raw");
+  pub_ = image_transport::create_camera_publisher(this, "out/image");
   sub_ = image_transport::create_camera_subscription(
-    this, "image_raw", std::bind(
+    this, "in/image", std::bind(
```
JWhitleyWork: In order to maintain at least some of the existing naming, can you please change these to `out/image_raw` and `in/image_raw`?

mikeferguson: done

JWhitleyWork (approved):

PR #600, opened 2020-09-18
[video_recorder] Add timer to record low fps video at the correct speed

708yamaguchi:
Without this pull request, when we record video with lower fps than `fps` rosparam, the output video will be fast and shortened.

For example, if we set `fps` rosparam as 15 but input `sensor_msgs::Image` topic's hz is 5 and we record video for 30 seconds, the output video is only 10 seconds and 3x speed.

This pull request solves the above problem by using timer callback and outputting video at regular intervals.

### DIFF
--- a/image_view/src/nodes/video_recorder.cpp
+++ b/image_view/src/nodes/video_recorder.cpp
@@ -40,7 +40,6 @@ double max_depth_range;
 bool use_dynamic_range;
 int colormap;
 cv::Mat image;
-ros::Time stamp;
 
 
 void callback(const sensor_msgs::ImageConstPtr& image_msg)
@@ -72,8 +71,7 @@ void callback(const sensor_msgs::ImageConstPtr& image_msg)
 
     }
 
-    stamp = image_msg->header.stamp;
-    if ((stamp - g_last_wrote_time) < ros::Duration(1.0 / fps))
+    if ((image_msg->header.stamp - g_last_wrote_time) < ros::Duration(1.0 / fps))
     {
       // Skip to get video with correct fps
       return;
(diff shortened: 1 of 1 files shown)


image_view/src/nodes/video_recorder.cpp
```diff
@@ -83,22 +86,26 @@ void callback(const sensor_msgs::ImageConstPtr& image_msg)
...
     }
 }
 
+void timercallback(const ros::TimerEvent&)
+{
+    if (!image.empty()) {
+      outputVideo << image;
+      ROS_INFO_STREAM("Recording frame " << g_count << "\x1b[1F");
+      g_count++;
+      g_last_wrote_time = stamp;
```
JWhitleyWork: This is no longer true. The timestamp of the last-received image can no longer be considered close enough to the actual time that the frame was written. Use `ros::Time::now()` instead.

image_view/src/nodes/video_recorder.cpp
```diff
@@ -70,7 +72,8 @@ void callback(const sensor_msgs::ImageConstPtr& image_msg)
 
     }
 
-    if ((image_msg->header.stamp - g_last_wrote_time) < ros::Duration(1.0 / fps))
+    stamp = image_msg->header.stamp;
+    if ((stamp - g_last_wrote_time) < ros::Duration(1.0 / fps))
```
JWhitleyWork: This change is unnecessary given my other comment.

JWhitleyWork (changes requested):
In general, I'm OK with this feature. Just needs some tweaks.

708yamaguchi:
Thank you for your review.

I followed your advice and use `ros::Time::now()`.

PR #601, opened 2020-09-19
Make GTK3 and harfbuzz optional

seanyen:
Those dependency is not mandatory and the `image_view` is already resilient enough when those packages are absent.

This pull request is to make them optional.

(This is motivated while we enabled it for ROS on Windows project.)

### DIFF
--- a/image_view/CMakeLists.txt
+++ b/image_view/CMakeLists.txt
@@ -40,10 +40,11 @@ endif()
 
 find_package(PkgConfig REQUIRED)
 pkg_check_modules(GTK3 gtk+-3.0)
-if(NOT GTK3_FOUND)
+if(GTK3_FOUND)
+  include_directories(${GTK3_INCLUDE_DIRS})
+else()
   message(STATUS "GTK3 is not found. Some features are disabled.")
 endif()
-include_directories(${GTK3_INCLUDE_DIRS})
 
 pkg_check_modules(PC_HB harfbuzz)
 if(PC_HB_FOUND)
--- a/image_view/CMakeLists.txt
+++ b/image_view/CMakeLists.txt
@@ -39,10 +39,16 @@ if(ANDROID)
 endif()
 
 find_package(PkgConfig REQUIRED)
-pkg_check_modules(GTK3 REQUIRED gtk+-3.0)
+pkg_check_modules(GTK3 gtk+-3.0)
+if(NOT GTK3_FOUND)
+  message(STATUS "GTK3 is not found. Some features are disabled.")
+endif()
 include_directories(${GTK3_INCLUDE_DIRS})
 
-pkg_check_modules(PC_HB REQUIRED harfbuzz)
+pkg_check_modules(PC_HB harfbuzz)
+if(NOT PC_HB_FOUND)
+  message(STATUS "harfbuzz is not found. Some features are disabled.")
+endif()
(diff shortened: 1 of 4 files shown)


image_view/CMakeLists.txt
```diff
@@ -39,10 +39,16 @@ if(ANDROID)
 endif()
 
 find_package(PkgConfig REQUIRED)
-pkg_check_modules(GTK3 REQUIRED gtk+-3.0)
+pkg_check_modules(GTK3 gtk+-3.0)
+if(NOT GTK3_FOUND)
+  message(STATUS "GTK3 is not found. Some features are disabled.")
+endif()
 include_directories(${GTK3_INCLUDE_DIRS})
```
JWhitleyWork: I recommend the following:

```suggestion
pkg_check_modules(GTK3 gtk+-3.0)
if(GTK3_FOUND)
  include_directories(${GTK3_INCLUDE_DIRS})
else()
  message(STATUS "GTK3 is not found. Some features are disabled.")
endif()
```

image_view/CMakeLists.txt
```diff
@@ -39,10 +39,16 @@ if(ANDROID)
...
 
-pkg_check_modules(PC_HB REQUIRED harfbuzz)
+pkg_check_modules(PC_HB harfbuzz)
+if(NOT PC_HB_FOUND)
+  message(STATUS "harfbuzz is not found. Some features are disabled.")
+endif()
 include_directories(${PC_HB_INCLUDE_DIRS})
```
JWhitleyWork: Same here:

```suggestion
pkg_check_modules(PC_HB harfbuzz)
if(PC_HB_FOUND)
  include_directories(${PC_HB_INCLUDE_DIRS})
else()
  message(STATUS "harfbuzz is not found. Some features are disabled.")
endif()
```

JWhitleyWork (changes requested):
On what systems/environments have you tested this change? Are you sure that all modules in `image_view` will run without these dependencies?

seanyen:
> On what systems/environments have you tested this change? Are you sure that all modules in image_view will run without these dependencies?

On Windows! We natively run the [ROS on Windows](https://discourse.ros.org/t/ros-on-windows-melodic-release-v20200915-0-0-2009211616). My test coverage might not be 100%, but the `image_view` gets the image feed and it seems to be functional by this configuration. Let me know if any use cases I need to pay attention to.


NOTE: Please disregard the test_nodelets error which is not related to `image_view`.

test

JWhitleyWork:
@seanyen Sorry I haven't been back around to this for a while. Would you mind rebasing on the upstream `noetic` so I can test?

JWhitleyWork:
Additionally, I would prefer to remove the GTK dependency completely and just let OpenCV deal with it. Would you mind implementing that in this PR?

JWhitleyWork:
@seanyen Ping for updates.

seanyen:
@JWhitleyWork
Sorry for being late on this. Just got back from my parental leave. 👶🍼

I rebased to the latest noetic, removed the harfbuzz dep (which seems to be extra), and removed GTK3 dep as you suggested.

JWhitleyWork (approved):

JWhitleyWork:
Sorry for the very late reply. Merged!

Issue #604, opened 2020-09-29
video_recorder skips frames even when they arrive just a fraction too early


cbteeple:
### Package
**image_view**

### Feature
_video_recorder_ skips frames when they arrive too soon to ensure proper framerate

### Expected Behavior
If a camera stream is running at 30 fps, then _video_recorder_ should be able to save a 30fps video without dropping frames

### Actual Behavior
_video_recorder_ skips frames when they arrive even a fraction of time too early. Since the opencv VideoWriter class expects frames to be given at the set framerate, we end up with a shorter video that plays back faster.

mikeferguson:
Calling this a duplicate of https://github.com/ros-perception/image_pipeline/issues/274 - at some point sec_per_frame needs a refactor

PR #605, opened 2020-09-29
Added frame-skipping tolerance parameter

cbteeple:
### Package/Feature:
**image_view**,  _video_recorder_ 

### Description:
Added a tolerance parameter for frame-skipping that enables the user to choose how much variation in framerate is acceptable.

### What this fixes:
This is a fix for Issue #604. Adding a frame-skipping tolerance allows us to change the condition for frame skipping to allow for some natural variation (which often occurs when streaming video from a webcam using [usb_cam](http://wiki.ros.org/usb_cam) for example).

### Details:
The tolerance parameter is a percent, and is passed to the _video_recorder_ node via the parameter server:
```xml
<node name="video_recorder_node" pkg="image_view" type="video_recorder" respawn="false" output="screen">
    <remap from="image" to="/usb_cam/image_raw"/>
    <param name="fps" value="30" />
    <param name="tolerance" value="15" />
    <param name="codec" value="avc1" />
    <param name="filename" value="output.mp4" />
</node>
```

The condition in _video_recorder_ for frame skipping changes from:
```cpp
if ( (image_msg->header.stamp - g_last_wrote_time) < ros::Duration(1.0 / fps) )
```
to 
```cpp
if ( (image_msg->header.stamp - g_last_wrote_time) < ros::Duration( (1.0 / fps)*(1.0-tolerance/100) ) )
```
This allows some amount of variation in the framerate to be acceptable.

### Testing
**Platform:** Ubuntu 20.04, ROS Noetic
**Hardware:** Logitech C920 Webcam

1. Start a camera stream using [usb_cam](http://wiki.ros.org/usb_cam), with 30fps using the mjpeg format to get the direct compressed stream from this camera.
2. Start a video_recorder with 30 fps, 15% tolerance, and the avc1 codec (for H.264 compression)
3. No frames are dropped!
4. Repeat this with 0% tolerance: approximately every 5th frame dropped, so video is 20% shorter than it should be, and plays back too fast.

I think this confirms that the addition of the frame-skip tolerance has fixed the problem.

### DIFF
--- a/image_view/src/nodes/video_recorder.cpp
+++ b/image_view/src/nodes/video_recorder.cpp
@@ -34,6 +34,7 @@ ros::Time g_last_wrote_time = ros::Time(0);
 std::string encoding;
 std::string codec;
 int fps;
+double tolerance;
 std::string filename;
 double min_depth_range;
 double max_depth_range;
@@ -70,9 +71,14 @@ void callback(const sensor_msgs::ImageConstPtr& image_msg)
 
     }
 
-    if ((image_msg->header.stamp - g_last_wrote_time) < ros::Duration(1.0 / fps))
+    if ((image_msg->header.stamp - g_last_wrote_time) < ros::Duration((1.0 / fps)*(1.0-tolerance/100)))
     {
       // Skip to get video with correct fps
+      ROS_INFO_STREAM("Skipping frame "<< g_count <<": "
+                << image_msg->header.stamp - g_last_wrote_time << " sec is outside of the " 
+                << tolerance<< "\% tolerance");
+      
+      g_count++;
       return;
     }
 
@@ -108,6 +114,7 @@ int main(int argc, char** argv)
     bool stamped_filename;
     local_nh.param("stamped_filename", stamped_filename, false);
     local_nh.param("fps", fps, 15);
+    local_nh.param("tolerance", tolerance, 15.0);
     local_nh.param("codec", codec, std::string("MJPG"));
     local_nh.param("encoding", encoding, std::string("bgr8"));
     // cv_bridge::CvtColorForDisplayOptions


image_view/src/nodes/video_recorder.cpp
```diff
@@ -108,6 +114,7 @@ int main(int argc, char** argv)
     bool stamped_filename;
     local_nh.param("stamped_filename", stamped_filename, false);
     local_nh.param("fps", fps, 15);
+    local_nh.param("tolerance", tolerance, 15.0);
```
JWhitleyWork, line 117: Would you mind making this a value from 0-1 instead? Additionally, I'd prefer something more descriptive to indicate that it's a tolerance for the FPS rate (e.g. `fps_tolerance_pct`).

image_view/src/nodes/video_recorder.cpp
```diff
@@ -70,9 +71,14 @@ void callback(const sensor_msgs::ImageConstPtr& image_msg)
 
     }
 
-    if ((image_msg->header.stamp - g_last_wrote_time) < ros::Duration(1.0 / fps))
+    if ((image_msg->header.stamp - g_last_wrote_time) < ros::Duration((1.0 / fps)*(1.0-tolerance/100)))
     {
       // Skip to get video with correct fps
+      ROS_INFO_STREAM("Skipping frame "<< g_count <<": "
+                << image_msg->header.stamp - g_last_wrote_time << " sec is outside of the " 
+                << tolerance<< "\% tolerance");
```
JWhitleyWork, line 79: Since this is likely to happen often on some systems, I would prefer this be a `_THROTTLE` at minimum and a `_DEBUG` if you're OK with it instead of an `_INFO`. Terminal spam is not recommended.

image_view/src/nodes/video_recorder.cpp
```diff
@@ -70,9 +71,14 @@ void callback(const sensor_msgs::ImageConstPtr& image_msg)
 
     }
 
-    if ((image_msg->header.stamp - g_last_wrote_time) < ros::Duration(1.0 / fps))
+    if ((image_msg->header.stamp - g_last_wrote_time) < ros::Duration((1.0 / fps)*(1.0-tolerance/100)))
```
JWhitleyWork, line 74: This still doesn't sovle the case where the frame arrives early, only when it arrives late within the tolerance value. I recommend using `std::abs()` to handle both cases.

JWhitleyWork (changes requested):

image_view/src/nodes/video_recorder.cpp
cbteeple: I'm not sure I understand your concern. My new condition was specifically designed to handle the case where the frame arrives early. In the original code, the frame-skipping only happens when frames arrive too early (when `image_msg->header.stamp - g_last_wrote_time` is LESS THAN 1 frame duration). If frames are late, they are not skipped.

My proposed change is to only skip frames if they come WAY earlier than they should. This new conditions checks if the new frame arrived earlier than `1.0-tolerance` times the frame duration, which IS correct. 

The result: frames are never skipped unless they arrive earlier than the frame duration-tolerance.

Am I misunderstanding something here?

JWhitleyWork: Nope, you didn't miss anything - I did. You're correct. Thanks for the clarification.

image_view/src/nodes/video_recorder.cpp
cbteeple: I understand the concern over terminal spam, so I'd be fine with setting this to `_DEBUG`.

I'm not sure what you mean by "throttle at minimum"? Do you mean I should change the debug print statement to say "_Throttle at maximum framerate_" or something like that?

JWhitleyWork: Sorry, no, I meant using one of the `_THROTTLE` macros so the output of this message is throttled to once per time _t_ (see http://wiki.ros.org/roscpp/Overview/Logging - the Throttle section).

image_view/src/nodes/video_recorder.cpp
cbteeple: I agree with switching it to 0-1. Then the "tolerance" can be interpreted as a fraction of the frame duration (1/fps). In that case, can I call the parameter something like `fps_tolerance_fraction` or maybe `fps_variation_allowance`?

image_view/src/nodes/video_recorder.cpp
cbteeple: It seems like you might be confusing durations with rates here. When a frame comes early, it has a SHORTER duration since the last frame was saved. Thus the tolerance condition should check if the frame arrival duration is smaller than the frame duration MINUS some tolerance. If so, then the frame should be skipped, otherwise, the frame should not be skipped (including if it arrives late, right?).

image_view/src/nodes/video_recorder.cpp
cbteeple: Oh nice! That's super handy. I didn't know the `_THROTTLE` type existed. Okay, so then what's a reasonable period? Once per second? Before, there was no notice of skipped frames at all. Do you think once every 2 seconds is fine?

Another thing: I'm not sure if it matters but I also changed the way frames are counted here (before, the skipped frames were not counted in `g_count`. But I felt it made sense for `g_count` to be "every frame that has ever arrived" rather than just "every frame that was saved". Thoughts on this?

JWhitleyWork:
Ping @cbteeple for changes based on feedback.

clalancette:
There's been no response to the review on this PR for over a year, so I'm going to close this out.  If you'd still like to get this in, feel free to reopen and respond to feedback.

RoboCrafty:
I’ve implemented this tolerance and even tried setting a tire like 50% and 30% but there is still a lot of frame skipping. At times a five minute video turns out to be 2 minutes because of skipping.

My launch file:
```
 <launch>
    <node pkg="image_view" type="video_recorder" name="video_recorder0" args="_image_transport:=compressed"> 
        <remap from="image" to="/zed_node/rgb/image_rect_color" />
        <param name="filename" value="/home/nvidia/surveillance/ZED-Cam_$(arg timestamp).mp4" />
			<param name="codec" value="avc1" />
    		<param name="fps" value="11" />
    		<param name="tolerance" value="50" />
    </node>
 
    <node pkg="image_view" type="video_recorder" name="video_recorder3" args="_image_transport:=compressed">
        <remap from="image" to="/usb_cam_03/image" />
			<param name="codec" value="avc1" />
        <param name="filename" value="/home/nvidia/surveillance/Back-Cam_$(arg timestamp).mp4" />
    		<param name="fps" value="20" />
    		<param name="tolerance" value="50" />
    </node>
</launch>
 ```

PR #608, opened 2020-10-01
Make HAVE_GTK depend on finding gtk

lucasw:
It looks like gtk was removed entirely then restored a few years ago https://github.com/ros-perception/image_pipeline/commit/0faa2e14cbd6b26bbc884ac051e7e77948807a14 .

If gtk isn't actually optional, then the change should be to add a `required` to the `find_package(GTK2)`.

### DIFF
--- a/image_view/CMakeLists.txt
+++ b/image_view/CMakeLists.txt
@@ -6,10 +6,12 @@ generate_dynamic_reconfigure_options(cfg/ImageView.cfg)
 
 catkin_package(CATKIN_DEPENDS dynamic_reconfigure)
 find_package(Boost REQUIRED COMPONENTS thread)
+find_package(GTK2 REQUIRED)
 find_package(OpenCV REQUIRED)
 
 include_directories(${Boost_INCLUDE_DIRS}
                     ${catkin_INCLUDE_DIRS}
+                    ${GTK2_INCLUDE_DIRS}
                     ${OpenCV_INCLUDE_DIRS}
 )
 
@@ -37,12 +39,6 @@ install(TARGETS extract_images image_saver video_recorder
 if(ANDROID)
   return()
 endif()
-
-find_package(GTK2)
-if(GTK2_FOUND)
-add_definitions(-DHAVE_GTK)
-include_directories(${GTK2_INCLUDE_DIRS})
-endif(GTK2_FOUND)
 
 # Nodelet library
 add_library(image_view src/nodelets/image_nodelet.cpp src/nodelets/disparity_nodelet.cpp src/nodelets/window_thread.cpp)
--- a/image_view/CMakeLists.txt
+++ b/image_view/CMakeLists.txt
@@ -39,8 +39,10 @@ if(ANDROID)
 endif()
 
 find_package(GTK2)
+if(GTK2_FOUND)
 add_definitions(-DHAVE_GTK)
 include_directories(${GTK2_INCLUDE_DIRS})
+endif(GTK2_FOUND)
 
 # Nodelet library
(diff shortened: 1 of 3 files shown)


JWhitleyWork:
@lucasw Any reason this isn't targeting a newer release? It's preferred to have this target the latest (e.g. `noetic`) and then back-port if necessary. More established versions should be changed less often.

lucasw:
> Any reason this isn't targeting a newer release?

I use kinetic in an environment that doesn't have gtk, but not (yet) noetic.  I can make another PR for noetic, but kinetic is what I'd like this for.

lucasw:
The noetic cmakelists.txt doesn't define HAVE_GTK, instead new gtk3 find:

```
pkg_check_modules(GTK3 REQUIRED gtk+-3.0)
include_directories(${GTK3_INCLUDE_DIRS})
```

but `#ifdef HAVE_GTK` still occurs in the code in a couple files.  I wonder if that code is dead.

```
#if CV_MAJOR_VERSION == 2
#ifdef HAVE_GTK
    g_signal_connect(GTK_WIDGET( cvGetWindowHandle("left") ),
                     "destroy", G_CALLBACK(destroy), NULL);
    g_signal_connect(GTK_WIDGET( cvGetWindowHandle("right") ),
                     "destroy", G_CALLBACK(destroy), NULL);
    g_signal_connect(GTK_WIDGET( cvGetWindowHandle("disparity") ),
                     "destroy", G_CALLBACK(destroy), NULL);
#endif
    cvStartWindowThread();
#endif
```

JWhitleyWork:
Sorry for the delay.

Yes, that code is probably dead in `noetic`. However, I think the "optional" GTK dependency doesn't make sense here either. With ROS packages, it doesn't make sense to define an "optional" dependency unless you leave it out of `package.xml`. Since, in this case, `gtk2` is defined as a `build_depend` for `image_view`, it doesn't make sense to have optional code blocks since it's guaranteed to be there.

Would you mind modifying this MR to remove the `HAS_GTK` variable and un-protect all blocks that involve it?

lucasw:
I removed the `HAVE_GTK`, but in my environment I now have to `catkin config --blacklist image_view` so none of the changes are exercised (except I ran it once without the blacklist and it properly failed because of gtk2 is now set to `REQUIRED` in the image_view CMakeLists.txt).

Is there no github action for kinetic?  Otherwise what tests this?

image_view/CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 cmake_minimum_required(VERSION 2.8)
 project(image_view)
 
-find_package(catkin REQUIRED COMPONENTS camera_calibration_parsers cv_bridge dynamic_reconfigure image_transport message_filters message_generation nodelet rosconsole roscpp std_srvs stereo_msgs)
+find_package(catkin REQUIRED COMPONENTS camera_calibration_parsers cv_bridge dynamic_reconfigure gtk2 image_transport message_filters message_generation nodelet rosconsole roscpp std_srvs stereo_msgs)
```
JWhitleyWork: `gtk2` is not a `catkin` package but rather a system dependency so this isn't correct. This needs to be a separate

```
find_package(GTK2 REQUIRED)
```

JWhitleyWork (changes requested):

JWhitleyWork:
@lucasw Please pull the upstream `kinetic` into your `kinetic` and rebase your `optional_gtk_kinetic` branch on it. I've just added a basic CI build/test workflow to the `kinetic` branch.

image_view/CMakeLists.txt
lucasw: I added that and organized so GTK2 is similar to the OpenCV and Boost dependencies.

JWhitleyWork: Thank you!

PR #617, opened 2020-11-21
Add option for compressed images

Flova:
At the moment, no calibration is possible when using images that are provided as a `CompressedImage` message.
This pull request adds a command-line option to tell if the topic provides such a compressed message.

### DIFF
--- a/camera_calibration/nodes/cameracalibrator.py
+++ b/camera_calibration/nodes/cameracalibrator.py
@@ -65,6 +65,9 @@ def main():
     parser.add_option("-c", "--camera_name",
                      type="string", default='narrow_stereo',
                      help="name of the camera to appear in the calibration file")
+    parser.add_option("--compressed",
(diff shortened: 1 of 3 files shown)


JWhitleyWork:
While I understand the use-case, I think a simpler method is the one recommended in the [`image_transport` documentation](http://wiki.ros.org/image_transport#Python_Usage) which is using a `republish` node like this:

```
rosrun image_transport republish compressed in:=<in_base_topic> raw out:=<out_base_topic>
```

Re-using existing functionality rather than adding complexity/new features to an existing node is preferred.

If you have a specific use-case that requires this to be implemented in the node, feel free to re-open the MR for further discussion.

Flova:
Thanks for your comment. While the republish seems to be a reasonable fix for most cases it is not an option for my specific case, because the messages would be larger than the maximum size without any compression.

Issue #628, opened 2021-01-30
[ROS2] Frame match in point_cloud_xyzrgb.cpp


fmrico:
Hi all,

Is it really necessary that the RGB image and the depth image match here?

https://github.com/ros-perception/image_pipeline/blob/683d980f4682305aeef05f67ecb3b16c8b232a04/depth_image_proc/src/point_cloud_xyzrgb.cpp#L168

The Astra driver, for example, publishes both images in different frames, but coincident frames. If I remove this condition, it works like a charm.

![Captura de pantalla 2021-01-30 09:21:21](https://user-images.githubusercontent.com/3810011/106351656-ba6e0900-62dd-11eb-82f4-7ca0196d615b.png)

Maybe the fix could be to check if these frames coincide...

Best
Francisco

mikeferguson:
Looks like fix was merged

PR #629, opened 2021-01-31
Warning instead of Error when frames are differents

fmrico:
Hi all,

Sometimes the depth and RGB images are not produced by the same device, or the frames set by the driver are not the same. For example, the webots simulator and Orbbec Astra Driver (#628) use different devices for each one, with different frame id. 

This PR proposes to manage this situation as a warning, instead of an error, and continuing the processing.

I hope it will be considered :)
Francisco

### DIFF
--- a/depth_image_proc/src/point_cloud_xyzi.cpp
+++ b/depth_image_proc/src/point_cloud_xyzi.cpp
@@ -150,10 +150,9 @@ void PointCloudXyziNode::imageCb(
 {
   // Check for bad inputs
   if (depth_msg->header.frame_id != intensity_msg_in->header.frame_id) {
-    RCLCPP_ERROR(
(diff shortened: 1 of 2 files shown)


pmusau17:
Tested this locally on foxy and it looks good. Need guidance from the maintainers, in terms of what they prefer @JWhitleyWork

JWhitleyWork:
@fmrico Can you please make these `RCLCPP_WARN_THROTTLE`s? If not, you will get one for every single frame which is terminal spam. Once this is done, I'm OK to approve and merge this.

fmrico:
Hi @JWhitleyWork 

I have used RCLCPP_WARN_THROTTLE (10 secs). Is it ok for you?

Best

JWhitleyWork (approved):

PR #630, opened 2021-02-03
Support MONO16 image encodings.

mikepurvis:
Needed for working with RealSense depth images, which are MONO16.


JWhitleyWork (approved):
Doesn't seem like it would cause any problems.

PR #632, opened 2021-02-04
Add missing cstdint, vector, cmath includes.

mikepurvis:
All of these were missing for a case where we were pulling in this header.


JWhitleyWork (approved):
Sure. LGTM.

PR #635, opened 2021-02-09
Provide the option to use reliable subscribers

mjeronimo:
Allow a calling launch script to configure the use of reliable subscribers for the disparity and point cloud nodes.

Signed-off-by: Michael Jeronimo <michael.jeronimo@openrobotics.org>


jacobperron:
I think we can close this PR, opting for the rclcpp feature for enabling QoS overrides.

Also, it probably makes for a better user-experience if we default to "reliable" transport for all pub/subs in image_pipeline.

PR #641, opened 2021-03-17
Add retry feature on the image publisher node when no stream is available or if the the stream dropsout

ashwinsushil:
#### Aim

At times it could happen that the video stream is:
  1. Not ready at the beginning
  2. Drops out for a short duration 

In these cases, the image publisher node throws an error and is not able to recover if the stream gets ready or recovers. The idea is to retry creating the video capture if this happens.

#### Parameters

1. retry (bool) - If enabled, will retry the `onInit()` method.
2. timeout (int) - Time after which retrying starts

### DIFF
--- a/image_publisher/include/image_publisher/image_publisher.hpp
+++ b/image_publisher/include/image_publisher/image_publisher.hpp
@@ -64,6 +64,8 @@ private:
   std::string filename_;
   bool flip_horizontal_;
   bool flip_vertical_;
+  bool retry_;  // If enabled will retry loading image from the filename_
+  int timeout_;  // Time after which retrying starts
 
   std::string frame_id_;
   double publish_rate_;
--- a/image_publisher/src/image_publisher.cpp
+++ b/image_publisher/src/image_publisher.cpp
@@ -55,6 +55,8 @@ ImagePublisher::ImagePublisher(const rclcpp::NodeOptions & options)
   frame_id_ = this->declare_parameter("frame_id", std::string("camera"));
   publish_rate_ = this->declare_parameter("publish_rate", static_cast<double>(10));
   camera_info_url_ = this->declare_parameter("camera_info_url", std::string(""));
+  retry_ = this->declare_parameter("retry", false);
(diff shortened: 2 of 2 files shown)


ashwinsushil:
I don't seem to understand why [this](https://build.ros2.org/job/Rpr__image_pipeline__ubuntu_focal_amd64/57/) fails

pmusau17:
Tested this on foxy and it passes all tests.

JWhitleyWork:
@ashwinsushil I'm not sure this is the correct approach for fixing this problem. Instead, it would help to validate that the subscribers have associated publishers and vice versa with functions like [`Publisher::get_subscription_count()`](https://docs.ros2.org/foxy/api/rclcpp/classrclcpp_1_1PublisherBase.html#a60c84260b2ff0fb364fad688ed91234f) and [`Subscription::get_publisher_count()`](https://docs.ros2.org/foxy/api/rclcpp/classrclcpp_1_1SubscriptionBase.html#a0f61104666e697839925e0edc060359a). After those are confirmed, it's best to send a message and then verify receipt. See the `create_publisher()` and `create_subscription()` functions in Autoware's [`fake_test_node`](https://gitlab.com/autowarefoundation/autoware.auto/AutowareAuto/-/blob/master/src/tools/fake_test_node/include/fake_test_node/fake_test_node.hpp) for an example.

ashwinsushil:
@JWhitleyWork Thanks for the comment but maybe I'm not understanding the suggestion correctly here. The error happens because the stream link is not available even if there are publishers-subscribers available. The error happens while reading an image from a file https://github.com/ros-perception/image_pipeline/blob/372a1b22f275c7330e94e64feee6cc5510b24ad1/image_publisher/src/image_publisher.cpp#L159-L172

JWhitleyWork:
@ashwinsushil You're right - I misunderstood the purpose. Kicking CI again. Rolling failures will be ignored due to an upstream issue with `image_transport`.

JWhitleyWork (approved):
LGTM

PR #661, opened 2021-06-01
Add color param to stereo_image_proc

### LINKED ISSUE #660: Add parameter to disable color in point cloud node
When generating point clouds, there should be an option to disable color if RGB clouds are not needed. This should reduce the message size.

rebecca-butler:
Closes #660

### DIFF
--- a/stereo_image_proc/test/test_point_cloud_node.py
+++ b/stereo_image_proc/test/test_point_cloud_node.py
@@ -68,12 +68,28 @@ def generate_test_description():
             ],
             output='screen'
         ),
-        # PointCloudNode
+        # PointCloudNode (color enabled)
         Node(
             package='stereo_image_proc',
             executable='point_cloud_node',
             name='point_cloud_node',
-            output='screen'
+            output='screen',
+            parameters=[
+                {'use_color': True}
+            ],
+        ),
+        # PointCloudNode (color disabled)
+        Node(
+            package='stereo_image_proc',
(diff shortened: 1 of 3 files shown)


stereo_image_proc/test/test_point_cloud_node.py
```diff
@@ -68,12 +68,28 @@ def generate_test_description():
...
+            output='screen',
+            parameters=[
+                {'use_color': True}
+            ],
+        ),
+        # PointCloudNode (color disabled)
+        Node(
+            package='stereo_image_proc',
+            executable='point_cloud_node',
+            name='point_cloud_node',
```
jacobperron: I would give this node a different name (or a namespace) to differentiate it from the other. If we use a namespace, then we can probably remove the remappings  argument below since the `points2` topic should appear in the namespace.

jacobperron (commented):
The changes look good to me! I'd like to test this locally before giving my approval and merging.

stereo_image_proc/test/test_point_cloud_node.py
rebecca-butler: I renamed the node, but didn't add a namespace as remappings would be required either way.

jacobperron (approved):
LGTM! I don't think the CI failures are related.

JWhitleyWork (approved):
LGTM. Thanks for the PR!

PR #662, opened 2021-06-04
clean up ROS2 depth_image_proc, add xyzrgb radial component

flynneva:
this PR does a few things:
- clean up `depth_image_proc` to a more standard .cpp / .hpp style
- move `depth_conversions.hpp` to just `conversions.hpp` and corresponding `.cpp` to hold other methods for `convertIntensity` and `convertRgb`
- add `xyzrgb_radial` component

still need to fix the lint errors, but **i think** this should be functional as-is.


flynneva:
@JWhitleyWork it looks like the rolling and galactic tests are unstable (but still passing I think?)

peci1:
Hi, I'm not reviewing your PR (just stumbled upon it from Discourse and I don't have enough knowledge of this package to be a good reviewer), but I've seen a few things that might speed up its review by others:

1. do not rename existing .h/.hpp files. They're part of the API and their renaming breaks downstream code. If there's a strong will for renaming, it should be done in a separate PR and only for unreleased ROS versions.
2. provide some images/screenshots of the results of the radial nodes

depth_image_proc/include/depth_image_proc/visibility.hpp
```diff
@@ -29,8 +29,8 @@
 // LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
 // ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 // POSSIBILITY OF SUCH DAMAGE.
-#ifndef DEPTH_IMAGE_PROC__VISIBILITY_H_
```
clalancette, line 32: There is actually nothing C++ specific about this header, so let's just leave it as a `.h` file.

depth_image_proc/src/point_cloud_xyz.cpp
```diff
@@ -29,50 +29,13 @@
 // LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
 // ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 // POSSIBILITY OF SUCH DAMAGE.
-#include <rclcpp/rclcpp.hpp>
-#include <image_transport/image_transport.hpp>
-#include <sensor_msgs/image_encodings.hpp>
-#include <image_geometry/pinhole_camera_model.h>
-#include <depth_image_proc/depth_conversions.hpp>
-#include <depth_image_proc/visibility.h>
```
clalancette, line 37: I am never in favor of this.  Every file should include what it uses; otherwise, you are just implicitly relying on headers files from elsewhere (which may change).  The same goes several times below.

clalancette (commented):
I left a couple of comments inline.

Overall, I think this would be easier to review if it was split into 2 or 3 PRs, roughly corresponding to each of the points in your initial description.  That way we can do a refactor (no change in behavior), rename files (with some backwards-compatibility header files thrown in), and then finally add the new stuff that you want to add.

I'm not the maintainer here, though, so I'll just leave this as a comment rather than "Request changes".

flynneva:
@clalancette thanks for the tips! I'll see about splitting up these changes into different PRs as you said.

flynneva:
> Hi, I'm not reviewing your PR (just stumbled upon it from Discourse and I don't have enough knowledge of this package to be a good reviewer), but I've seen a few things that might speed up its review by others:
> 
> 
> 
> 1. do not rename existing .h/.hpp files. They're part of the API and their renaming breaks downstream code. If there's a strong will for renaming, it should be done in a separate PR and only for unreleased ROS versions.
> 
> 2. provide some images/screenshots of the results of the radial nodes


Good tips, thanks! And I only added the 'xyzrgb_radial' node, the others were already there. As for screenshots I'll see if I can get one today of the radial node I added.

depth_image_proc/src/point_cloud_xyz.cpp
flynneva: i dont really follow this logic. would you mind linking to some more in-depth info on this? I thought the point of header files was to define the object and its methods. without the header file, the cpp file is useless if you use this structure (which I thought was a cpp standard...?).

it sounds like what you're saying is to include the dependencies in the cpp file as well AND include the header file that defines the object. right? i guess ive been doing this wrong then.....

depth_image_proc/include/depth_image_proc/visibility.hpp
flynneva: @clalancette i thought with ROS2 we were [supposed to migrate to use `.hpp`](https://industrial-training-master.readthedocs.io/en/melodic/_source/session7/ROS1-to-ROS2-porting.html) (number 7 in that link)

i thought the approach to handle this would be to keep the `.h` file, but just leave it blank and include the new `.hpp` standard?

clalancette: We aren't totally consistent on this, but in my view it should only be `.hpp` if it is a C++-only header file.  Leaving it as `.h` where it works in C means you can use this header file for C files as well (which is admittedly unlikely).

depth_image_proc/src/point_cloud_xyz.cpp
clalancette: > it sounds like what you're saying is to include the dependencies in the cpp file as well AND include the header file that defines the object. right?

Correct.

depth_image_proc/include/depth_image_proc/visibility.hpp
flynneva: so is this just your view or is it the ROS2 standard? Looking at the documentation it seems that the standard is to migrate to `.hpp`.  @peci1 makes a good point though that this change might break existing code, so what i've seen before is a blank `.h` file that includes the `.hpp` file, so "new" code can use the new `.hpp` standard, while old code wont be affected.

JWhitleyWork: ROS2 standard is `.hpp`. Since this is only going to the Galactic branch, I think changing them is fine as very few users will be affected.

depth_image_proc/src/point_cloud_xyz.cpp
JWhitleyWork: This is the Include-What-You-Use principal. I tend to agree with this as it's part of the Google Style Guide, which the ROS2 Style Guide is based on. However, you should only include those files which you use in the appropriate file. For example, if you reference a message type in a function signature, it should be included in both the header and the source file (because it's used in both places). However, if you only use `std::sort` in the source file, only include `<algorithm>` in the source file.

depth_image_proc/include/depth_image_proc/visibility.hpp
clalancette: > so is this just your view or is it the ROS2 standard? Looking at the documentation it seems that the standard is to migrate to `.hpp`. @peci1 makes a good point though that this change might break existing code, so what i've seen before is a blank `.h` file that includes the `.hpp` file, so "new" code can use the new `.hpp` standard, while old code wont be affected.

Yeah, and we do that elsewhere we we ported ROS 1 C++ code that was in a `.h` file, and now we want to move it to `.hpp` to have it conform to ROS 2 standards.  In this case, though, there is really no need for the churn; the existing `.h` file will work for both C and C++.

> ROS2 standard is `.hpp`.

"ROS 2 standard is `.hpp` *for C++ code*".  Since this isn't purely C++ code (it is really C code that C++ understands), there's no reason in my view to change it.

That all said, this is a bit of bikeshedding and doesn't matter a huge amount at the end of the day.  I think it should be left as a `.h` file, but if the maintainers disagree we can change it.

JWhitleyWork: While I *mostly* agree with you, the kicker here is `extern "C"`. This only has meaning (and will ony compile) with a C++ compiler, not a C compiler. If we want to remove that and make the file a "pure C" header, then it should be included from other files as:

```c++
extern "C" {
#include <visibility.h>
}
```

Otherwise, as it is, it's a C++ header.

clalancette:
```
#ifdef __cplusplus
extern "C"
{
#endif
```

Makes it work for both C and C++.

JWhitleyWork: You have a point.

flynneva:
opened up the first PR off of this one #665  that only cleans up the structure of the existing code base.

PR #671, opened 2021-06-19
support rgba8 and bgra8 encodings by skipping alpha channel

80avin:
Added support for RGBA8 and BGRA8 encodings, similar to RGB8 & BGR8 and ignoring the alpha channel.
It is required because Images provided by CARLA simulator have BGRA8 encoding, and even when the image have sufficient data, it didn't work with it.


JWhitleyWork (approved):
LGTM.

PR #672, opened 2021-06-21
Fix time-stamps in Video Recorder

larsll:
Proposed fix for #673

### DIFF
--- a/image_view/src/video_recorder_node.cpp
+++ b/image_view/src/video_recorder_node.cpp
@@ -34,7 +34,7 @@ namespace image_view
 VideoRecorderNode::VideoRecorderNode(const rclcpp::NodeOptions & options)
 : rclcpp::Node("video_recorder_node", options),
   g_count(0),
-  g_last_wrote_time(rclcpp::Time((int64_t) 0, RCL_ROS_TIME)),
+  g_last_wrote_time(rclcpp::Time((int64_t) 0, RCL_SYSTEM_TIME)),
   recording_started(false)
 {
   bool stamped_filename;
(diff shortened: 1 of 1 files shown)


JWhitleyWork:
@larsll Is there a specific reason you went with `RCL_SYSTEM_TIME` instead of `RCL_ROS_TIME`? The latter allows for the use of simulated time with the `/clock` topic while the former does not. If you are replaying a bag and making a recording from it, I would assume that you would want to at least have the option to use the timestamps from the bag instead of only ever being able to use the system time.

larsll:
> @larsll Is there a specific reason you went with `RCL_SYSTEM_TIME` instead of `RCL_ROS_TIME`? The latter allows for the use of simulated time with the `/clock` topic while the former does not. If you are replaying a bag and making a recording from it, I would assume that you would want to at least have the option to use the timestamps from the bag instead of only ever being able to use the system time.

Either option works for me - in the end I don't think it matters as the values are only used within video_recorder, and only used to ensure the right fps.

The perfect solution might be to detect what kind of clock you are receiving, but I am not sure how to do that 'correctly' based on image_msg->header.stamp.

JWhitleyWork:
`RCL_ROS_TIME` will automatically use either the system clock or the rosbag `/clock` topic if it is available and `use_sim_time: true` is set on the node. Please use this instead.

pmusau17:
Made requested change to RCL_ROS_TIME. [#675](https://github.com/ros-perception/image_pipeline/pull/675)

JWhitleyWork:
Closing in favor of #675.

PR #669, opened 2021-06-17
Add missing test dependency

jacobperron:
I'm hope this will fix an issue with CI.

PR #659, opened 2021-05-21
Replace deprecated geometry2 headers

jacobperron:
tf2_geometry_msgs.h was deprecated in https://github.com/ros2/geometry2/pull/418
tf2_eigen.h was deprecated in https://github.com/ros2/geometry2/pull/413

Note, we might want to hold this PR until the referenced changes in geometry2 are released into Rolling.

PR #674, opened 2021-06-23
Update deprecated headers

pmusau17:
Fix for stereo_image_proc tests. Tested using rolling docker image. Passes all tests. 
This PR is based on https://github.com/ros-perception/image_pipeline/pull/659/files by @jacobperron and will fix the timeout error in https://github.com/ros-perception/image_pipeline/pull/669

### DIFF
--- a/depth_image_proc/src/register.cpp
+++ b/depth_image_proc/src/register.cpp
@@ -40,7 +40,7 @@
 #include <sensor_msgs/image_encodings.hpp>
 #include <image_geometry/pinhole_camera_model.h>
 #include <Eigen/Geometry>
-#include <tf2_eigen/tf2_eigen.hpp>
+#include <tf2_eigen/tf2_eigen.h>
 #include <depth_image_proc/depth_traits.hpp>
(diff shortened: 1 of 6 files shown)


jacobperron:
Cool, thanks for the fixes! I'll try to merge #659 separately, so I don't think we need to include that change in this PR. 
 
I'm curious why we need to set `approximate_sync` to true, since we are using the exact same timestamp for inputs, e.g.

https://github.com/ros-perception/image_pipeline/blob/6257bda1449a6a9ffb918f05f0b92ffbfee6a623/stereo_image_proc/test/fixtures/stereo_image_publisher.py#L97-L101

Maybe it was `use_system_default_qos` that fixed the flake?

Also, are package.xml changes to image_publisher and image_view relevant?

pmusau17:
For some reason it failed at first without those additions. I just retested it on galactic and rolling and it works. Also looks like it was  ```use_system_default_qos``` that fixed it. I just updated the commit.

jacobperron:
Please revert a293252; I don't think it is compatible with Foxy or Galactic.

PR #679, opened 2021-07-07
Replace deprecated geometry2 headers

clalancette:
tf2_geometry_msgs.h was deprecated in https://github.com/ros2/geometry2/pull/418
tf2_eigen.h was deprecated in https://github.com/ros2/geometry2/pull/413

Signed-off-by: Jacob Perron <jacob@openrobotics.org>

Note that this is an exact duplicate of #659, to see if we can get around the current CI issues.


clalancette:
Closing.

PR #363, opened 2018-11-03
[depth_image_proc/register] option to fill sparsity holes on upsampling

ipa-rmb:
Added the option to fill the sparse areas with nearest neighbor depth values on upsampling operations in depth_image_proc/register . We make use of this function quite frequently. The PR does not alter the standard behavior.

PR #692, opened 2021-08-20
Port upsampling interpolation from #363 to ROS2

schornakj:
Add the algorithm updates and new parameters from #363 (which is included in the `noetic` branch, but not the ROS2 branches) to allow interpolating the upscaled depth image when registering it to an image of higher resolution.

Like the original PR, this includes a bool parameter that is `false` by default to preserve the original functionality of this node. However, I'm not sure if there's a use case in which a user would want to purposefully leave the gaps in the depth image.

### DIFF
--- a/depth_image_proc/src/register.cpp
+++ b/depth_image_proc/src/register.cpp
@@ -241,8 +241,7 @@ void RegisterNode::convert(
 
       double depth = DepthTraits<T>::toMeters(raw_depth);
 
-      if (fill_upsampling_holes_ == false)
-      {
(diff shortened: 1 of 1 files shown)


JWhitleyWork (approved):

clalancette:
The CI is failing here because cpplint is failing, which has nothing to do with this PR.  Thus, I'm going to go ahead and merge this anyway, and we'll clean up cpplint in a separate PR.

clalancette (approved):

PR #693, opened 2021-08-20
Fix ResizeNode to allow remapping both image and camera_info topics

schornakj:
Due to how `image_transport::create_camera_subscription` creates a `camera_info` topic from the provided image topic name, the current implementation of `image_proc::ResizeNode` ends up using the same topic name `/camera_info` for both the input image subscriber and the output resized image publisher. This prevents remapping the input and output CameraInfo topics.

This change makes `image` and `resize` topic namespaces instead of image topic names, which allows the node to correctly separately use `/image/camera_info` and `/resize/camera_info`.

### DIFF
--- a/image_proc/src/resize.cpp
+++ b/image_proc/src/resize.cpp
@@ -51,10 +51,10 @@ ResizeNode::ResizeNode(const rclcpp::NodeOptions & options)
 : rclcpp::Node("ResizeNode", options)
 {
   // Create image pub
-  pub_image_ = image_transport::create_camera_publisher(this, "resize");
+  pub_image_ = image_transport::create_camera_publisher(this, "resize/image_raw");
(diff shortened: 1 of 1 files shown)


afrixs:
This resolves #703 . Linking the issue to maybe attract maintainers, because this is a crucial bug fix to make the node usable and it's been hanging here unmerged for 6 months...

schornakj:
Based on who's been merging recent PRs, I think @JWhitleyWork is the right person to tag here for visibility.

JWhitleyWork:
Sorry for the incredibly long delay. LGTM.

schornakj:
> Sorry for the incredibly long delay. LGTM.

Thank you for following up! I just created #786 to backport this fix to Humble as well.

PR #697, opened 2021-08-27
Fix image saver bug and time-based image saving

jacobperron:
Add missing 'request_start_end' ROS parameter (from ROS 1) and ensure related
member variables are initialized properly. This restores the time-based image
saving feature from ROS 1 and fixes a bug where member variables
'save_image_service' and 'request_start_end' were not initialized properly
which could lead to images not properly saving.

I've also renamed some member variables to have a trailing underscore, for
consistency with other members.


JWhitleyWork (approved):
Seems reasonable.

PR #710, opened 2021-11-12
[rolling] Populate CameraInfo camera matrix in test fixture

jacobperron:
By adding non-zero values we avoid a divide by zero during stereo processing and avoid an issue
leading to rclpy to crash: https://github.com/ros2/rclpy/issues/848


JWhitleyWork:
Nice! Thanks for the fix!

PR #712, opened 2021-11-18
fix publisher issues

harsha-vk:
crop_decimate: fixed same topic names for publisher and subscriber. Commented out getNumSubscribers line which is not allowing publisher to publish messages even when there is a subscriber.

### DIFF
--- a/image_proc/src/crop_decimate.cpp
+++ b/image_proc/src/crop_decimate.cpp
@@ -119,9 +119,9 @@ CropDecimateNode::CropDecimateNode(const rclcpp::NodeOptions & options)
   int interpolation = this->declare_parameter("interpolation", 0);
   interpolation_ = static_cast<CropDecimateModes>(interpolation);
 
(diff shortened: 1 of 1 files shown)


image_proc/src/crop_decimate.cpp
```diff
@@ -131,9 +131,9 @@ void CropDecimateNode::imageCb(
 {
   /// @todo Check image dimensions match info_msg
 
-  if (pub_.getNumSubscribers() < 1) {
-    return;
-  }
+  //if (pub_.getNumSubscribers() < 1) {
+    //return;
+  //}
```
JWhitleyWork, line 136: We want this code to be here. It bypasses publishing unless at least one subscriber is present on the topic.

JWhitleyWork (changes requested):
Dashing is no longer supported. Please target this at a newer release (`foxy`, `galactic` or `rolling`).

PR #714, opened 2021-11-26
Delete line 330 in calibrator.py 

cgdubois:
Unable to selected fisheye option during camera_calibration since CAMERA_MODEL.PINHOLE is hardcoded on line 330. This should be set by set_cammodel. Removing this line allows switching between the two options via the GUI.


JWhitleyWork:
I don't believe that deleting this line is the correct solution. This either needs to be passed in as a parameter to the `Calibrator` class or needs to be set as a default by line 330 and be overridden later, which can be done by calling `set_cammodel`. If you are interested in implementing one of these solutions, please let me know. Otherwise, I will close the PR.

PR #715, opened 2021-12-06
Printout fixes (topic names, ros2 help), exception fixes

guilyx:
- Rewrite some of the `rosrun` help left from ROS1
- Use `transport` as a ros arg parameter instead of a cli arg
    - Before: `ros2 run image_view extract_images transport:=raw` would break if adding `--ros-args -r image:=image_raw`
    - Now: `ros2 run image_view extract_images --ros-args -r image:=image_raw -p transport:=raw` which uses the API's capabilities
- `stereo_view_node` would throw an exception because of a double trailing slash in the topic name

### DIFF
--- a/image_view/src/disparity_view_node.cpp
+++ b/image_view/src/disparity_view_node.cpp
@@ -70,7 +70,7 @@ DisparityViewNode::DisparityViewNode(const rclcpp::NodeOptions & options)
   if (topic == "image") {
     RCLCPP_WARN(
       this->get_logger(), "Topic 'image' has not been remapped! Typical command-line usage:\n"
-      "\t$ rosrun image_view disparity_view image:=<disparity image topic>");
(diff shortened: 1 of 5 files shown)


JWhitleyWork:
@guilyx It looks like you're still making some fixes to this branch. Please let me know when you are done and it is ready for review.

guilyx:
@JWhitleyWork good to go! 👍🏾

JWhitleyWork:
Your changes to `image_view` are causing it to fail the `uncrustify` linter. Please go into the `image_view` folder and run `ament_uncrustify --reformat .` to fix these problems. Then, go up to the base of your workspace and run `colcon test` and verify that all tests pass with `colcon test-result --verbose`.

guilyx:
@JWhitleyWork should be good

JWhitleyWork (approved):
LGTM

PR #716, opened 2021-12-20
Update ROI parameters on resize

furushchev:
On resizing image, ROI parameters in camera info message should be scaled as same as other fields in the message.


JWhitleyWork (approved):

JWhitleyWork:
@furushchev Thanks for this! Would you mind porting this to `ros2` as well?

furushchev:
@JWhitleyWork I still don't use ros2, but please go ahead! (If we need to do anything, just let me know! :-) )

Issue #696, opened 2021-08-26
[depth_image_proc] [ROS2] symbol lookup error


Kettenhoax:
When I build depth_image_proc version 6b649d2b592d6c07461f9232f86e186017fb4442 in galactic, and run the `depth_image_proc::PointCloudXyzNode` component, I get a symbol lookup error:

```bash
/opt/ros/galactic/lib/rclcpp_components/component_container: symbol lookup error: ../lib/libdepth_image_proc.so: undefined symbol: _ZN16depth_image_proc12convertDepthIfEEvRKSt10shared_ptrIKN11sensor_msgs3msg6Image_ISaIvEEEERS1_INS3_12PointCloud2_IS5_EEERKN14image_geometry18PinholeCameraModelEd
```

It seems that 635bbb94a7ca72be11bb914a736b5052a6ec0b8d introduced changes in which the definition of the convertDepth function, which seems to be the missing symbol, is moved to a cpp file. PointCloudXyzNode only includes depth_image_proc/conversions.hpp.
Am I missing something, or is the template never instantiated when building PointCloudXyzNode, and the symbol never generated?

shrini-mm:
I am facing the same problem. Is there any update on this?

dcconner:
Ditto with Foxy build

shrini-mm:
I fixed it by installing the apt package of `depth_image_proc` that came with  ubuntu 20.04. But I am not sure how to fix it yet by doing our own build. Hope it helps.

quentingllmt:
For those looking, applying https://github.com/ros-perception/image_pipeline/pull/718 fixed the runtime linker error for me on the ros2 branch.

flynneva:
@Kettenhoax @shrini-mm @dcconner this should now be covered with #757 backporting the fix from #718. Sorry for causing so much trouble for everyone 😓 

fyi @JWhitleyWork

JWhitleyWork:
Should be fixed on `ros2` branch now. Don't worry about it, @flynneva - it happens to everyone!

PR #718, opened 2022-01-04
Fix linker error caused by templating in the conversions.cpp file

flynneva:
Fixes linker error found by @simwijs on [this thread here](https://github.com/ros-perception/image_pipeline/commit/635bbb94a7ca72be11bb914a736b5052a6ec0b8d#commitcomment-62905772)

Essentially I made the mistake of including template implementations in the .cpp file when they all need to be in the header.

Signed-off-by: Evan Flynn <evan.flynn@apex.ai>

### DIFF
--- a/depth_image_proc/include/depth_image_proc/conversions.hpp
+++ b/depth_image_proc/include/depth_image_proc/conversions.hpp
@@ -48,20 +48,94 @@ void convertDepth(
   const sensor_msgs::msg::Image::ConstSharedPtr & depth_msg,
   sensor_msgs::msg::PointCloud2::SharedPtr & cloud_msg,
   const image_geometry::PinholeCameraModel & model,
-  double range_max = 0.0);
+  double range_max = 0.0)
+{
+  // Use correct principal point from calibration
+  float center_x = model.cx();
+  float center_y = model.cy();
+
+  // Combine unit conversion (if necessary) with scaling by focal length for computing (X,Y)
+  double unit_scaling = DepthTraits<T>::toMeters(T(1) );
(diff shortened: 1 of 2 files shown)


flynneva:
@simwijs let me know if this fix solves your issue - I tested it locally using the cmake flags you mentioned and it seemed to work I think.

simwijs:
You're fast lol :+1: 
I'll test it right away!

simwijs:
Works beautifully. You have my :1st_place_medal: 
Nice to see an engineer in action :) Keep moving forward! Also glad to have contributed something, even though you made the fix. I learned from it! (I spent way too much time finding this error and trying to fix it when you fixed it in minutes haha)

flynneva:
@simwijs you give me too much credit 😅 but glad to help out!

This is what open source is all about - finding bugs you otherwise would miss and fixing them together.

flynneva:
@JWhitleyWork friendly ping, fixes a hidden bug in the `ros2` `depth_image_proc` package (template implementation in cpp file instead of all in the header)

clalancette (approved):

clalancette:
CI is failing, but that has nothing to do with this PR.  We'll fix that up separately.  This looks like a good fix to me, so merging it in.

JWhitleyWork:
@flynneva Would you mind back-porting this to Foxy and Galactic? It appears to be present there too based on #696.

flynneva:
@JWhitleyWork id be happy to! Sorry for causing such a stink with this.

JWhitleyWork:
@flynneva Don't worry about it! As long as the outcome is improvement then the journey is always worth it!

PR #720, opened 2022-01-04
Support sensor data QOS on image_proc

coalman321:
Adds support for image_proc nodes to enable Sensor Data QOS profile on input and output image topics. This fix is designed to allow camera data to come in from the gazebo_ros camera plugin which uses Sensor Data QOS by default. This also patches an issue with the stereo_image_proc launch file where the right input topic was not re-mapped in the launch file, and only the left topic was remapped. 

This PR has already been linted via `colcon test` as well as `ament_uncrustify --reformat`. The changes have no impact on the default behavior of all changed nodes. All changes have been tested against ROS2 galactic and have been confirmed to work as intended.

### DIFF
--- a/image_proc/include/image_proc/rectify.hpp
+++ b/image_proc/include/image_proc/rectify.hpp
@@ -56,6 +56,7 @@ private:
   image_transport::CameraSubscriber sub_camera_;
 
   int queue_size_;
+  bool use_qos_;
   int interpolation;
   std::mutex connect_mutex_;
   image_transport::Publisher pub_rect_;
--- a/image_proc/launch/image_proc.launch.py
+++ b/image_proc/launch/image_proc.launch.py
@@ -46,6 +46,9 @@ def generate_launch_description():
             package='image_proc',
             plugin='image_proc::DebayerNode',
             name='debayer_node',
+            parameters=[{
+                'use_system_default_qos': LaunchConfiguration('use_system_default_qos'),
+            }],
         ),
         ComposableNode(
             package='image_proc',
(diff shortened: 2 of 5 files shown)


image_proc/include/image_proc/rectify.hpp
```diff
@@ -56,6 +56,7 @@ class RectifyNode
   image_transport::CameraSubscriber sub_camera_;
 
   int queue_size_;
+  bool use_qos_;
```
JWhitleyWork, line 59: Is there a reason this was implemented as a member variable in `rectify` but a temporary in `debayer`?

coalman321: Apologies for taking a minute on this, but it was due to it being needed in two separate methods, the constructor and subscribeToCamera. It could be instead be a local value and retrieved multiple times, or passed into the subscribeToCamera method if those seem like better paradigms.

JWhitleyWork: I get it now. Thanks for the clarification. It's fine.

JWhitleyWork:
@coalman321 `cpplint` is showing errors. Please correct these before I merge.

coalman321:
> @coalman321 `cpplint` is showing errors. Please correct these before I merge.

Looking at the build farm logs, these are all issues with include order. I did not have any changes made to them in this commit, so this would have been failing prior. I can fix them, but they may have already been fixed.

mikeferguson:
Assigning to myself to look at this as part of https://github.com/ros-perception/image_pipeline/issues/847

mikeferguson:
Superseded by #922 - which implements the newer QoS overriding approach (which is now the default throughout ROS2). Our old parameter was actually listed as an "ad-hoc" work around in the design document for this feature: http://design.ros2.org/articles/qos_configurability.html

PR #722, opened 2022-01-17
depth_image_proc register cover all angles with fill_upsampling_holes

### LINKED ISSUE #721: Make depth_image_proc register fill_upsampling_holes work with full range of relative angles
The current implementation doesn't draw anything from certain relative angles between the source depth image frame and target frame (called rgb in the code and documentation but it is just any valid camera info with a valid frame), draws increasingly thin slivers as it approaches those angles, and then when a range check fails it doesn't draw anything at all.

I have the beginning of a quick fix here https://github.com/lucasw/image_pipeline/commit/478fd5bf4ac5a0229e33e1ca7ff7c8e226f00124 but it'll have to do 3 or 4x the computation.  It's not the full triangle rasterization mentioned in a `todo` but would get decent results in more cases.  It gets a little ridiculous to embed what looks more and more like a full software 3D renderer here into this node but making it a little better seems reasonable.

register_upsampling_slivers

This image shows off the filled areas getting reduced to slivers using the min/max of two values approach in the commit above, without it the slivers can get reduced to nothing at all because the range check can fail entirely.

This is the original code with for loops that sometimes don't execute at all:
https://github.com/ros-perception/image_pipeline/blob/noetic/depth_image_proc/src/nodelets/register.cpp#L291-L293

There look to be some other obvious refactors/cleanups to do in that same code (it's doing the same thing in three nearly identical blocks).  Also I can look at creating a unit test for this (or is there one somewhere and I'm not seeing it?).

Probably few people who are using this have crazy high angles between the depth camera and the other camera frame, I'm only running into it now trying this out with synthetic data (I may get some of the synthetic data generation into the unit test, but it's nice to have manual test support also).

lucasw:
Resolves #721

This is without the PR (the rviz CameraInfo plugin looks to not draw lines properly from certain angles)

https://user-images.githubusercontent.com/1334122/149801865-d275bfec-a51d-4a97-941e-8f6c0c83c2da.mp4

This is with the PR applied:

https://user-images.githubusercontent.com/1334122/149801889-8defd1e5-6062-47b2-82a9-c0d6f743c6f9.mp4

Here's another but without the negative z clipping
https://user-images.githubusercontent.com/1334122/149797323-fc6bf43f-3d12-47ba-9d99-fb334f4c220e.mp4

There are still gaps when there is a big angle between the two tf frames or big depth difference between adjacent pixels/points, making assumptions about adjacent points (that they are part of a continuous surface) and doing the whole software triangle rasterization approach could fix that but seems excessive.

I found adding a lookup transform timeout a62174afd357152af57a58502d8c7b205fc5e131 necessary in my test setup (probably in most real world cases `/tf` is way ahead of >megapixel depth images, but here I'm synthesizing 320x240 depth images so few image transport delays) but left it out of this PR, it could go into a different one?

I can adapt the test setup to here if there is interest, but it will bring in a few `<test_depends>`.

Could squash this down to one or two commits but maybe the commit progression makes it easier to review.

There is a really basic unit test, I can add more to it after getting some feedback.

### DIFF
--- a/depth_image_proc/src/nodelets/register.cpp
+++ b/depth_image_proc/src/nodelets/register.cpp
@@ -285,6 +285,9 @@ void RegisterNodelet::convert(const sensor_msgs::ImageConstPtr& depth_msg,
           continue;
 
         // TODO(lucasw) this is identical to above except adding 0.5 instead of subtracting
+        // TODO(lucasw) need a third and fourth position with u+0.5 and v-0.5 and u-0.5 and v+0.5
+        // (and then could do the triangle rastering mentioned above, or just take the min
+        // and max of all of them)
         xyz_depth_2 << ((u+0.5f - depth_cx)*depth - depth_Tx) * inv_depth_fx,
                        ((v+0.5f - depth_cy)*depth - depth_Ty) * inv_depth_fy,
                        depth,
@@ -302,9 +305,9 @@ void RegisterNodelet::convert(const sensor_msgs::ImageConstPtr& depth_msg,
 
         // fill in the square defined by uv range
         const T new_depth = DepthTraits<T>::fromMeters(0.5*(xyz_rgb_1.z()+xyz_rgb_2.z()));
-        for (int nv=v_rgb_1; nv<=v_rgb_2; ++nv)
+        for (int nv=std::min(v_rgb_1, v_rgb_2); nv<=std::max(v_rgb_1, v_rgb_2); ++nv)
         {
-          for (int nu=u_rgb_1; nu<=u_rgb_2; ++nu)
+          for (int nu=std::min(u_rgb_1, u_rgb_2); nu<=std::max(u_rgb_2, u_rgb_2); ++nu)
           {
             T& reg_depth = registered_data[nv*registered_msg->width + nu];
             // Validity and Z-buffer checks
--- a/depth_image_proc/src/nodelets/register.cpp
+++ b/depth_image_proc/src/nodelets/register.cpp
@@ -247,9 +247,6 @@ bool transform_depth(
   return true;
 }
 
-// TODO(lucasw) need a unit test for this, simple low res image (e.g. 4x4 pixels) with
-// depth ranges of around 1.0
-// shift it to the right 1.0 units, view from 90 degrees
 template<typename T>
 void RegisterNodelet::convert(const sensor_msgs::ImageConstPtr& depth_msg,
                               const sensor_msgs::ImagePtr& registered_msg,
--- a/depth_image_proc/src/nodelets/register.cpp
+++ b/depth_image_proc/src/nodelets/register.cpp
@@ -195,6 +195,9 @@ void RegisterNodelet::imageCb(const sensor_msgs::ImageConstPtr& depth_image_msg,
   pub_registered_.publish(registered_msg, registered_info_msg);
 }
 
+// TODO(lucasw) need a unit test for this, simple low res image (e.g. 4x4 pixels) with
+// depth ranges of around 1.0
+// shift it to the right 1.0 units, view from 90 degrees
 template<typename T>
 void RegisterNodelet::convert(const sensor_msgs::ImageConstPtr& depth_msg,
                               const sensor_msgs::ImagePtr& registered_msg,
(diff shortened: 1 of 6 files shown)


depth_image_proc/test/test_register.py
```diff
@@ -0,0 +1,119 @@
...
+        while i in range(6):
+            if self.count > 1:
+                break
+            rospy.sleep(1)
+        rospy.loginfo("done waiting")
+
+    def depth_callback(self, msg):
+        rospy.loginfo("received depth")
+        registered_depth = self.cv_bridge.imgmsg_to_cv2(msg, "32FC1")
+        self.assertAlmostEqual(registered_depth[1, 1], 7.0, 1)
```
lucasw: pass in a z offset parameter from the launch file instead of the 7.0 magic number that depends on the static tf publishers and the generated points above

depth_image_proc/test/test_register.py
```diff
@@ -0,0 +1,119 @@
...
+        depth = np.empty((ht, wd, 1), np.float32)
+        depth[0, 0] = 1.0
+        depth[0, 1] = 1.0
+        depth[0, 2] = 1.0
+        depth[1, 0] = 1.0
+        depth[1, 1] = 1.0
+        depth[1, 2] = 1.0
+        depth[2, 0] = 1.0
+        depth[2, 1] = 1.0
+        depth[2, 2] = 1.0
```
lucasw: Just do np.ones

depth_image_proc/src/nodelets/register.cpp
```diff
@@ -207,93 +265,100 @@ void RegisterNodelet::convert(const sensor_msgs::ImageConstPtr& depth_msg,
...
+          }
+        }
+
+        if (count == 0) {
           continue;
+        }
 
-        for (int nv=v_rgb_1; nv<=v_rgb_2; ++nv)
+        // fill in the square defined by uv range
+        const T new_depth = DepthTraits<T>::fromMeters(new_depth_sum / static_cast<double>(count));
```
lucasw, line 350: Could keep track of the depth values at each corner of the square and interpolate between them

depth_image_proc/src/nodelets/register.cpp
```diff
@@ -195,6 +197,62 @@ void RegisterNodelet::imageCb(const sensor_msgs::ImageConstPtr& depth_image_msg,
...
+  v_rgb = (rgb_fy*xyz_rgb.y() + rgb_Ty)*inv_Z + rgb_cy + 0.5;
+  if (v_rgb < 0 || v_rgb >= height)
+    return false;
+
+  return true;
+}
+
+// TODO(lucasw) need a unit test for this, simple low res image (e.g. 4x4 pixels) with
+// depth ranges of around 1.0
+// shift it to the right 1.0 units, view from 90 degrees
```
lucasw: delete this

depth_image_proc/src/nodelets/register.cpp
```diff
@@ -195,6 +197,62 @@ void RegisterNodelet::imageCb(const sensor_msgs::ImageConstPtr& depth_image_msg,
...
+    T& new_depth)
+{
+  // TODO(lucasw) pulling these out shouldn't cost anything
+  const auto depth_cx = depth_model_.cx(), depth_cy = depth_model_.cy();
+  const auto depth_Tx = depth_model_.Tx(), depth_Ty = depth_model_.Ty();
+  const auto rgb_fx = rgb_model_.fx(), rgb_fy = rgb_model_.fy();
+  const auto rgb_cx = rgb_model_.cx(), rgb_cy = rgb_model_.cy();
+  const auto rgb_Tx = rgb_model_.Tx(), rgb_Ty = rgb_model_.Ty();
+
+  // TODO(lucasw) factor out common code below
```
lucasw: done, delete this

depth_image_proc/src/nodelets/register.cpp
```diff
@@ -174,6 +174,8 @@ void RegisterNodelet::imageCb(const sensor_msgs::ImageConstPtr& depth_image_msg,
   registered_msg->width  = resolution.width;
   // step and data set in convert(), depend on depth data type
 
+  // TODO(lucasw) want a version of this where an rgb image that is lined up
+  // with input depth_image pixel-for-pixel is converted
```
lucasw: Something for another PR, delete this

JWhitleyWork:
@lucasw Once you've implemented your own suggestions (?) I'll give this a look. Please ping me when it's ready.

JWhitleyWork:
@lucasw It looks like you've done a lot of work on this. Is it ready for an external review?

lucasw:
I have some other commits in another branch that are related but those could go in a different PR, but I'll scan through them for anything that ought to be here.  Yes if you can make time try it out and review.

PR #723, opened 2022-01-18
Add support for floating point fps

jeduardo211:
This pull request adds support for floating point fps values for the video_recorder node.


JWhitleyWork:
Seams reasonable. Would you mind forward-porting this to the `ros2` (and potentially `rolling`) branch(es)?

Issue #713, opened 2021-11-25
ros2: variables undefined


azhural:
51a60e723c331507c05e3be37734884cf79ebff9 has introduced a `fisheye_flags` variable in camera_calibration's camera_calibrator.py L108, which is undefined. L153 reads `self.checkerboard_flags`, but should read `self._checkerboard_flags`.

After fixing that locally I ran into a `boards` undefined error. I'm unsure if that's user error.

mikeferguson:
The _checkerboard_flags part is fixed - but not the boards issue

mikeferguson:
Assigning to myself - going to run through the various calibration issues in the next two weeks

mikeferguson:
This appears to be correct in code at this point - can't really determine what the "boards error" was

PR #724, opened 2022-01-21
quick fix to run cameracalibrator #713

jonathanTIE:
Without these name fixes, I can't use cameracalibrator. I don't think it's clean but I don't know enough about the package to make a proper one.  
Issue #713

### DIFF
--- a/camera_calibration/src/camera_calibration/calibrator.py
+++ b/camera_calibration/src/camera_calibration/calibrator.py
@@ -312,7 +312,7 @@ class Calibrator():
         if pattern == Patterns.Chessboard:
             # Make sure n_cols > n_rows to agree with OpenCV CB detector output
             self._boards = [ChessboardInfo("chessboard", max(i.n_cols, i.n_rows), min(i.n_cols, i.n_rows), i.dim) for i in boards]
-        if pattern == Patterns.ChArUco:
+        elif pattern == Patterns.ChArUco:
             self._boards = boards
         elif pattern == Patterns.ACircles:
             # 7x4 and 4x7 are actually different patterns. Assume square-ish pattern, so n_rows > n_cols.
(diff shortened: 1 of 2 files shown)


JWhitleyWork:
@jonathanTIE I get the change to fix the `fisheye_flags` variable, but why the change of the `if`/``else` structure?

marshallpt:
> @jonathanTIE I get the change to fix the `fisheye_flags` variable, but why the change of the `if`/``else` structure?

I was wondering the same thing. When I tried to apply @jonathanTIE 's fix without changing the `if/else` structure, the program still crashed. 

```python
  File ".../dev_ws/install/camera_calibration/lib/python3.8/site-packages/camera_calibration/calibrator.py"", line 402, in get_corners
    for b in self._boards:
AttributeError: 'MonoCalibrator' object has no attribute '_boards'
```
So, that would be why. Their change ensures `self._boards` always gets set by adding an `else` at the end if the `if/elif` branch. Currently, there is no `else`, so if none of the `if`s trigger, `self._boards` never gets set.

However, their fork did not work for me on Foxy. It launches successfully, but it did not detect my chessboard when I held it in front of the camera. I am new to the package, so it's entirely possible I'm doing something wrong. Going off [this tutorial](https://navigation.ros.org/tutorials/docs/camera_calibration.html), it should start detecting it when I hold it in front of the camera, but it does not.

clalancette:
I've retargeted this to the `rolling` branch, and I've also changed the if structure so that it throws an exception when you pass a pattern that isn't supported.  I *think* this is the behavior we want, though I don't have the infrastructure right now to test it.  @jonathanTIE @marshallpt Could you please test out the latest code here and let me know if it works for you?

JWhitleyWork:
@jonathanTIE @marshallpt Are either of you able to test this?

mikeferguson:
Assigning myself this one to review. Planning to do a full round of testing on calibration stuff in the next two weeks.

marshallpt:
hey, sorry I never replied to the comments asking me to test this, unsure how I missed them. Back when I commented on this issue originally I was running Foxy and somehow missed that there is a `foxy` branch. I switched to that branch and no longer had any issues. complete blunder on my part.

mikeferguson:
Tested locally, logic seems sound for the if/else changes

PR #733, opened 2022-02-18
Updates for Ubuntu 22

jspricke:


JWhitleyWork:
AFAIK, ROS Noetic is not and will never be supported on Ubuntu 22.04 so this PR is not valid.

jspricke:
Hi Joshua,

>AFAIK, ROS Noetic is not and will never be supported on Ubuntu 22.04 so this PR is not valid.

While you are free to refuse this PR as a maintainer, I would like to 
give you a different view. First, the PR is valid for Noetic independent 
if it targets Ubuntu 22 or not and the CI is green. I could retitle it 
to "Fix CMake and Boost bind" or something, if that makes a difference 
for you.
Second, The package is part of Ubuntu 22 (with the patches in this PR):

https://launchpad.net/ubuntu/+source/ros-image-pipeline

So there is already ROS 1 for Ubuntu 22 and we agreed with the OSRF 
to provide ROS 1 support through them. In addition there are users 
compiling ROS 1 themself and we collect patches for that here:

https://github.com/ros-o/

We are fine with maintaining the patches there but it would simplify 
things if they would be part of this repo.

Cheers Jochen

Edit: fixed link.

JWhitleyWork:
Fair enough. Thanks for taking the time to explain!

PR #743, opened 2022-04-15
ROS 2 node that flips image and publishes updated transform 

dcconner:
(e.g. for camera mounted upside down)

The node does a simple image rotation  (by 90, 180, 270) as simple pixel transpose .

The node publishes a tf to allow images to be viewed right side up in RViz if camera is mounted upside down. 

This is less general than image_rotate, but performs simpler operations.  

It also preserves the image size and aspect ratio.

It does NOT attempt to remap the distortion parameters at this time.

The code has been tested under Foxy and Galactic.  Foxy required camera topics as parameters due to issue with extracting camera/camera_info names from base topic.  

Posting pull request for review and comment.  It does not currently have any specific unit tests.


dcconner:
Closing and superseding by #756

PR #745, opened 2022-04-26
Fix all of the failing tests

clalancette:
There were quite a few failing tests when running this repository under CI.  The vast majority of the failing tests had to do with include order from cpplint, but there were also some actual failing tests, some flake8 failures, etc.

This very large PR fixes all of that, and does a pass over the dependencies to make sure all dependencies are explicitly stated in the package.xml.  With this in place, all tests on all packages should pass.  Let me know if you'd like me to break this up into smaller PRs for easier review.

Once this is in, I think we'll be in a good place to do a release into Rolling and Humble.

### DIFF
--- a/camera_calibration/src/camera_calibration/calibrator.py
+++ b/camera_calibration/src/camera_calibration/calibrator.py
@@ -41,11 +41,10 @@ import numpy.linalg
 import pickle
 import random
 import sensor_msgs.msg
+import sys
 import tarfile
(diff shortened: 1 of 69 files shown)


JWhitleyWork (approved):
I did a quick once-over and it looks good to me. Additionally, I started on many of these same changes on a branch but didn't finish. Thanks so much for this work!

PR #747, opened 2022-05-13
Add conversion for YUV422-YUY2 image encoding

falckt:
Some webcams use the encoding "YUV422-YUY2". The current code already delegates color space conversion for YUV422 encoded images to cv_bridge. This PR does the same for the -YUY2 variant. CvBridge added support for this in ros-perception/vision_opencv#396.


JWhitleyWork (approved):
LGTM! Thanks!

PR #750, opened 2022-05-24
Updated depth_image_proc for ros2

ksommerkohrt:
Instantiated template for convertDepth, added options to register, and
changed register from a class loader to an RCLPP component.

### DIFF
--- a/depth_image_proc/src/conversions.cpp
+++ b/depth_image_proc/src/conversions.cpp
@@ -97,4 +97,12 @@ void convertRgb(
     }
   }
 }
+
+// force template instantiation
+template void convertDepth<uint16_t>(
+  const sensor_msgs::msg::Image::ConstSharedPtr & depth_msg,
+  sensor_msgs::msg::PointCloud2::SharedPtr & cloud_msg,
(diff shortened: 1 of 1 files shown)


depth_image_proc/src/register.cpp
```diff
@@ -271,7 +271,7 @@ void RegisterNode::convert(
 
 }  // namespace depth_image_proc
 
-#include "class_loader/register_macro.hpp"
+#include "rclcpp_components/register_node_macro.hpp"
 
 // Register the component with class_loader.
-CLASS_LOADER_REGISTER_CLASS(depth_image_proc::RegisterNode, rclcpp::Node)
+RCLCPP_COMPONENTS_REGISTER_NODE(depth_image_proc::RegisterNode)
```
JWhitleyWork: To truly be registered as a component, this also requires registration at the CMake level with the macro `rclcpp_components_register_node(s)`.

JWhitleyWork:
Ping to @ksommerkohrt

JWhitleyWork:
Most of this PR has been implemented in `rolling` already except for the template specialization. Merging.

PR #752, opened 2022-05-31
forward container namespace from stereo_image_proc -> image_proc

ihasdapie:
In the following launch file, we expect that the debayer and rectify nodes are launched, but because of the `<push-ros-namespace>` tag, they are not. This is because the namespace [is not prepended](https://github.com/ros-perception/image_pipeline/blob/c2136cda98768408772ee975d1154af4301401e2/stereo_image_proc/launch/stereo_image_proc.launch.py#L227) to the container name passed to `image_proc.launch.py`. 

Example launchfile:

```
<launch>
  <group>
    <push-ros-namespace namespace="navcam" />
    <include file="$(find-pkg-share stereo_image_proc)/launch/stereo_image_proc.launch.py" />
  </group>
</launch>
```

This PR uses a `PathJoinSubstitution` with some `PythonExpression`s to prepend the namespace to the container name and has been tested [i.e. tried on my system] for empty & nested namespaces as well.

### DIFF
--- a/stereo_image_proc/launch/stereo_image_proc.launch.py
+++ b/stereo_image_proc/launch/stereo_image_proc.launch.py
@@ -40,6 +40,8 @@ from launch.conditions import LaunchConfigurationEquals
 from launch.conditions import LaunchConfigurationNotEquals
 from launch.launch_description_sources import PythonLaunchDescriptionSource
 from launch.substitutions import LaunchConfiguration
+from launch.substitutions import PathJoinSubstitution
+from launch.substitutions import PythonExpression
 from launch_ros.actions import ComposableNodeContainer
 from launch_ros.actions import LoadComposableNodes
 from launch_ros.actions import PushRosNamespace
@@ -215,7 +217,7 @@ def generate_launch_description():
         SetLaunchConfiguration(
             condition=LaunchConfigurationEquals('container', ''),
             name='container',
-            value='stereo_image_proc_container'
+            value='stereo_image_proc_container',
         ),
         GroupAction(
             [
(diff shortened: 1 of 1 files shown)


stereo_image_proc/launch/stereo_image_proc.launch.py
```diff
@@ -224,7 +226,15 @@ def generate_launch_description():
                     PythonLaunchDescriptionSource([
                         FindPackageShare('image_proc'), '/launch/image_proc.launch.py'
                     ]),
-                    launch_arguments={'container': LaunchConfiguration('container')}.items()
+                    launch_arguments={
+                        'container': PathJoinSubstitution([
```
jacobperron: I don't think we should use PathJoinSubstitution since it is intended for joining system paths, not ROS namespaces. E.g. on Windows, I presume it uses a different separator (`\`).

stereo_image_proc/launch/stereo_image_proc.launch.py
```diff
@@ -224,7 +226,15 @@ def generate_launch_description():
...
+                            PythonExpression([
+                                '"/".join("',
+                                LaunchConfiguration('ros_namespace'),
+                                '".split("/")[:-1])'
```
jacobperron: I presume this is to remove the "left_namespace" part? What if we instead prepend the ROS namespace at L220? I think this might make the change a bit simpler.

stereo_image_proc/launch/stereo_image_proc.launch.py
```diff
@@ -215,7 +217,7 @@ def generate_launch_description():
         SetLaunchConfiguration(
             condition=LaunchConfigurationEquals('container', ''),
             name='container',
-            value='stereo_image_proc_container'
+            value='stereo_image_proc_container',
```
jacobperron: Can we just write something like this (I didn't test it myself):
```suggestion
            value=[LaunchConfiguration('ros_namespace'), '/', 'stereo_image_proc_container']
```

stereo_image_proc/launch/stereo_image_proc.launch.py
ihasdapie: Noted. We don't need this anymore if we prepend earlier as you suggested

stereo_image_proc/launch/stereo_image_proc.launch.py
ihasdapie: More or less except we need to wrap it in a `PythonExpression` to deal with the no-namespace case 
Fixed in latest commit https://github.com/ros-perception/image_pipeline/pull/752/commits/59d1677fa83a5fe954fe0a6fae0cad46966870f2

jacobperron (approved):
LGTM, thanks for the patch!

PR #758, opened 2022-07-11
Fix camera_info_url not respected on launch - ROS2

alonfaraj:
Current behavior ignore the `camera_info_url` when pass it as a param from launch file and handle it only if we call it after node already started, during reconfigure.
This PR do the same work happen in the reconfigure callback, during `onInit()`


JWhitleyWork:
@alonfaraj What you're proposing here doesn't make sense. The `camera_info_url` is created as a parameter [in the constructor](https://github.com/ros-perception/image_pipeline/blob/ea216eba6d2c611adbd78bb9ef9db25816d1405c/image_publisher/src/image_publisher.cpp#L57) which happens _before_ `onInit()` and I have not heard other issues with it not being respected. Please open an issue instead. Thanks.

PR #766, opened 2022-07-30
Cropping pointcloud before publishing using min/max XYZ values

arsh09:
A use case would be to crop the pointclouds after certain depth. 

If min/max XYZ are not provided as params, their default values are Inf/-Inf.


JWhitleyWork:
While I understand the need for this, I believe this should be implemented in all `point_cloud_xyz...` nodes and not just `xyzrgb`. Would you mind implementing it in the others as well?

JWhitleyWork:
@arsh09 Friendly ping.

arsh09:
@JWhitleyWork I am sorry for late response. Following your comments, I will add this in all the ```point_cloud_xyz...``` classes

PR #771, opened 2022-08-29
Fix QoS incompatibility camera_calibration ROS2

### LINKED ISSUE #770: camera_calibration: QoS incompatibility ROS2 galactic
When running the camera_calibration node, if the topic of the camera has QoS with reliability BestEffort (which is not crazy at all for image topics), it won't receive anything because of QoS incompatibility. Additionally, I see the following warning detailing that:

```
[WARN][1661803083.618488137][cameracalibrator]: New publisher discovered on topic 'image', offering incompatible QoS. No messages will be received from it. Last incompatible policy: RELIABILITY
```

charlielito:
Currently, if QoS of the camera topic doesn't match, the calibration won't work. For that request the QoS of the topics at runtime to resolve that. Fixes #770

### DIFF
--- a/camera_calibration/src/camera_calibration/camera_calibrator.py
+++ b/camera_calibration/src/camera_calibration/camera_calibrator.py
@@ -48,6 +48,9 @@ from message_filters import ApproximateTimeSynchronizer
 from std_msgs.msg import String
 from std_srvs.srv import Empty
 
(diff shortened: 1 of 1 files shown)


JWhitleyWork (approved):
LGTM! Thanks for the fix!

PR #773, opened 2022-09-09
Feat: Allow to update frame id for rectified images

boitumeloruf:
When rectifying an image by the rectification map provided in the `camera_info`, the image is transformed into a new coordinate system frame which is located at a slightly different position than the coordinate frame of the original image.
In turn, when doing a mutli-sensor fusion, where you want to fuse both the original input image as well as the resulting point cloud of a stereo camera, it is necessary to provide different coordinate frames for the two modalities. Consequently, this also needs to be represented by the `frame_ids` in the headers of the corresponding messages.

Thus, we have **extended the `image_proc/rectify`** nodelet to allow to add a suffix to the `frame_id` of the rectified images. This allows to specify two different coordinate frames for the original and distorted camera image as well the rectified images. 

Furthermore, we have also **extended the `stereo_image_proc/disparity`** nodelet to adopt the `frame_id` from the rectified image for the disparity image (and successively for the point cloud when further using  `stereo_image_proc/point_cloud2`), instead of using it from the RGB image. 

We would be happy if that could be merged into the main repository. Thank you!


image_proc/src/nodelets/rectify.cpp
```diff
@@ -84,6 +85,7 @@ void RectifyNodelet::onInit()
 
   // Read parameters
   private_nh.param("queue_size", queue_size_, 5);
+  private_nh.param("frame_id_suffix", frame_id_suffix_, std::string(""));
```
JWhitleyWork: This name is ambiguous. Could you please instead do something like "rectified_frame_suffix"?

boitumeloruf: Done! I have implemented the change. See [88f848a](https://github.com/ros-perception/image_pipeline/pull/773/commits/88f848a9f33f8776feef01a74c999de78e94c9d8)

JWhitleyWork (approved):
LGTM. Thanks!

PR #783, opened 2022-11-18
Adds scaling option for calibrators

onurtore:
New personal computers can handle the input without the need for downscaling, this commit makes it parametric.

### DIFF
--- a/camera_calibration/nodes/cameracalibrator.py
+++ b/camera_calibration/nodes/cameracalibrator.py
@@ -133,7 +133,8 @@ def main():
                      help="Do not use samples where the calibration pattern is moving faster \
                      than this speed in px/frame. Set to eg. 0.5 for rolling shutter cameras.")
     group.add_option("--scale", type="float", default=1.0,
-                     help="Scaling option for the input image, decrease if your camera ")
+                     help="Scaling option for the input image during pattern \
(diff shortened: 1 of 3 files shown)


onurtore:
@JWhitleyWork,
Friendly ping.

JWhitleyWork:
@onurtore Thanks for the PR. However, this pretty significantly changes the behavior of the node. I'm OK with the idea that you're presenting but not with the change in the default behavior. Instead, can you check for the existence of the parameter and, if it isn't defined, use the current behavior?

onurtore:
Hi @JWhitleyWork ,
You are right, I shouldnt have changed the default behaviour, many people using this node. 

While I fixed that with my latest commit, I belive the current version of the code is a little bit problematic, see the scale parameter is the scale of the input  image relative to 640x480, however this behaviour does not look like intuitive, I believe when someone use such parameter it would expect making scale >1.0 scales up the input image, however the current implementation decreases it, so maybe I should change how the scale param calculated. Anyway, I belive this version is still acceptable if its okay, I can create a seperate PR for what I  mentioned

onurtore:
@JWhitleyWork,
Friendly ping.

onurtore:
@JWhitleyWork,
Friendly ping.

PR #791, opened 2022-12-09
Add option to use the RGB image timestamp for the registered depth image (rebased)

roehling:
For some applications, it is beneficial that the registered depth image
adopts the time stamp from the RGB image, e.g. if one wants to pair them
with ExactTime policy afterwards. It also makes sense from a theoretical
point of view, since the registration aligns the frames both spatially
and temporally.

The new behavior can be enabled on a opt-in basis by setting the
parameter "use_rgb_timestamp".


JWhitleyWork (approved):
LGTM!

PR #792, opened 2022-12-11
fix calibration service topics and cleanup

christianrauch:
This fixes some issues with checking and calling the `set_camera_info` service, so that the `cameracalibrator` node can be used to set the calibration for a camera, e.g. via
```sh
ros2 run camera_calibration cameracalibrator --size 6x8 --square 0.026 --ros-args --remap /camera/set_camera_info:=/set_camera_info --remap /image:=/camera/image_raw
```

### DIFF
--- a/camera_calibration/src/camera_calibration/camera_calibrator.py
+++ b/camera_calibration/src/camera_calibration/camera_calibrator.py
@@ -106,16 +106,19 @@ class CalibrationNode(Node):
         self.set_right_camera_info_service = self.create_client(sensor_msgs.srv.SetCameraInfo, right_camera + "/set_camera_info")
 
         if service_check:
+            available = False
             # assume any non-default service names have been set.  Wait for the service to become ready
             for cli in [self.set_camera_info_service, self.set_left_camera_info_service, self.set_right_camera_info_service]:
                 print("Waiting for service", cli.srv_name, "...")
                 # check all services so they are ready.
-                try:
-                    cli.wait_for_service(timeout_sec=5)
+                if cli.wait_for_service(timeout_sec=1):
+                    available = True
                     print("OK")
-                except Exception as e:
-                    print("Service not found: %s".format(e))
-                    rclpy.shutdown()
+                else:
+                    print(f"Service {cli.srv_name} not found.")
+
+            if not available:
+                raise RuntimeError("no camera service available")
 
         self._boards = boards
         self._calib_flags = flags
--- a/camera_calibration/src/camera_calibration/camera_calibrator.py
+++ b/camera_calibration/src/camera_calibration/camera_calibrator.py
@@ -86,7 +86,7 @@ class ConsumerThread(threading.Thread):
         self.function = function
 
     def run(self):
-        while True:
+        while rclpy.ok():
             m = self.queue.get()
             self.function(m)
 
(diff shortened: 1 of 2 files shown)


camera_calibration/src/camera_calibration/camera_calibrator.py
```diff
@@ -93,28 +93,24 @@ def __init__(self, name, boards, service_check = True, synchronizer = message_fi
...
         super().__init__(name)
 
-        left_camera = self.declare_parameter("left_camera", "left_camera").get_parameter_value().string_value
-        right_camera = self.declare_parameter("right_camera", "right_camera").get_parameter_value().string_value
-        camera = self.declare_parameter("camera", "camera").get_parameter_value().string_value
-
-        self.set_camera_info_service = self.create_client(sensor_msgs.srv.SetCameraInfo, camera + "/set_camera_info")
-        self.set_left_camera_info_service = self.create_client(sensor_msgs.srv.SetCameraInfo, left_camera + "/set_camera_info")
-        self.set_right_camera_info_service = self.create_client(sensor_msgs.srv.SetCameraInfo, right_camera + "/set_camera_info")
+        self.set_camera_info_service = self.create_client(sensor_msgs.srv.SetCameraInfo, "/camera/set_camera_info")
```
JWhitleyWork: Could you please remove the leading `/` here? This causes issues with namespaces.

camera_calibration/src/camera_calibration/camera_calibrator.py
```diff
@@ -93,28 +93,24 @@ def __init__(self, name, boards, service_check = True, synchronizer = message_fi
...
                 print("Waiting for service", cli.srv_name, "...")
                 # check all services so they are ready.
-                try:
-                    cli.wait_for_service(timeout_sec=5)
+                if cli.wait_for_service(timeout_sec=1):
+                    available = True
                     print("OK")
-                except Exception as e:
-                    print("Service not found: %s".format(e))
-                    rclpy.shutdown()
```
JWhitleyWork, line 121: Do we not still need to catch the potential exception here?

JWhitleyWork (changes requested):
There are a lot of changes in this PR not explained by the PR name. Would you mind giving some rationale? I marked a few of them but there are others as well (`call_async` -> `call`, removing `setDaemon`, etc.) that warrant explanation.

camera_calibration/src/camera_calibration/camera_calibrator.py
christianrauch: Which exception could that be? The call only returns a boolean value, therefore the check with if.

camera_calibration/src/camera_calibration/camera_calibrator.py
christianrauch: Good catch. Thanks.

christianrauch:
> There are a lot of changes in this PR not explained by the PR name. Would you mind giving some rationale? I marked a few of them but there are others as well (`call_async` -> `call`, removing `setDaemon`, etc.) that warrant explanation.

I renamed the PR a bit. There are many things wrong with the calibration pipeline. So this PR simply adds commits to fix the pipeline plus some cleanup. The explanations are in the commit messages and I think if you know the API, then the commits are easy to understand.

 E.g. for `call_async` -> `call`, this is simply the wrong usage of the async feature. The waiting happens straight after the request so this is synchronous anyway. But there is no event loop so this would block forever.

JWhitleyWork (approved):
Thanks for working on this!

PR #203, opened 2016-06-19
[image_view] Stamped video output filename for video recorder

wkentaro:
- `_filename:=output.avi _stamped_filename:=false` -> `output.avi`
- `_filename:=_out.avi _stamped_filename:=true` -> `1466299931584632829_out.avi`
- `_filename:=$HOME/.ros/.avi _stamped_filename:=true` -> `/home/ubuntu/.ros/1466299931584632829.avi`

PR #806, opened 2023-04-11
Add option to prepend timestamp to image filename in image_saver node

fdila:
Just a small PR to add the option to optionally prepend the timestamp to the image filename when using image_saver.

I've seen some issues already that have been closed by linking another PR (https://github.com/ros-perception/image_pipeline/pull/203), but to me it looks like that is only for the video recorder.


JWhitleyWork (approved):
LGTM

PR #808, opened 2023-04-19
Fix QoS incompatibility camera_calibration ROS 2 Humble

Wojcik98:
Basically the same as #771 (connected to issue #770):

> Currently, if QoS of the camera topic doesn't match, the calibration won't work. For that request the QoS of the topics at runtime to resolve that.

but this PR is for Humble and includes one minor change (set default QoS history & depth settings).

How to make sure it will be included in future releases? Should I create PR for Rolling or something else?


JWhitleyWork (approved):

JWhitleyWork:
LGTM. Yes, please make a PR to `rolling` to include it in future releases.

PR #813, opened 2023-05-04
feat: image_saver reports an error on file save

russkel:
This makes it more obvious if the `image_saver` is unable to save the file so you know to track down an issue.

Can this be back ported to humble as well please?


JWhitleyWork (approved):
Thanks for the helpful change!

JWhitleyWork:
@russkel Back-porting is not automatic. Could you please submit a separate PR to the `humble` branch?

russkel:
> @russkel Back-porting is not automatic. Could you please submit a separate PR to the `humble` branch?

Done in #819

PR #786, opened 2022-12-05
Backport "Fix ResizeNode to allow remapping both image and camera_info topics #693"

schornakj:
Backport the fix for `image_proc::ResizeNode` from #693 to the `humble` branch.

PR #814, opened 2023-05-12
[Humble] Use the same QoS profiles as publishers in image_proc

### LINKED ISSUE #807: Image debayer node is not compatable with SensorDataQoS in humble
The current default QOS for image debayer node is reliable which is not compatible with sensor data qos while the debayer input is sensor data.

Wojcik98:
Currently nodes in `image_proc` work only with default QoS profiles. This change retrieves QoS profiles from topic's publishers and uses the same ones. This allows working with sensor data QoS which is often used for cameras. It also fixes #807.

If approved, I will create analogous PR for `rolling`.

### DIFF
--- a/image_proc/include/image_proc/utils.hpp
+++ b/image_proc/include/image_proc/utils.hpp
@@ -1,4 +1,4 @@
-// Copyright 2008, 2019 Willow Garage, Inc., Andreas Klintberg, Joshua Whitley
+// Copyright 2023 Willow Garage, Inc., Michal Wojcik
 // All rights reserved.
 //
 // Software License Agreement (BSD License 2.0)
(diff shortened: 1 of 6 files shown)


Wojcik98:
@JWhitleyWork could you please review and merge this PR before #786? I will then prepare new PR that fixes that issue in all nodes in `image_proc`.

zkytony:
I am running into a related issue where RViZ is unable to visualize point cloud published by depth_image_proc due to incompatible QoS profile (between the depth_image_proc publisher and rviz subscriber). I get the following warning message by rviz and nothing is visualized:
```
[WARN] [1684180727.224134262] [rviz]: New publisher discovered on topic '<point_cloud_topic>', 
offering incompatible QoS. No messages will be sent to it. Last incompatible policy: RELIABILITY_QOS_POLICY
```
I believe the QoS depth_image_proc uses is the default SENSOR_DATA configuration, which in Python can be obtained like this
```python
from rclpy.qos import QoSPresetProfiles
QoSPresetProfiles.get_from_short_key("sensor_data")
```
But on the RViZ side, its subscriber seems to be using the default QoS ([not sure if this is the right ref](https://github.com/ros2/rviz/blob/bc93b3488e9b204b1e0e9df2455e1c66c552cb41/rviz_common/include/rviz_common/properties/qos_profile_property.hpp#L58))

Wojcik98:
@zkytony I don't think it's related to this PR, it should be fixed by changing QoS settings in RViz UI. On the list you select desired PointCloud, expand "Topic" subtree and in "Reliability Policy" change "Reliable" to "Best effort". Have you tried that?
image

zkytony:
Thanks! I never knew about this setting. I will give it a try.

Wojcik98:
@JWhitleyWork would you have time to look at this PR? If not is there someone else who could review it?

JWhitleyWork (approved):
Smart. I'll have to use this elsewhere too.

hemalshahNV:
The change in QoS profiles to Best Effort reliability has increased frame drops significantly in our nightly ros2_benchmark benchmarking with image_proc killing over 53/221 workflows. The update itself seems to be incorrect too since now we always get the default sensor QoS rather than forwarding QoS from publisher as intended. Is anyone else seeing issues?

PR #830, opened 2023-07-21
added more aruco dicts from opencv

J-C-Haley:
Expanded the number of patterns available


mikeferguson:
There are no changes in this PR - presumably the changes ended up being put in via a different PR

Issue #833, opened 2023-08-23
NameError: name 'CAMERA_MODEL' is not defined


haormj:
- branch: foxy
- ros2 run camera_calibration cameracalibrator --size 6x8 --square 0.02 --ros-args -r image:=/front_camera/image_raw -p camera:=/front_camera
- camera type fisheye
- then prompt "NameError: name 'CAMERA_MODEL' is not defined"

mikeferguson:
This is fixed in humble and later, so all current ROS 2 releases

PR #834, opened 2023-08-23
fix(foxy): name CAMERA_MODEL is not defined in foxy branch

haormj:
Fixes https://github.com/ros-perception/image_pipeline/issues/833


ahcorde:
This PR is open against `foxy` which is already [end of life](https://docs.ros.org/en/foxy/Releases/End-of-Life.html).  Closing this 

Feel free to open this PR against `rolling` if the problem persists, then we can backport the change to `iron` and/or `humble`.

PR #838, opened 2023-10-22
ros2 fixed async request

dariodenardi:
In the Humble version, there is a bug: when you click 'commit' after calibration, an error appears:

```
!!!!!!!!!!!!!!!!!!!!
Not available
!!!!!!!!!!!!!!!!!!!!
```

Upon inspecting the code, the issue seems to be related to asynchronous requests:

```
response = self.set_left_camera_info_service.call_async(req)
rv = rv and self.check_set_camera_info(response)
```

The request is sent, but the second function checks the response before it has actually arrived. I believe the solution is to send both asynchronous requests and then wait for the responses.

### DIFF
--- a/camera_calibration/src/camera_calibration/camera_calibrator.py
+++ b/camera_calibration/src/camera_calibration/camera_calibrator.py
@@ -48,6 +48,7 @@ except ImportError:
 from camera_calibration.calibrator import CAMERA_MODEL
 from rclpy.qos import qos_profile_system_default
 from rclpy.qos import QoSProfile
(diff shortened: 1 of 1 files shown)


mikeferguson:
Assigning myself to test this out - going to do a pass through the calibration tickets in the next two weeks

mikeferguson:
It looks like https://github.com/ros-perception/image_pipeline/pull/792 actually fixed this in a different way (by switching away from async entirely). I think we should instead port that back to humble (it's already in rolling and iron)

PR #843, opened 2023-12-30
[camera_calibration] fix typo self.checkerboard_flags to self._checkerboard_flags

HiroIshida:
This PR fixes an obvious bug.

The attributename is  `_cehckerboard_flag`.
https://github.com/ros-perception/image_pipeline/blob/948847514311772cf1d0fd84ef8bebbac94c940f/camera_calibration/src/camera_calibration/camera_calibrator.py#L131

So, `self.checkerboard_flags` is a typo


mikeferguson (approved):
I'm not actually trying to maintain the ROS 1 version - but this one is so clearly correct

PR #844, opened 2024-01-10
Depth image transport configure susbcribers

ahcorde:
Subscribers are setted to SystemDefault QoS sometimes this is not the right QoS of the publisher. 

Adding the option to modify the QoS parameters.


adityapande-1995 (approved):
Could we add a small test here ? Otherwise LGTM

PR #822, opened 2023-06-06
fix: change type of epsilon

wep21:
This PR fixes type of epsilon to prevent the epsilon from always being set to 1 second.

PR #849, opened 2024-01-17
fix: change type for epsilon (backport #822)

mikeferguson:


mikeferguson:
This bug is also present in iron (feature didn't exist in humble) - so we might as well fix it there

PR #815, opened 2023-05-17
add support for lazy subscribers

mikeferguson:
This implements #780 for ROS 2 distributions after Iron, where we have:

 * Connect/disconnect callbacks, per https://github.com/ros2/rmw/issues/330 (this made it into Iron)
 * Updated APIs in https://github.com/ros-perception/image_common/pull/272 (this is only in Rolling currently)

PR #853, opened 2024-01-18
make remaining components lazy

mikeferguson:
missed a few components in #815


mikeferguson:
tested locally by starting each component, then ros2 topic echo the output topic and see that it subscribes to appropriate topics

ahcorde (approved):

PR #862, opened 2024-01-19
image_rotate: clean up

### LINKED ISSUE #740: ros2 version: image_rotate does not initialize the target and source _vectors in constructor
Need to add the following after onInit() call to update data; target_vector and source_vector default to zero

` 
  target_vector_.vector.x = config_.target_x;
  target_vector_.vector.y = config_.target_y;
  target_vector_.vector.z = config_.target_z;

  source_vector_.vector.x = config_.source_x;
  source_vector_.vector.y = config_.source_y;
  source_vector_.vector.z = config_.source_z;

`

mikeferguson:
This is the first component/node with a cleanup pass to be fully implemented:

 * Fix #740 by initializing vectors. Do this by declaring parameters AFTER we define the callback
 * Implement lazy subscribers (I missed this in the earlier PRs)
 * Add image_transport parameter so we can specify that desired transport of our subscriptions
 * Update how we test for connectivity (and update the debug message - has nothing to do with whether we are remapped, it's really about whether we are connected)

### DIFF
--- a/image_rotate/src/image_rotate.cpp
+++ b/image_rotate/src/image_rotate.cpp
@@ -38,13 +38,6 @@ int main(int argc, char ** argv)
   // Force flush of the stdout buffer.
   setvbuf(stdout, NULL, _IONBF, BUFSIZ);
 
-  if (argc <= 1) {
-    RCUTILS_LOG_WARN(
-      "Topic 'image' has not been remapped! Typical command-line usage:\n"
-      "\t$ ros2 run image_rotate image_rotate image:=<image topic>");
-    return 1;
-  }
-
   rclcpp::init(argc, argv);
   rclcpp::NodeOptions options;
   auto node = std::make_shared<image_rotate::ImageRotateNode>(options);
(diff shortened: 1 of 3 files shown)


image_rotate/src/image_rotate.cpp
```diff
@@ -38,13 +38,6 @@ int main(int argc, char ** argv)
   // Force flush of the stdout buffer.
   setvbuf(stdout, NULL, _IONBF, BUFSIZ);
 
-  if (argc <= 1) {
-    RCUTILS_LOG_WARN(
-      "Topic 'image' has not been remapped! Typical command-line usage:\n"
-      "\t$ ros2 run image_rotate image_rotate image:=<image topic>");
-    return 1;
-  }
```
mikeferguson, line 46: This debug message didn't work at all - made a more robust version by checking the overlap of our topic name with the actual topics that exist (as many of our depth_image_proc stuff does)

image_rotate/src/image_rotate_node.cpp
```diff
@@ -128,13 +106,31 @@ ImageRotateNode::ImageRotateNode(const rclcpp::NodeOptions & options)
       source_vector_.vector.x = config_.source_x;
       source_vector_.vector.y = config_.source_y;
       source_vector_.vector.z = config_.source_z;
-      if (subscriber_count_) {  // @todo: Could do this without an interruption at some point.
-        unsubscribe();
-        subscribe();
-      }
```
mikeferguson, line 134: this is unneeded - none of the parameters that are able to be updated would impact how the subscribers would actually be setup

ahcorde (approved):

PR #866, opened 2024-01-19
Add support for floating point fps

ahcorde:
Related with this PR https://github.com/ros-perception/image_pipeline/pull/723


mikeferguson (approved):

mikeferguson:
This one should probably get back ported to humble/iron since it's a fairly egregious bug

PR #869, opened 2024-01-19
support rgba8 and bgra8 encodings by skipping alpha channel

ahcorde:
Related with the change in ROS 1 https://github.com/ros-perception/image_pipeline/pull/671/files


mikeferguson (changes requested):
It looks like this is only part of the changes in the ROS 1 linked PR? There were 3 other files that got this support - should we add those as well?

mikeferguson (approved):

PR #876, opened 2024-01-19
image_view: consistent image_transport

mikeferguson:
All of these nodes already have the proper remapping support - but image_transport parameter support was scattered

### DIFF
--- a/image_view/src/extract_images_node.cpp
+++ b/image_view/src/extract_images_node.cpp
@@ -75,13 +75,15 @@ ExtractImagesNode::ExtractImagesNode(const rclcpp::NodeOptions & options)
   auto node_base = this->get_node_base_interface();
   std::string topic = node_base->resolve_topic_or_service_name("image", false);
 
-  this->declare_parameter<std::string>("transport", std::string("raw"));
+  // TransportHints does not actually declare the parameter
+  this->declare_parameter<std::string>("image_transport", "raw");
+  image_transport::TransportHints hints(this);
   std::string transport = this->get_parameter("transport").as_string();
 
   sub_ = image_transport::create_subscription(
     this, topic, std::bind(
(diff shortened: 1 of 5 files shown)


image_view/src/stereo_view_node.cpp
```diff
@@ -97,8 +97,9 @@ StereoViewNode::StereoViewNode(const rclcpp::NodeOptions & options)
   std::string format_string = this->get_parameter("filename_format").as_string();
   filename_format_.parse(format_string);
 
-  this->declare_parameter<std::string>("transport", std::string("raw"));
-  std::string transport = this->get_parameter("transport").as_string();
+  // TransportHints does not actually declare the parameter
+  this->declare_parameter<std::string>("image_transport", std::string("raw"));
```
mikeferguson, line 101: NOTE: if we back port this - we probably need to revert the change in name of this parameter

image_view/src/extract_images_node.cpp
```diff
@@ -75,13 +75,15 @@ ExtractImagesNode::ExtractImagesNode(const rclcpp::NodeOptions & options)
   auto node_base = this->get_node_base_interface();
   std::string topic = node_base->resolve_topic_or_service_name("image", false);
 
-  this->declare_parameter<std::string>("transport", std::string("raw"));
+  // TransportHints does not actually declare the parameter
+  this->declare_parameter<std::string>("image_transport", "raw");
```
mikeferguson, line 79: Same here - renamed parameter for consistency with image_transport specification - I've noted both instances of this in my Jazzy: Documentation ticket so we can mention the changed parameter name

ahcorde (changes requested):
linters
```bash
- image_view.cpplint whitespace/line_length [2] (/__w/image_pipeline/image_pipeline/src/image_view/src/video_recorder_node.cpp:79)
```

ahcorde (approved):

PR #883, opened 2024-01-19
resize/recify: consistent image_transport

mikeferguson:
* support image_transport parameter
 * proper renaming so compressed/etc topics work as expected

Additional minor fixes:
 * rename interpolation -> interpolation_ for consistency
 * move parameter declaration BEFORE we create a publisher (and possibly get a subscriber created in connect callback)
 * put the getTopicQosProfile() for publisher right in front of publisher declaration for clarity

### DIFF
--- a/image_proc/src/rectify.cpp
+++ b/image_proc/src/rectify.cpp
@@ -32,6 +32,7 @@
 
 #include <functional>
 #include <mutex>
+#include <string>
 
 #include "cv_bridge/cv_bridge.hpp"
 #include "tracetools_image_pipeline/tracetools.h"
--- a/image_proc/src/rectify.cpp
+++ b/image_proc/src/rectify.cpp
@@ -51,9 +51,16 @@ namespace image_proc
 RectifyNode::RectifyNode(const rclcpp::NodeOptions & options)
 : rclcpp::Node("RectifyNode", options)
 {
-  auto qos_profile = getTopicQosProfile(this, "image");
+  // TransportHints does not actually declare the parameter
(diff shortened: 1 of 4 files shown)


image_proc/src/rectify.cpp
```diff
@@ -51,9 +51,16 @@ namespace image_proc
 RectifyNode::RectifyNode(const rclcpp::NodeOptions & options)
 : rclcpp::Node("RectifyNode", options)
 {
-  auto qos_profile = getTopicQosProfile(this, "image");
+  // TransportHints does not actually declare the parameter
+  this->declare_parameter<std::string>("image_transport", "raw");
```
ahcorde, line 56: include header `<string>`

image_proc/src/resize.cpp
```diff
@@ -51,7 +51,22 @@ namespace image_proc
 ResizeNode::ResizeNode(const rclcpp::NodeOptions & options)
 : rclcpp::Node("ResizeNode", options)
 {
-  auto qos_profile = getTopicQosProfile(this, "image/image_raw");
+  // TransportHints does not actually declare the parameter
+  this->declare_parameter<std::string>("image_transport", "raw");
```
ahcorde, line 56: include header `<string>`

ahcorde (changes requested):

image_proc/src/rectify.cpp
mikeferguson: string is already in the hpp for this file - hence why the linter isn't asking for it to be added to the cpp file

ahcorde (approved):

PR #884, opened 2024-01-19
image_proc: consistent image_transport

mikeferguson:
* consistent image_transport parameter for crop_decimate, crop_non_zero and debayer nodes
* consistent remapping support for compressed/etc topics in all three nodes
* add lazy subscription support to crop_non_zero

Additional minor fixes:

 * put the getTopicQosProfile() for publisher right in front of publisher declaration for clarity

### DIFF
--- a/image_proc/src/crop_non_zero.cpp
+++ b/image_proc/src/crop_non_zero.cpp
@@ -52,14 +52,34 @@ namespace image_proc
 CropNonZeroNode::CropNonZeroNode(const rclcpp::NodeOptions & options)
 : Node("CropNonZeroNode", options)
 {
+  // TransportHints does not actually declare the parameter
+  this->declare_parameter<std::string>("image_transport", "raw");
+
+  // For compressed topics to remap appropriately, we need to pass a
+  // fully expanded and remapped topic name to image_transport
+  auto node_base = this->get_node_base_interface();
+  image_topic_ = node_base->resolve_topic_or_service_name("image_raw", false);
+
+  // Create image pub with connection callback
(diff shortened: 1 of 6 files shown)


image_proc/src/crop_non_zero.cpp
```diff
@@ -52,14 +52,34 @@ namespace image_proc
 CropNonZeroNode::CropNonZeroNode(const rclcpp::NodeOptions & options)
 : Node("CropNonZeroNode", options)
 {
+  // TransportHints does not actually declare the parameter
+  this->declare_parameter<std::string>("image_transport", "raw");
```
ahcorde, line 57: include header `<string>`

image_proc/src/debayer.cpp
```diff
@@ -52,7 +52,14 @@ namespace image_proc
 DebayerNode::DebayerNode(const rclcpp::NodeOptions & options)
 : Node("DebayerNode", options)
 {
-  auto qos_profile = getTopicQosProfile(this, "image_raw");
+  // TransportHints does not actually declare the parameter
+  this->declare_parameter<std::string>("image_transport", "raw");
```
ahcorde, line 57: include header `<string>`

ahcorde (changes requested):

image_proc/src/crop_non_zero.cpp
mikeferguson: string is already in the header file for this class - hence why the linter is asking for it to be added in this file

ahcorde (approved):

PR #900, opened 2024-01-19
depth_image_proc: consistent image_transport

mikeferguson:


mikeferguson:
This still needs testing around the camera_info topics remapping properly - just opening for visibility

mikeferguson:
Changes to document:

 * xyz_radial, xyzi_radial - renamed depth/image_raw to depth/image_rect for consistency with every other node
 * all of the "point cloud node" depth input topics use depth_image_transport - even if there isn't an rgb or intensity image being subscribed to. This is a change from ROS 1, but image_transport wasn't even supported in this package previously in ROS 2, so not a change that will impact people.

Testing notes:

 * As with stereo_image_proc - you can remap the camera_info topics to override the name, but you have to use the name of the topic after the camera topic itself is remapped
 * every node was manually tested to be sure that our topics remap as expected (camera_info follows camera topic rules, compressed gets remapped as expected)
 * Convert metric and crop foremost use image_transport for output, so update those to support remapping properly

ahcorde (approved):

PR #903, opened 2024-01-19
stereo_image_proc: consistent image_transport

mikeferguson:
* make image_transport work
* make remap work as expected with image_transport
* make subscribers lazy

### DIFF
--- a/stereo_image_proc/src/stereo_image_proc/disparity_node.cpp
+++ b/stereo_image_proc/src/stereo_image_proc/disparity_node.cpp
@@ -48,6 +48,7 @@
 
 #include <stereo_image_proc/stereo_processor.hpp>
 
+#include <image_transport/camera_common.hpp>
 #include <image_transport/image_transport.hpp>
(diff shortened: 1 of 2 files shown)


mikeferguson:
Testing notes:

 * Can remap output properly (```--ros-args -r points2:=test```)
 * Can remap individual camera topics:
   * ```-r left/image_raw:=test/image_raw``` remaps both left/image_raw and left/camera_info as expected
   * When also adding ```-p image_transport:=compressed``` we get the expected test/image_raw/compressed and test/camera_info
 * Can remap camera_info topic - but it's slightly more complex, since you need to remap the remapped topic (I see no way to cleanly remap "left/camera_info" while allowing camera_info to automatically remap properly for when "left/image_raw" has been changed. This isn't a common use case, and is kind of outside the expected standard, so we will simply document it)
 * Cannot remap just "left" and "right" - has to be "left/image_raw" - this is a bit different from ROS 1, but I think it is just how ROS 2 does lower level name spacing

ahcorde (approved):

PR #905, opened 2024-01-20
depth_image_proc: update launch files

mikeferguson:
* follow up to #900 - had not noticed these launch files at the time
* remove camera_info topics that auto remap now


ahcorde (approved):

mikeferguson:
This does not need a back port - we're only changing these topics in Jazzy and later (to avoid breaking people)

Issue #388, opened 2019-03-04
radial point cloud nodes assume unrectified images


madsherlock:
The depth_image_proc/point_cloud_xyz_radial and depth_image_proc/point_cloud_xyz_radial nodes apply a projective transformation to depth images to produce point clouds.

This projective transformation is calculated from the **camera matrix** K from the camera_info topic. In my opinion, this transformation should be calculated from the **projection matrix** P from the camera_info topic, if available, as well as from the rectification rotation matrix R. These can be provided to the cv::undistortPoints() function from within the initMatrix() method of the nodelet.

It would allow the camera_info topic to be fully utilized; I need a translation contained within the projection matrix to be applied to the point cloud before it is in the right frame.

Is the current behavior (only using K) intended or a bug?

mikeferguson:
I started to look into this, and it turned into a bit of a rabbit hole:

For starters - let's review which nodes use what (I'm only looking at the rolling branch for now, because any changes are likely to only go there):
 * xyz, xyzi, xyzrgb use image_geometry::PinholeCameraModel - which means they actually use the P matrix
 * xyz_radial, xyzi_radial, xyzrgb_radial use a transform matrix derived from K and D.

Per the [sensor_msgs/CameraInfo](https://github.com/ros2/common_interfaces/blob/rolling/sensor_msgs/msg/CameraInfo.msg) message:
 * K is the "Intrinsic camera matrix for the raw (distorted) images"
 * P "specifies the intrinsic (camera) matrix of the processed (rectified) image"
 * R is the "Rectification matrix (stereo cameras only)" and is "A rotation matrix aligning the camera coordinate system to the ideal stereo image plane so that epipolar lines in both stereo images are parallel."

There is quite a bit of good discussion on this [answers.ros.org](https://answers.ros.org/question/119506/what-does-projection-matrix-provided-by-the-calibration-represent/) - but basically, it agrees with CameraInfo:
 * If you are subscribing to image_raw, use K and D
 * If you are subscribing to image_rect, use P and set D to the identity matrix.

[...]

PR #906, opened 2024-01-21
radial nodes: should all sub to raw topics

mikeferguson:
Per findings in https://github.com/ros-perception/image_pipeline/issues/388#issuecomment-1902487162 - instead of renaming xyz_radial and xyzi_radial to image_rect, I should have made the xyzrgb_radial use image_raw (since these nodes use matrices K & D):

 * Revert the change in xyzi_radial - topic is depth/image_raw as it has always been
 * Revert the change in xyz_radial, although it is still changed slightly from the old "image_raw" -> "depth/image_raw" for consistency with the other nodes.
 * Update xyzrgb_radial:
   * depth_registered/image_rect -> depth/image_raw
   * rgb/image_rect_color -> rgb/image_raw
 * update launch files accordingly (and remove camera_info since it no longer needs to be renamed, happens automagically). Note: these launch files are probably epically bad since realsense doesn't output radial images... but we'll leave them as documentation for these nodes.

### DIFF
--- a/depth_image_proc/include/depth_image_proc/conversions.hpp
+++ b/depth_image_proc/include/depth_image_proc/conversions.hpp
@@ -60,7 +60,7 @@ void convertDepth(
   float center_y = model.cy();
 
   // Combine unit conversion (if necessary) with scaling by focal length for computing (X,Y)
-  double unit_scaling = DepthTraits<T>::toMeters(T(1) );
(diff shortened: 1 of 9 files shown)


ahcorde (approved):

mikeferguson:
This does not need a back port - we're only changing these topics in Jazzy and later (to avoid breaking people)

PR #920, opened 2024-01-31
Update extract_images_sync to add sec_per_frame parameter

### LINKED ISSUE #726: missing "sec_per_frame" param in "extract_images_sync" node
a "sec_per_frame" ros  parameter as the one in the c++ node "extract_images" is missing not allowing to decimate the frames extracted from topics.

artineering:
Added sec_per_frame parameter to allow decimation of frames being synchronized and captured.
fixes #726

### DIFF
--- a/image_view/scripts/extract_images_sync
+++ b/image_view/scripts/extract_images_sync
@@ -51,10 +51,19 @@ class ExtractImagesSync(Node):
         super().__init__('extract_images_sync')
         self.get_logger().info('Extract_Images_Sync Node has been started')
         self.seq = 0
+        self.time = self.get_clock().now()
+
         self.fname_fmt = self.declare_parameter(
-            'filename_format', 'frame%04i_%i.jpg').value   
+            'filename_format', 'frame%04i_%i.jpg').value
+
+        # Do not scale dynamically by default.
         self.do_dynamic_scaling = self.declare_parameter(
             'do_dynamic_scaling', False).value
+
+        # Limit the throughput to 10fps by default.
+        self.sec_per_frame = self.declare_parameter(
+            'sec_per_frame', 0.1).value
+
         img_topics = self.declare_parameter('inputs', None).value
 
         if img_topics is None:
@@ -80,6 +89,17 @@ Typical command-line usage:
         sync.registerCallback(self.save)
 
     def save(self, *imgmsgs):
+        delay = self.get_clock().now() - self.time
+
+        # Decimation is enabled if it is set to greater than
+        # zero. If it is enabled, exit if delay is less than
+        # the decimation rate specified.
+        if (self.sec_per_frame > 0) and (delay < self.sec_per_frame):
+            return
+        
+        # Update the current time
+        self.time = self.get_clock().now()
+
         seq = self.seq
         bridge = cv_bridge.CvBridge()
         for i, imgmsg in enumerate(imgmsgs):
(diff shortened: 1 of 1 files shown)


mikeferguson:
Running this locally I hit: 
```
  File "/home/fergs/image/install/image_view/lib/image_view/extract_images_sync", line 97, in save
    if (self.sec_per_frame > 0) and (delay < self.sec_per_frame):
TypeError: '<' not supported between instances of 'Duration' and 'float'
```

image_view/scripts/extract_images_sync
```diff
@@ -51,10 +51,19 @@ class ExtractImagesSync(Node):
...
-            'filename_format', 'frame%04i_%i.jpg').value   
+            'filename_format', 'frame%04i_%i.jpg').value
+
+        # Do not scale dynamically by default.
         self.do_dynamic_scaling = self.declare_parameter(
             'do_dynamic_scaling', False).value
+
+        # Limit the throughput to 10fps by default.
+        self.sec_per_frame = self.declare_parameter(
+            'sec_per_frame', 0.1).value
```
mikeferguson: we should probably not change default behavior - so this should probably be set to 0 by default

image_view/scripts/extract_images_sync
```diff
@@ -51,10 +51,19 @@ class ExtractImagesSync(Node):
         super().__init__('extract_images_sync')
         self.get_logger().info('Extract_Images_Sync Node has been started')
         self.seq = 0
+        self.time = self.get_clock().now()
```
mikeferguson: let's use a more descriptive name than "time" - something like "last_frame" or "last_frame_time", etc

image_view/scripts/extract_images_sync
```diff
@@ -80,6 +89,17 @@ Typical command-line usage:
         sync.registerCallback(self.save)
 
     def save(self, *imgmsgs):
+        delay = self.get_clock().now() - self.time
+
+        # Decimation is enabled if it is set to greater than
+        # zero. If it is enabled, exit if delay is less than
+        # the decimation rate specified.
+        if (self.sec_per_frame > 0) and (delay < self.sec_per_frame):
```
mikeferguson: you can't directly compare a Duration and a float - need to do a conversion here:
```
  File "/home/fergs/image/install/image_view/lib/image_view/extract_images_sync", line 97, in save
    if (self.sec_per_frame > 0) and (delay < self.sec_per_frame):
TypeError: '<' not supported between instances of 'Duration' and 'float'
```

mikeferguson (changes requested):

image_view/scripts/extract_images_sync
```diff
@@ -51,10 +51,19 @@ class ExtractImagesSync(Node):
...
-            'filename_format', 'frame%04i_%i.jpg').value   
+            'filename_format', 'frame%04i_%i.jpg').value
+
+        # Do not scale dynamically by default.
         self.do_dynamic_scaling = self.declare_parameter(
             'do_dynamic_scaling', False).value
+
+        # Limit the throughput to 10fps by default.
+        self.sec_per_frame = self.declare_parameter(
+            'sec_per_frame', 0).value
```
mikeferguson: default needs to be 0.0 so that the parameter is a float rather than an integer

mikeferguson (changes requested):

mikeferguson:
I just pushed a commit that:

 * makes the param a float, rather than an integer
 * does proper conversion (there is no to_sec(), but we can make a Duration from secs_per_frame)

Tested locally and works

mikeferguson (approved):

Issue #847, opened 2024-01-17
Jazzy: Follow REP-2003 and allow overrides


mikeferguson:
* All publishers should be SystemDefaultQOS with allowable overrides
 * All subscribers should be SensorDataQOS with allowable overrides

Notes on overrides: https://docs.ros.org/en/iron/How-To-Guides/Overriding-QoS-Policies-For-Recording-And-Playback.html

From point_cloud_node:
```
auto sub_opts = rclcpp::SubscriptionOptions();
sub_opts.qos_overriding_options = rclcpp::QosOverridingOptions::with_default_policies();
```

- [x] image_save should subscribe with SensorDataQOS (https://github.com/ros-perception/image_pipeline/issues/828)
- [x] image_proc
- [x] stereo_image_proc
- [x] depth_image_proc
- [x] image_view
- [x] Add tutorial to image_pipeline tutorials on how to use QoS (link to design doc above also)

tcappellari-bdai:
Is this still being worked on? We are having issues with these nodes as the QoS for subscribers are not sensor defaults

mikeferguson:
I finally got around to finishing this work - PR is here: https://github.com/ros-perception/image_pipeline/pull/1019

There are a few nodes/components that are not updated due to API limitations in message_filters and/or image_transport - I'll file a separate ticket with those nodes listed and what upstream API stuff we should be pushing for K-turtle

mikeferguson:
Merged into both Jazzy and Rolling - I'll cut a new release shortly

PR #922, opened 2024-02-01
QoS improvements for image_proc and stereo_image_proc

mikeferguson:
First part of #847 
* Add QoS overrides for all publishers (in the new, standard way)
* stereo_image_proc: Default subscriber QoS to SensorDataQoS
* Clean up some of the comments around lazy subscribers, make them more consistent across nodes

### DIFF
--- a/image_proc/src/crop_non_zero.cpp
+++ b/image_proc/src/crop_non_zero.cpp
@@ -61,7 +61,7 @@ CropNonZeroNode::CropNonZeroNode(const rclcpp::NodeOptions & options)
   auto node_base = this->get_node_base_interface();
   image_topic_ = node_base->resolve_topic_or_service_name("image_raw", false);
 
-  // Create image pub with connection callback
+  // Setup lazy subscriber using publisher connection callback
   rclcpp::PublisherOptions pub_options;
   pub_options.event_callbacks.matched_callback =
     [this](rclcpp::MatchedInfo &)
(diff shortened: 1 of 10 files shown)


mikeferguson:
A few notes:

 * The default QoS for the publishers has actually always been a bit sketchy and nondeterministic - since the publisher we are subscribing to may or may not actually exist yet when the QoS is probed. In those cases, we get a default of SensorDataQoS (which, is maybe not what we want, if we want to follow REP-2003 we would actually want SystemDefaultsQoS - although we may be stretching REP-2003 a bit here). The situation is a bit better for subscribers now that we lazily subscribe to them, which means their creation is delayed.

I almost think we should just remove the automatic publisher QoS and set it to SystemDefaultQoS + Overrides (which is at least deterministic).

image_proc/src/crop_non_zero.cpp
```diff
@@ -78,8 +79,11 @@ CropNonZeroNode::CropNonZeroNode(const rclcpp::NodeOptions & options)
       }
     };
 
-  // Create publisher with same QoS as subscribed topic
-  auto qos_profile = getTopicQosProfile(this, "image_raw");
+  // Allow overriding QoS settings (history, depth, reliability)
+  pub_options.qos_overriding_options = rclcpp::QosOverridingOptions::with_default_policies();
+
+  // Create publisher with QoS matched to subscribed topic publisher
+  auto qos_profile = getTopicQosProfile(this, image_topic_);
```
mikeferguson, line 86: only change here is actually passing in image_topic_ rather than "image_raw" (technically still worked before, but leaves open the possibility of the topic names getting out of sync)

stereo_image_proc/launch/stereo_image_proc.launch.py
```diff
@@ -94,6 +92,10 @@ def generate_launch_description():
                     'left/image_rect_color',
                     [LaunchConfiguration('left_namespace'), '/image_rect_color']
                 ),
+                (
+                    'right/image_rect_color',
+                    [LaunchConfiguration('right_namespace'), '/image_rect_color']
+                ),
```
mikeferguson, line 98: this was a relevant patch from #720 (which was an alternative approach, before the overriding was really available)

ahcorde (approved):

PR #925, opened 2024-02-04
migrate and update image_proc docs

mikeferguson:
* move component documentation from ros wiki, update for various changes
 * add tutorial on how to run components
 * update tutorial on debayer/rectify to use launch file
 * remove image_proc node, it has always been completely broken and the launch file has the same (but working) functionality
 * update launch file to support namespace parameter for tutorial

PR #927, opened 2024-02-05
fixup bash command rendering

mikeferguson:
I didn't actually rebuild with the suggestions in #925 - but this is actually proper rendering (even my three-ticks version wasn't quite pretty)


mikeferguson:
Just FYI @ahcorde - I wasn't aware that the ```bash syntax would't work with rosdoc2/sphinx

PR #928, opened 2024-02-05
migrate stereo_image_proc docs

mikeferguson:


mikeferguson:
I did not end up porting the last tutorial - because it requires a BUNCH of updates to actually work - that is now ticketed here: https://github.com/ros-perception/image_pipeline/issues/932 and noted in the docs that the wiki tutorial exists but isn't up to date.

stereo_image_proc/doc/components.rst
```diff
@@ -0,0 +1,123 @@
...
+   synchronization. Set to true if the left and right cameras do not produce
+   exactly synced timestamps.
+ * **image_transport** (string, default: raw): Image transport to use for left
+   image subscriber.
+ * **queue size** (int, default: 5): Size of message queue for each synchronized
+   topic. You may need to raise this if images take significantly longer to travel
+   over the network than camera info and/or the delay from disparity processing
+   is too long.
+
+approximate_sync_tolerance_seconds
```
ahcorde: remove ?

ahcorde (approved):
just a minor comment

stereo_image_proc/doc/components.rst
mikeferguson: I'll look into it - not sure if I meant to document a parameter that wasn't previously documented, or I just left something around - as it turns out, staring at all these pages that are just slightly different ends up blurring stuff together...

PR #929, opened 2024-02-05
migrate image_pipeline docs

mikeferguson:
* Migrates image_pipeline overview page
 * Migrates CameraInfo wiki page
 * Adds links to the other packages in this stack
 * Updates depth_image_proc and image_proc to have the overview page properly named and in the TOC


image_pipeline/doc/index.rst
```diff
@@ -0,0 +1,43 @@
...
+ * **Visualization**: The ``image_view`` package provides a lightweight
+   alternative to ``rviz2`` for viewing an image topic. It also includes
+   a ``stereo_view`` tool for viewing stereo pairs and disparity images.
+
+.. toctree::
+   :maxdepth: 2
+
+   self
+   camera_info
+   camera_calibration <../camera_calibration/index.html#http://>
```
mikeferguson, line 33: this is a little hacky, but according to stack overflow - it's literally the only way to do a relative link in a TOC - if you don't have http:// in there somewhere, it tries to treat it as a filename, which obviously doesn't exist at build time.

ahcorde (approved):

PR #934, opened 2024-02-08
migrate image_view docs

mikeferguson:
* migrate docs from ROS wiki
 * FIX: video recorder start/end services should only be advertised when feature is enabled
 * FIX: image remapping didn't work as expected/documented in stereo_image_proc


mikeferguson:
In porting the docs (and finding that many parameters weren't really documented), I came to this conclusion as well: https://github.com/ros-perception/image_pipeline/issues/935

ahcorde (approved):

Issue #940, opened 2024-02-11
image_transport publishers don't remap correctly


mikeferguson:
The subscribers have all been fixed - but anywhere we use image_transport for a publisher, we also need to do the fully resolved name trick or we get things like:

```
    /stereo/left/image_rect/compressed: sensor_msgs/msg/CompressedImage
    /stereo/left/image_rect/compressedDepth: sensor_msgs/msg/CompressedImage
    /stereo/left/image_rect/theora: theora_image_transport/msg/Packet
    /stereo/left/image_rect_color: sensor_msgs/msg/Image
```

It appears this possibly affects (based on grep for image_transport functions):

 * [x] depth_image_proc/register
 * [x] depth_image_proc/crop_foremost (already does full remapping)
 * [x] depth_image_proc/convert_metric (already does full remapping)
 * [x] image_proc/crop_decimate
 * [x] image_proc/crop_non_zero
 * [x] image_proc/debayer
 * [x] image_proc/resize
 * [x] image_publisher/image_publisher (already does full remapping)
 * [x] image_rotate/image_rotate

mikeferguson:
Fixed in 941

PR #941, opened 2024-02-12
fix image publisher remapping

mikeferguson:
Addresses #940 - fixes the compressed/etc topic remapping for publishers

### DIFF
--- a/image_proc/src/debayer.cpp
+++ b/image_proc/src/debayer.cpp
@@ -56,10 +56,14 @@ DebayerNode::DebayerNode(const rclcpp::NodeOptions & options)
   // TransportHints does not actually declare the parameter
   this->declare_parameter<std::string>("image_transport", "raw");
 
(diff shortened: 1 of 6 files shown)


image_proc/src/debayer.cpp
```diff
@@ -56,10 +56,14 @@ DebayerNode::DebayerNode(const rclcpp::NodeOptions & options)
   // TransportHints does not actually declare the parameter
   this->declare_parameter<std::string>("image_transport", "raw");
 
+  debayer_ = this->declare_parameter("debayer", 3);
```
mikeferguson, line 59: slightly unrelated fix - but I noticed this parameter gets declared after we create publisher (and publisher callback) - so I moved this up with other parameters

ahcorde (approved):

Nosille:
This fix seems to have overlooked the rectify_node.

PR #943, opened 2024-02-17
924 introduce default depth param

### LINKED ISSUE #924: range_max sensor value introduced but not used
I was wondering why in point_cloud_xyz.cpp in line 112 convertDepth is called but there is no way to set the max_sensor range that is part of the function based on a parameter given to the nodelet. I mean why implement the max sensor range variable and then not make it usable?

The option to set this parameter when calling convertDepth is not used in a single nodelet definition and would be a really useful tool for map generation. Perhaps someone can shed some light on that choice.

PhilippPolterauer:
this pull request attempts to fix #924 

changes:
- [x] changed the argument name `range_max` which was missleading to `invalid_depth` which captures the true meaning of the paramter
- [x] Introduction of a parameter called `invalid_depth`, which is the value that will be assigned to points that are not valid according to `DepthTraits<T>::valid`
- [x] small fixes regarding use of non const shared ptr references
- [x] changed for loop from int to uint32_t avoiding many type casts

Open:

- [ ] UB when reinterpret casting the pointcloud data? 
- [ ] introduction of a true "max_depth" or "max_range" paramter 

Comment:
i think the term range and depth is not always used consistently, my understanding is
*range* meaning point to point distance, and *depth* point to camera plane distance

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,3 @@
 *pyc
-.vscode/settings.json
+.vscode/
 */doc/generated
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,3 @@
 *pyc
 .vscode/
 */doc/generated
-/build/
-/log/
-/install/
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
 *pyc
 .vscode/
 */doc/generated
+/build/
+/log/
+/install/
--- a/depth_image_proc/include/depth_image_proc/conversions.hpp
+++ b/depth_image_proc/include/depth_image_proc/conversions.hpp
@@ -51,9 +51,9 @@ namespace depth_image_proc
 template<typename T>
 void convertDepth(
   const sensor_msgs::msg::Image::ConstSharedPtr & depth_msg,
-  sensor_msgs::msg::PointCloud2::SharedPtr & cloud_msg,
+  const sensor_msgs::msg::PointCloud2::SharedPtr & cloud_msg,
   const image_geometry::PinholeCameraModel & model,
-  double default_depth = 0.0)
+  double invalid_depth = 0.0)
 {
   // Use correct principal point from calibration
   float center_x = model.cx();
@@ -64,18 +64,18 @@ void convertDepth(
   float constant_x = unit_scaling / model.fx();
   float constant_y = unit_scaling / model.fy();
   float bad_point = std::numeric_limits<float>::quiet_NaN();
-  
-  // the following lines ensure that the computation only happens in case we have a default depth
-  const T default_depth_cvt;
-  bool use_default_depth = default_depth != 0.0;
-  if (use_default_depth) {
-    default_depth_cvt=DepthTraits<T>::fromMeters(default_depth);
+
+  // ensure that the computation only happens in case we have a default depth
+  T invalid_depth_cvt = T(0);
+  bool use_invalid_depth = invalid_depth != 0.0;
+  if (use_invalid_depth) {
+    invalid_depth_cvt = DepthTraits<T>::fromMeters(invalid_depth);
   }
   sensor_msgs::PointCloud2Iterator<float> iter_x(*cloud_msg, "x");
   sensor_msgs::PointCloud2Iterator<float> iter_y(*cloud_msg, "y");
   sensor_msgs::PointCloud2Iterator<float> iter_z(*cloud_msg, "z");
 
-  // TODO: i think this is undefined behaviour we should favor memcpy? https://stackoverflow.com/questions/55150001/vector-with-reinterpret-cast
+  // TODO(philipppolterauer): i think this is undefined behaviour we should favor memcpy? https://stackoverflow.com/questions/55150001/vector-with-reinterpret-cast
   const T * depth_row = reinterpret_cast<const T *>(&depth_msg->data[0]);
   uint32_t row_step = depth_msg->step / sizeof(T);
   for (uint32_t v = 0; v < cloud_msg->height; ++v, depth_row += row_step) {
@@ -84,8 +84,8 @@ void convertDepth(
 
       // Missing points denoted by NaNs
       if (!DepthTraits<T>::valid(depth)) {
-        if (use_default_depth) {
-          depth = default_depth_cvt;
+        if (use_invalid_depth) {
+          depth = invalid_depth_cvt;
         } else {
           *iter_x = *iter_y = *iter_z = bad_point;
           continue;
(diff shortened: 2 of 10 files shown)


.gitignore
```diff
@@ -1,3 +1,6 @@
 *pyc
-.vscode/settings.json
+.vscode/
 */doc/generated
+/build/
+/log/
+/install/
```
mikeferguson: these folders shouldn't be inside your checked out repo - the usual layout is:

```
build
log
install
src
   this_repo
   other_repos
```

See https://docs.ros.org/en/rolling/Tutorials/Beginner-Client-Libraries/Creating-A-Workspace/Creating-A-Workspace.html

depth_image_proc/include/depth_image_proc/conversions.hpp
```diff
@@ -51,9 +51,9 @@ namespace depth_image_proc
 template<typename T>
 void convertDepth(
   const sensor_msgs::msg::Image::ConstSharedPtr & depth_msg,
-  sensor_msgs::msg::PointCloud2::SharedPtr & cloud_msg,
+  const sensor_msgs::msg::PointCloud2::SharedPtr & cloud_msg,
   const image_geometry::PinholeCameraModel & model,
-  double range_max = 0.0)
+  double invalid_depth = 0.0)
```
mikeferguson, line 56: I actually think range_max is a better name simply because that's more in line with what a number of depth camera drivers define (as well as most laser range finders). I'd suggest we keep the name of this - but add a doxygen comment to the function to explain what range_max is used for.

depth_image_proc/include/depth_image_proc/conversions.hpp
```diff
@@ -65,19 +65,27 @@ void convertDepth(
...
+  T invalid_depth_cvt = T(0);
+  bool use_invalid_depth = invalid_depth != 0.0;
+  if (use_invalid_depth) {
+    invalid_depth_cvt = DepthTraits<T>::fromMeters(invalid_depth);
+  }
   sensor_msgs::PointCloud2Iterator<float> iter_x(*cloud_msg, "x");
   sensor_msgs::PointCloud2Iterator<float> iter_y(*cloud_msg, "y");
   sensor_msgs::PointCloud2Iterator<float> iter_z(*cloud_msg, "z");
+
+  // TODO(philipppolterauer): i think this is undefined behaviour we should favor memcpy? https://stackoverflow.com/questions/55150001/vector-with-reinterpret-cast
```
mikeferguson: this isn't undefined behavior - the PointCloud2Iterator is actually designed for exactly this use case - it computes what the offset is for each of the x/y/z elements inside the point cloud buffer

.gitignore
PhilippPolterauer: Agreed i will remove the changes in the gitigniore.

depth_image_proc/include/depth_image_proc/conversions.hpp
PhilippPolterauer: Did you check the implementation? If i Recall correctly ( i am currently writing from my phone) this parameter is only used for invalid points. So if there are Points with more than `range_max` distance (depth) they will not be clamped to this value, which ist contrary to what its name indicates IMHO

depth_image_proc/include/depth_image_proc/conversions.hpp
```diff
@@ -65,19 +65,27 @@ void convertDepth(
...
+  for (uint32_t v = 0; v < cloud_msg->height; ++v, depth_row += row_step) {
+    for (uint32_t u = 0; u < cloud_msg->width; ++u, ++iter_x, ++iter_y, ++iter_z) {
       T depth = depth_row[u];
 
       // Missing points denoted by NaNs
       if (!DepthTraits<T>::valid(depth)) {
-        if (range_max != 0.0) {
-          depth = DepthTraits<T>::fromMeters(range_max);
+        if (use_invalid_depth) {
+          depth = invalid_depth_cvt;
```
PhilippPolterauer, line 87: here you can see that the `range_max` values only use was to replace invalid data with this, but not limit other points.

depth_image_proc/include/depth_image_proc/conversions.hpp
PhilippPolterauer: >  const T * depth_row = reinterpret_cast<const T *>(&depth_msg->data[0]);

this is the line my comment is refering to.  I am uncertain about strict aliasing rules here. as there is no constraint on T guaranteeing this should be defined behavior. My suggestion would be to use std::copy which should in most trivial cases be optimized into a basic reinterpretation.

PhilippPolterauer:
in the meantime i further research the UB topic, and after watching [Video](https://www.youtube.com/watch?v=_qzMpk-22cc) i am now fairly certain, that most use of reinterpret_cast is actually UB. Should i raise a new issue dedicated to this?

mikeferguson:
Yeah, go ahead and open a new issue where we can further investigate the reinterpret_cast

depth_image_proc/include/depth_image_proc/conversions.hpp
mikeferguson: Let's remove the comment - but file a ticket to investigate

depth_image_proc/src/point_cloud_xyz.cpp
```diff
@@ -58,6 +58,9 @@ PointCloudXyzNode::PointCloudXyzNode(const rclcpp::NodeOptions & options)
   // Read parameters
   queue_size_ = this->declare_parameter<int>("queue_size", 5);
 
+  // values used for invalid points for pcd conversion
+  invalid_depth_ = this->declare_parameter<double>("invalid_depth", 0.0);
```
mikeferguson, line 62: for completeness, we should probably add this parameter to the xyzrgb and xyzi components. This new parameter also needs to be added to the documentation for each of the components - https://github.com/ros-perception/image_pipeline/blob/rolling/depth_image_proc/doc/components.rst

mikeferguson (changes requested):
A couple of minor items - once we address those, this can be merged (I'm going to just double check things by firing up my robot on this branch tomorrow to make sure the default behavior hasn't changed even though it all looks correct)

PhilippPolterauer:
Thanks for the feedback. As soon as i have time, i will implement the changes.

depth_image_proc/include/depth_image_proc/conversions.hpp
PhilippPolterauer:
```suggestion
```

PhilippPolterauer:
- [x] added `invalid_depth` to parameter description in the docs
- [x] removed comment
- [x] added the `invalid_depth` to the other  non radial nodes as well.

If there is something more to do, please let me know!

mikeferguson (approved):

mikeferguson:
This looks good - thanks for the contribution - just waiting on the final CI to pass

PhilippPolterauer:
i just clicked `update branch` which i think was a mistake, since now the ci is waiting again. sorry for the confusion :)

Issue #935, opened 2024-02-08
Merge ExtractImagesNode into ImageSaver


mikeferguson:
Functionality is largely the same, we should:

 * [ ] Replace "secs_per_frame" by adding new "fps" parameter to ImageSaver - when the parameter is set, we should setup a timer to fire and save images. (see also https://github.com/ros-perception/image_pipeline/issues/274)
 * [ ] Update file format to be "frameX" rather than "leftX"
 * [ ] Use the encoding of the image message, falling back to the "encoding" parameter only when empty.

artineering:
Hello Mike,
I have some cycles this month to take a look at this. Will take a stab at it soon.

PR #944, opened 2024-02-18
Initial Merging: Setting up ImageSaverNode

CursedRock17:
This PR looks to solve issue #935 by Merging `ExtractImagesNode` into `ImageSaver`. Before we go ahead and do it, I assume we want `ExtractImagesNode` entirely removed since ImageSaver will now have all the functionailty and more of that previous node, opposed to just inheriting from that class.


mikeferguson:
Yes - there really isn't any reason to have two separate nodes that have similar, but not identical features. If image_saver supports saving at a particular rate, then it replaces the functionality of extract images node

image_view/src/image_saver_node.cpp
```diff
@@ -125,6 +126,7 @@ bool ImageSaverNode::saveImage(
 {
   cv::Mat image;
   try {
+    encoding_ = image_msg->encoding;
     image = cv_bridge::toCvShare(image_msg, encoding_)->image;
```
mikeferguson: I would suggest what we probably want to do here instead is:

```
std::string encoding = image_msg->encoding.empty() ? encoding_ : image_msg->encoding;
image = cv_bridge::toCvShare(image_msg, encoding)->image;
```

This lets us specify the encoding via parameter if for some reason there is no encoding in the image, but use the message provided encoding if it exists.

image_view/src/image_saver_node.cpp
```diff
@@ -93,7 +93,8 @@ ImageSaverNode::ImageSaverNode(const rclcpp::NodeOptions & options)
       &ImageSaverNode::callbackWithoutCameraInfo, this, std::placeholders::_1),
     hints.getTransport());
 
-  g_format = this->declare_parameter("filename_format", std::string("left%04i.%s"));
+  g_format = this->declare_parameter("filename_format", std::string("frame%04i.%s"));
+  fps_ = this->declare_parameter("fps", 0.1);
```
mikeferguson: we won't want to change the default behavior of the image saver node - so the default here should probably be 0.0 (in which case we don't do the fps-defined behavior)

CursedRock17:
> Adding [a] new "fps" parameter to ImageSaver - when the parameter is set, we should setup a timer to fire and save images.

Do we need to make a `TimerBase` with a new callback, or does that seem excessive? Furthermore, if we had that callback we would need to get access to that image, from inside the constructor of `ImageSaverNode`

mikeferguson:
> Do we need to make a `TimerBase` with a new callback, or does that seem excessive? Furthermore, if we had that callback we would need to get access to that image, from inside the constructor of `ImageSaverNode`

The end goal is to save at a fixed rate, regardless of the rate that the image shows up - I haven't come up with anything better than a ROS timer. The timer would require another callback (the function signature is different). When FPS parameter is set, we would want the image callback to save the latest image in a class variable so that the timer callback can access it (and that also means adding a mutex to protect that shared variable in case somebody loads the component with a multi-threaded executor).

image_view/include/image_view/image_saver_node.hpp
```diff
@@ -68,6 +68,7 @@ class ImageSaverNode
 
 private:
   std::string g_format;
+  int fps_;
```
CursedRock17, line 71: We can leave `fps_` as an int since there's no weird math similar to what was happening in `secs_per_frame`, right?

CursedRock17:
@mikeferguson friendly ping, this is ready for review

ahcorde (changes requested):
conflicts

Issue #582, opened 2020-08-04
 MODE_SGBM_3WAY in opencv


copaah:
I am having a hard time understanding how to set the stereo mode in SGBM. 

[The opencv documentation](https://docs.opencv.org/3.4/d2/d85/classcv_1_1StereoSGBM.html) mentions four different modes. is there support for all four in `stereo_image_proc` and if so, how to set the mode to `MODE_SGBM_3WAY`?

mikeferguson:
It appears this is not currently supported - PRs welcome for adding ROS parameters to set the mode

PR #945, opened 2024-02-18
Initial commit: Exchange boolean full_dp to integer sgbm_mode.

PabloDArandaR:
This PR aims to solve the issue #582. Currently, only the SGBM and HH modes are allowed. To change this behavior, I switched the full_dp boolean parameter to the sgbm_mode parameter. This could be problematic for some users that are currently using the full_dp parameter.

### DIFF
--- a/image_pipeline/doc/changelog.rst
+++ b/image_pipeline/doc/changelog.rst
@@ -38,3 +38,7 @@ There are several major change between ``Iron`` and ``Jazzy``:
    ``rgb/image_rect_color`` to ``rgb/image_raw`` to make clear that the
    unrectified camera projection matrix is used, and for consistency with
    other radial nodes.
+ * The boolen parameter ``full_dp`` from the DisparityNode has been deleted
+   and a new integer parameter ``sgbm_mode`` added to enable all the 
+   variations of the stereo matching algorithm SGBM available from the
+   OpenCV library.
--- a/stereo_image_proc/doc/components.rst
+++ b/stereo_image_proc/doc/components.rst
@@ -30,6 +30,14 @@ Published Topics
 Parameters
 ^^^^^^^^^^
 
+*Disparity algorithm variant*
+ * **sgbm_mode** (int, default: 0): Stereo matching algorithm variation:
+
+   * SGBM (0)
+   * HH (1)
+   * SGBM_3WAY (2)
+   * HH4 (3)
+
 *Disparity pre-filtering* 
 
  * **prefilter_size** (int, default: 9): Normalization window size, pixels.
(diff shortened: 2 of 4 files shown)


mikeferguson (changes requested):
Thank you for the contribution - code looks good - we just need to update the documentation:

 * Update the parameters here: https://github.com/ros-perception/image_pipeline/blob/rolling/stereo_image_proc/doc/components.rst (see https://github.com/ros-perception/image_pipeline/blob/rolling/image_proc/doc/components.rst for examples of how to document the 4 mode variables)
 * Add a note about the (breaking) change to params in the changes page: https://github.com/ros-perception/image_pipeline/blob/rolling/image_pipeline/doc/changelog.rst

PabloDArandaR:
@mikeferguson Shouldn't the edited changes page be https://github.com/ros-perception/image_pipeline/blob/rolling/stereo_image_proc/CHANGELOG.rst ?

stereo_image_proc/CHANGELOG.rst
```diff
@@ -2,6 +2,11 @@
 Changelog for package stereo_image_proc
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
+5.0.1 (####-##-##)
+------------------
+* Breaking change: deleted full_dp parameter and created new parameter sgbm_mode to select the algorithm for stereo matching (`https://github.com/ros-perception/image_pipeline/issues/582`)
+* Contributors: Pablo David Aranda
```
mikeferguson: ah - this part is autogenerated - we can revert the changes to this file (it will get added when I do the next release - but editing it manually like this actually breaks the release tools).

The file I was asking to update is the actual documentation that we use for big changes between releases (so that when you upgrade from say humble/iron->jazzy, you know what changed) - that gets rendered here: http://docs.ros.org/en/rolling/p/image_pipeline/changelog.html - and comes from the file previously mentioned here https://github.com/ros-perception/image_pipeline/blob/rolling/image_pipeline/doc/changelog.rst - and yes, it's in a different "package" - because image_pipeline is the "metapackage" that rolls up all of these. When the Jazzy release gets closer, we'll end up also linking this page into the main documentation - in a page like this: http://docs.ros.org/en/rolling/Releases/Release-Iron-Irwini.html#changes-since-the-humble-release

I've added notes about this to https://github.com/ros-perception/image_pipeline/issues/946 - where I'm going to update our CONTRIBUTIONS.md file so these sort of things are better documented

mikeferguson (changes requested):

stereo_image_proc/CHANGELOG.rst
PabloDArandaR: Okay, in the meanwhile I will add it to that file in the section that describes the changes between Iron and Jazzy, but the changes are being merged into the rolling branch, should I change the target?

mikeferguson: Rolling is the correct branch - Jazzy doesn't exist yet (the buildfarm will come on line in probably 3-4 weeks, and then we can make initial releases into Jazzy, although we typically wait to fork off a branch until we are really forced to since every extra branch adds to the maintainer overhead)

PabloDArandaR: Perfect, all the changes are done then.

mikeferguson (approved):

PR #947, opened 2024-02-21
add downsample_rate param and use image header header time. #274 

### LINKED ISSUE #274: sec_per_frame_ is unintuitive
The [preprocessor variable](https://github.com/ros-perception/image_pipeline/blob/indigo/image_view/src/nodes/extract_images.cpp#L119) `VIDEO_` makes no sense in the `extract_images` node.
Due to that, the very un-intuitive parameter `sec_per_frame_` is necessary in the [callback](https://github.com/ros-perception/image_pipeline/blob/indigo/image_view/src/nodes/extract_images.cpp#L112).

So I vote for getting rid of `VIDEO_` due to `video_recorder` node.
Then, `sec_per_frame_` is no longer necessary by that definition.
Its default parameter can be then set to 0, or it should completely replaced by e.g. `store_nth_frame` to store ever n'th frame which respects the image stream more precisely.

mishra39:
This PR aims to resolve issue #274 where sec_per_frame_ parameter is unintuitive.

- [✔️  ] Add new parameter to extract every Nth frame.
- [ ✔️ ] Replace time::now() with msg->header.time to compute time between frames


mikeferguson:
I haven't really reviewed this yet - but it is going to conflict heavily with #944 (which is trying to remove this node entirely, will merging the functionality into image_saver)

Issue #589, opened 2020-08-21
Downsampling in calibrator.py makes detection fail


martimorta-at-work:
Hi,
I was having issues detecting a checkerboard until I changed the downsampling size (below) from VGA for the actual size of the camera output.
https://github.com/ros-perception/image_pipeline/blob/458ff618a53ec92713544c3ff24d2153975d8f3f/camera_calibration/src/camera_calibration/calibrator.py#L523

I think having that as a parameter would allow better calibrations unless I am missing something here.
I am happy to submit a PR with VGA size being a parameter if there aren't other implications.

Cheers!

JWhitleyWork:
@martimorta-at-work I know it's going on almost a year since you submitted this but we would be happy to accept a PR with this.

onurtore:
@JWhitleyWork,
Still interested? I can create a PR

MedisettiRenukeswar:
Hi, I’d like to work on this issue.

My plan is to replace the hardcoded VGA resolution (640x480) in calibrator.py with configurable parameters, while keeping default values unchanged for backward compatibility.

Please let me know if this approach sounds good.

MedisettiRenukeswar:
Submitted a PR to address this issue: 
https://github.com/MedisettiRenukeswar/image_pipeline/pull/1

PR #951, opened 2024-02-26
Vga scale

CursedRock17:
Addresses issue #589 which asks to add a `parameter` to the camera calibrator. This allows the user to preset their desired calibration scale for more precision.

### DIFF
--- a/camera_calibration/src/camera_calibration/calibrator.py
+++ b/camera_calibration/src/camera_calibration/calibrator.py
@@ -406,6 +406,9 @@ class Calibrator():
     def set_cammodel(self, modeltype):
         self.camera_model = modeltype
 
+    def set_vga_scale(self, scale):
+        self.vga_scale = scale
+
     def is_slow_moving(self, corners, ids, last_frame_corners, last_frame_ids):
         """
         Returns true if the motion of the checkerboard is sufficiently low between
@@ -537,7 +540,7 @@ class Calibrator():
         height = img.shape[0]
         width = img.shape[1]
         if self.vga_scale == 0:
-            scale = math.sqrt((width*height) / (640.*480.))
+            scale = math.sqrt((width * height) / (640. * 480.))
         else:
             scale = self.vga_scale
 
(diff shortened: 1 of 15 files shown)


camera_calibration/src/camera_calibration/calibrator.py
```diff
@@ -343,6 +343,7 @@ def __init__(self, boards, flags=0, fisheye_flags = 0, pattern=Patterns.Chessboa
         self.pattern = pattern
         self.br = cv_bridge.CvBridge()
         self.camera_model = CAMERA_MODEL.PINHOLE
+        self.declare_parameter('vga_scale', 0)
```
ahcorde: `Calibrator` is not a ROS Node, this code is incorrect. You can check some failures in the CI

ahcorde (changes requested):

camera_calibration/src/camera_calibration/calibrator.py
CursedRock17: If that's the case, I assume it's more practical to make the `scale`, a class member opposed to a function param since it's not required and would break API

camera_calibration/src/camera_calibration/calibrator.py
```diff
@@ -343,6 +343,7 @@ def __init__(self, boards, flags=0, fisheye_flags = 0, pattern=Patterns.Chessboa
         self.pattern = pattern
         self.br = cv_bridge.CvBridge()
         self.camera_model = CAMERA_MODEL.PINHOLE
+        self.vga_scale = 0
```
ahcorde, line 346: do we need a method to set this attribute ?

camera_calibration/src/camera_calibration/calibrator.py
```diff
@@ -535,7 +536,11 @@ def downsample_and_detect(self, img):
         # Scale the input image down to ~VGA size
         height = img.shape[0]
         width = img.shape[1]
-        scale = math.sqrt( (width*height) / (640.*480.) )
+        if self.vga_scale == 0:
+            scale = math.sqrt((width*height) / (640.*480.))
```
ahcorde:
```suggestion
            scale = math.sqrt((width * height) / (640. * 480.))
```

ahcorde (changes requested):

camera_calibration/src/camera_calibration/calibrator.py
CursedRock17: Another method has been added, just a simple way to set the scale

Issue #952, opened 2024-02-26
image_proc / image_proc.launch.py not working with namespace for ROS2 Iron


aaalloc:
Hey, the current version of  ```image_proc.launch.py``` on iron branch is not updated like on the rolling branch, so it is not possible to run ```image_proc.launch.py``` like this ```ros2 launch image_proc image_proc.launch.py namespace:=my_camera``` as stated on the documentation https://docs.ros.org/en/rolling/p/image_proc/tutorials.html#using-image-proc-launch-file

https://github.com/ros-perception/image_pipeline/blame/rolling/image_proc/launch/image_proc.launch.py#L46

https://github.com/ros-perception/image_pipeline/blame/iron/image_proc/launch/image_proc.launch.py

mikeferguson:
Launch file changes backported to Iron (note that some other changes, like the camera_info remapping will NOT be back ported - because these could possibly break anyone who worked around them)

PR #953, opened 2024-02-27
[backport iron] Node namespace parameter

CursedRock17:
Backport pull request #925 which solves issue #952 by adding the `namespace` parameter to the `image_proc` launch file

### DIFF
--- a/image_proc/launch/image_proc.launch.py
+++ b/image_proc/launch/image_proc.launch.py
@@ -41,20 +41,27 @@ from launch_ros.descriptions import ComposableNode
 
 
 def generate_launch_description():
+
+    arg_namespace = DeclareLaunchArgument(
+        name='namespace', default_value='',
+        description=('namespace for all components loaded')
+    )
+
     composable_nodes = [
         ComposableNode(
             package='image_proc',
             plugin='image_proc::DebayerNode',
             name='debayer_node',
+            namespace=LaunchConfiguration('namespace'),
         ),
         ComposableNode(
             package='image_proc',
             plugin='image_proc::RectifyNode',
             name='rectify_mono_node',
+            namespace=LaunchConfiguration('namespace'),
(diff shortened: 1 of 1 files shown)


ahcorde (approved):

image_proc/launch/image_proc.launch.py
```diff
@@ -41,27 +41,35 @@
...
         ),
         ComposableNode(
             package='image_proc',
             plugin='image_proc::RectifyNode',
             name='rectify_mono_node',
+            namespace=LaunchConfiguration('namespace'),
             # Remap subscribers and publishers
             remappings=[
                 ('image', 'image_mono'),
-                ('camera_info', 'camera_info'),
```
mikeferguson: we should not remove this - while it isn't doing much in this particular example, Iron does not have the fixes for properly remapping camera_info topics alongside the image topic (and it won't get those fixes, since it would possibly break anyone who has manually remapped them appropriately)

mikeferguson (changes requested):

mikeferguson (approved):

image_proc/launch/image_proc.launch.py
```diff
@@ -41,16 +41,24 @@
 
 
 def generate_launch_description():
+
+    arg_namespace = DeclareLaunchArgument(
+        name='namespace', default_value='',
+        description=('namespace for all components loaded')
+    )
+
```
mikeferguson: linter seems to think that you have some spaces on this line (it shows in the individual second commit as well -although I can't actually see it highlighted here as part of the PR - which is odd)

CursedRock17: Yea I noticed the same thing when the `flake` test failed, just assumed it was a CI glitch. I've tried to retrigger by removing any sort of white space that may have come up

PR #959, opened 2024-03-08
Fix parameter names in components.rst

NoahMollerstuen:
In the docs for  `image_proc::CropDecimateNode` , change the parameter names `x_offset` and `y_offset` to `offset_x` and `offset_y`, matching the actual names of parameters defined  in [crop_decimate.cpp](https://github.com/ros-perception/image_pipeline/blob/b8839433dda96c7671e4f9b00fad37a9824fc0e7/image_proc/src/crop_decimate.cpp#L130-L131).

### DIFF
--- a/image_proc/doc/components.rst
+++ b/image_proc/doc/components.rst
@@ -42,8 +42,8 @@ Parameters
  * **queue_size** (int, default: 5): Size of message queue for synchronizing
    image and camera_info topics. You may need to raise this if images take
    significantly longer to travel over the network than camera info.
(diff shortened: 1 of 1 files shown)


mikeferguson (approved):

mikeferguson:
Thanks for catching that!

CI is going to fail due to rolling transition - but this doesn't change code, so merging.

NoahMollerstuen:
Wow, thanks for the quick review!

mikeferguson:
> Wow, thanks for the quick review!

You got lucky, I typically do one or two checks a day on GitHub notifications, and this came through just as I was doing that check.

Issue #969, opened 2024-04-23
disutils deprecated as of python 3.10, removed 3.12


ScottMonaghan:
While working on #968 and #572 I discovered the following remaining warning in `camera_calibration` when running colcon test:
```
=============================== warnings summary ===============================

src/camera_calibration/calibrator.py:47

  Warning: The distutils package is deprecated and slated for removal in Python 3.12. Use setuptools or check PEP 632 for potential alternatives

```

iirc, I believe python 3.12 will be standard with noble, so I think this is a necessary fix.

I'm happy to look into it and see what I can do.

ScottMonaghan:
Fix submitted in PR #970. Once that is merged, this issue can be closed.

PR #970, opened 2024-04-23
replace disutils with python3-semver

ScottMonaghan:
Fix for 
 - #969 
 
I added a dependency for `python3-semver` to replace version parsing with `disutils`.

Please let me know if you have any questions, concerns, or additional requested changes.


mikeferguson (approved):
LGTM - but I'm not a Python wizard - so letting @ahcorde also take a look before merging. The one CI job is failing because Github's "ubuntu-latest" is still 22.04 and there is not yet support for 24.04

ahcorde (approved):

PR #971, opened 2024-04-28
[Iron] Fix aruco dictionary names 

MRo47:
There was a aruco dictionary naming issue in `iron`, this seems to be fixed in `rolling`.  

```
"6x6_50"    : cv2.aruco.DICT_6x6_50,
```
where the `x` had to be changed to `X`

PR #972, opened 2024-04-28
[Iron] Added stereo calibration using aruco markers

MRo47:
This was a TODO in the repository, opening this PR to add this feature.
- The main issue why this wasn't possible imo is the way `mk_obj_points` works. I'm using the inbuilt opencv function to get the points there.
- The other is a condition when aruco markers are detected they are added as good points, This is fine in case of mono but in stereo these have to be the same number as the object points to find matches although this should be possible with aruco.
- This also has the dictionary name corrections I mentioned in #971 

Noob to contributing in this repo, any guidelines, rules I should be following before doing PR's here?


ahcorde (changes requested):
This PR is targeting `iron`, you should change retarget this PR to `rolling`, then if this is not breaking API we can backport the changes to `iron`

MRo47:
@ahcorde , opened a PR for this #976 in rolling a few weeks ago, waiting on this till then.

ahcorde:
Closing this PR  in favour of this other https://github.com/ros-perception/image_pipeline/pull/1003 created with mergify

Issue #973, opened 2024-04-28
Making camera calibration code more maintainable


MRo47:
- `calibrator.py` file is almost 1.5K lines long and difficult to navigate.
- This could break dependencies? I'd like to know more if this is true.
- Ive fixed the imports in the remaining files.

Have something [ready](https://github.com/MRo47/image_pipeline/tree/iron-separate-calib-types/camera_calibration/src/camera_calibration) but thought it would be better to discuss if its even required.

ahcorde:
Can you open the PR ? It's better to discuss this there.

MyronRodrigues-StreetDrone:
sure thing!

Issue #785, opened 2022-11-20
Several Improvements for Calibration Tool


onurtore:
Hi!,
I have been using calibration tool for a while, and I believe there are some features that definitely improve the output of the calibration. I want to create PR for them, but first wanted to get your opinions so that I do not spend unnecessary time.

1. Calibration Requirements

Currently, one of the calibration requirement is to have 40 picture of the pattern. I believe, this should be parametric.

Opencv fish-eye calibration implementation uses this [paper](https://www.researchgate.net/publication/6899685_A_Generic_Camera_Model_and_Calibration_Method_for_Conventional_Wide-Angle_and_Fish-Eye_Lenses/link/02e7e52f39b8fb374d000000/download), and the normal calibration tool uses [this](https://ieeexplore.ieee.org/document/888718), both of the papers does not require 40 images for the calibration, and this number depends on many of the features of the calibration pattern, or the lens type. The default can be 40, but otherwise I believe hardcoding this as 40 is wrong.

2. Save and Load option for calibration data

When I used this tool, I realized that many of the calibration images are blurred, I now there is an option for --max-chessboard-speed however, i believe loading and saving the calibration data through gui will give much more flexibility. The idea is, to ability to save images, and their corresponding parameters to a folder, and then reloading them with the help of gui, this way users can modify/delete/add new images and parameters to the calibration.

I realized that there is a tar calibrator, and it is kinda doing what I referrred, but I believe adding this feature to camera calibrator is a good option.

3. Verbose option

I believe, it will be beneficial to add verbose option to see the re-projection error, number of images and number of control points during the calibration.

JWhitleyWork:
@onurtore I'm fine with all of these. Please make sure to implement them in separate PRs. Thanks!

PR #975, opened 2024-04-29
[Iron] Improve maintainability of calibration code

MRo47:
For #973 

- `calibrator.py` file is almost 1.5K lines long and difficult to navigate.
- This could break dependencies? I'd like to know more if this is true.
- Ive fixed the imports in the remaining files.

### DIFF
--- a/camera_calibration/doc/index.rst
+++ b/camera_calibration/doc/index.rst
@@ -11,8 +11,8 @@ need to extend or make a new calibration tool.
 For details on the camera model and camera calibration process, see
 http://docs.opencv.org/master/d9/d0c/group__calib3d.html
 
-.. autoclass:: camera_calibration.calibrator.MonoCalibrator
+.. autoclass:: camera_calibration.mono_calibrator.MonoCalibrator
     :members:
 
-.. autoclass:: camera_calibration.calibrator.StereoCalibrator
+.. autoclass:: camera_calibration.stereo_calibrator.StereoCalibrator
     :members:
--- a/camera_calibration/scripts/tarfile_calibration.py
+++ b/camera_calibration/scripts/tarfile_calibration.py
@@ -39,21 +39,24 @@ import cv2
 import cv_bridge
 import tarfile
 
-from camera_calibration.calibrator import MonoCalibrator, StereoCalibrator, CalibrationException, ChessboardInfo
+from camera_calibration.mono_calibrator import MonoCalibrator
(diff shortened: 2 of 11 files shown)


mikeferguson:
I see there is some discussion of this in a ticket as well - but I just want to note that a PR of this size/scale should really target Rolling and then be backported as needed to earlier releases.

mikeferguson:
Another thought: it appears that a number of the changes are basically just linting - I would actually suggest to create a standalone PR that is just linting changes (because that will be much easier to review) and then put the structural changes in a follow up PR.

MRo47:
Understood. Will do the linting first, think my auto-formatter did it > PR for rolling

arthurlovekin:
As it happens, I recently rewrote most of the code in the image_pipeline in order to make it cleaner for internal use at my company and also incorporate some features for thermal cameras. I chose to rewrite the code because there were several large structural changes that made it very hard to work with: 
1. The ROS interface is heavily interwoven with OpenCV. For example, all of the OpenCV GUI functionality is embedded in a OpenCVCalibrationNode, making it hard to add additional GUI functionality. Ideally, this package should be a ROS package that provides an interface for a completely independent python calibration library, which itself is just a wrapper for OpenCV (plus a few helper functions to, for instance, ensure good sample-point distribution and provide a GUI).
2. The class structure does not capture the structure of the calibration process well, so there is a lot of duplication between classes. This is most notable in the StereoCalibrator and MonoCalibrator classes.
3. The file names and general file structure could be a lot better. One glaring example is that there is both cameracalibrator.py and also camera_calibrator.py. However, most of the files could be organized better according to their functionality (but first should make changes 1 and 2).
4. Lots of small readability issues, like passing around the sample distribution as a list(zip(self._param_names, min_params, max_params, progress)) as opposed to using a named dictionary or other object.
5.I think there is an opportunity here to also address a number of other Issues/enhancements that have been posted, like #785.

Unfortunately due to export restrictions I cannot simply share my code, and at this point I would actually have to request access to see it again. However, if someone is making large structural changes to this package I would be interested in helping. I have some time right now so I can help with both broad discussion of things I found to be most broken, and also writing code if someone else (@MRo47 or @mikeferguson) can help.

MRo47:
@arthurlovekin I've just opened a PR #1000 

Probably you can add to it?

ahcorde:
Closing this one, PR in `rolling` in merged https://github.com/ros-perception/image_pipeline/pull/1000

Issue #965, opened 2024-04-17
image_publisher_node not read camera_info_url on initialization


Kotochleb:
As in the title, `image_publisher_node` doesn't account for `camera_info_url` on the initialization when launched as a standalone node on Humble. For some reason, it skips reading it. When the same parameter is changed dynamically after the node started, everything works just fine.
The simplest walk-round I found was to change `return result` with `break` in the `pram_change_callback` like so:

```diff
if (parameter.get_name() == "filename") {
    filename_ = parameter.as_string();
    RCLCPP_INFO(get_logger(), "Reset filename as '%s'", filename_.c_str());
    ImagePublisher::onInit();
-   return result;
+   break;
    } else if (parameter.get_name() == "flip_horizontal") {
    flip_horizontal_ = parameter.as_bool();
    RCLCPP_INFO(get_logger(), "Reset flip_horizontal as '%d'", flip_horizontal_);
    ImagePublisher::onInit();
-   return result;
+   break;
  } else if (parameter.get_name() == "flip_vertical") {
    flip_vertical_ = parameter.as_bool();
    RCLCPP_INFO(get_logger(), "Reset flip_vertical as '%d'", flip_vertical_);
    ImagePublisher::onInit();
-   return result;
+   break;
```

Yet, I feel like there should be a better solution to this, and I do not feel confident enough with the original implementation to submit a good enough bugfix PR.

mikeferguson:
I'm not sure why there is a "return result" in any of those cases - I would think we should iterate through all the parameter settings by completing the for loop (and then there is a return after the reconfigure callback is called). I think we should actually remove the lines entirely (not change to a break).

Kotochleb:
> I'm not sure why there is a "return result" in any of those cases

@mikeferguson this is precisely what caused my confusion and said lack of said confidence in making changes. If you think removing the `return` entirely is a fine fix I can submit PR soon.

mikeferguson:
Looks like fix is merged to Rolling, backports in progress

PR #983, opened 2024-05-25
[rolling] image_publisher: Fix loading of the camera info parameters on startup

Kotochleb:
As described in https://github.com/ros-perception/image_pipeline/issues/965 camera info is not loaded from the file on node initialization, but only when the parameter is reloaded.

This PR resolves this issue and should be straightforward to port it to `Humble`, `Iron` and `Jazzy`.


ahcorde (commented):
can you target `rolling` ? then I will backport this to the other distros

Kotochleb:
@ahcorde the changes were retargeted to `rolling` as requested.

image_publisher/src/image_publisher.cpp
```diff
@@ -71,36 +71,46 @@ ImagePublisher::ImagePublisher(
   auto param_change_callback =
     [this](std::vector<rclcpp::Parameter> parameters) -> rcl_interfaces::msg::SetParametersResult
     {
+      RCLCPP_INFO(get_logger(), "param_change_callback");
```
ahcorde: remove

image_publisher/src/image_publisher.cpp
```diff
@@ -224,9 +234,7 @@ void ImagePublisher::onInit()
   camera_info_.p = {1, 0, static_cast<float>(camera_info_.width / 2), 0, 0, 1,
     static_cast<float>(camera_info_.height / 2), 0, 0, 0, 1, 0};
 
-  timer_ = this->create_wall_timer(
```
ahcorde, line 244: if you remove these lines the `dowork` method will never be called and the image will be never flipped

ahcorde (changes requested):

image_publisher/src/image_publisher.cpp
Kotochleb: Done

image_publisher/src/image_publisher.cpp
Kotochleb: The timer is set in the `reconfigureCallback`. I didn't want to repeat the `CameraInfoManager` block of code in `onInit` and did not want to create timer there just to recreate it once again in `reconfigureCallback`.

Kotochleb:
@ahcorde FYI changes from https://github.com/ros-perception/image_pipeline/pull/985 are already merged in here.

ahcorde (approved):

ahcorde:
https://github.com/Mergifyio backport jazzy humble iron

PR #984, opened 2024-05-25
[humble] image_publisher: Fix image constantly flipping when static image is published

Kotochleb:
When publishing video stream from a camera, the image was flipped correctly. Yet for a static image, which was loaded once, the flip function was applied every time `ImagePublisher::doWork()` was called, resulting in the published image being flipped back and forth all the time.

This PR should be straightforward to port it to `Iron`, `Jazzy` and `Rolling`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 image_pipeline
 ==============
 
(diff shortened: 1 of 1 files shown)


ahcorde (changes requested):
Can you target `rolling` ? Then I can backport this to the other distros.

PR #985, opened 2024-05-26
[rolling] image_publisher: add field of view parameter

Kotochleb:
Currently, the default value for focal length when no camera info is provided defaults to `1.0` rendering whole approximate intrinsics and projection matrices useless.  Based on [this article](https://learnopencv.com/approximate-focal-length-for-webcams-and-cell-phone-cameras/), I propose a better approximation of the focal length based on the field of view of the camera. 

For most of the use cases, users will either know the field of view of the camera the used, or they already calibrated it ahead of time.

If there is some documentation to fill. please let me know.

This PR should be straightforward to port it to `Humble`, `Iron` and `Jazzy`.

### DIFF
--- a/image_publisher/src/image_publisher.cpp
+++ b/image_publisher/src/image_publisher.cpp
@@ -33,6 +33,7 @@
 
 #include <cmath>
 #include <chrono>
+#include <limits>
 #include <string>
 #include <thread>
 #include <vector>
@@ -61,7 +62,7 @@ ImagePublisher::ImagePublisher(
   std::string topic_name = node_base->resolve_topic_or_service_name("image_raw", false);
   pub_ = image_transport::create_camera_publisher(this, topic_name);
 
-  field_of_view_ = this->declare_parameter("field_of_view", static_cast<double>(50));
+  field_of_view_ = this->declare_parameter("field_of_view", static_cast<double>(0));
   flip_horizontal_ = this->declare_parameter("flip_horizontal", false);
   flip_vertical_ = this->declare_parameter("flip_vertical", false);
   frame_id_ = this->declare_parameter("frame_id", std::string("camera"));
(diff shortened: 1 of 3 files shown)


image_publisher/src/image_publisher.cpp
```diff
@@ -211,10 +218,13 @@ void ImagePublisher::onInit()
   camera_info_.height = image_.rows;
   camera_info_.distortion_model = "plumb_bob";
   camera_info_.d = {0, 0, 0, 0, 0};
-  camera_info_.k = {1, 0, static_cast<float>(camera_info_.width / 2), 0, 1,
+
+  // Based on https://learnopencv.com/approximate-focal-length-for-webcams-and-cell-phone-cameras/
+  double f_approx = (camera_info_.width / 2) / tan((field_of_view_ * M_PI / 180) / 2);
```
ahcorde:
```suggestion
  double f_approx = (camera_info_.width / 2) / std::tan((field_of_view_ * M_PI / 180) / 2);
```

ahcorde (changes requested):
Do you mind to target `rolling` ? then we can backport this to the other distros.

Kotochleb:
@ahcorde the changes were retargeted to `rolling` as requested, plus I added documentation for the new parameter since this branch accounts for that.

Kotochleb:
@ahcorde the changes were retargeted to `rolling` as requested.

image_publisher/src/image_publisher.cpp
```diff
@@ -218,10 +225,13 @@ void ImagePublisher::onInit()
   camera_info_.height = image_.rows;
   camera_info_.distortion_model = "plumb_bob";
   camera_info_.d = {0, 0, 0, 0, 0};
-  camera_info_.k = {1, 0, static_cast<float>(camera_info_.width / 2), 0, 1,
+
+  // Based on https://learnopencv.com/approximate-focal-length-for-webcams-and-cell-phone-cameras/
+  double f_approx = (camera_info_.width / 2) / std::tan((field_of_view_ * M_PI / 180) / 2);
```
ahcorde: is the default value `50` returning `1` ? as the current behavior ?

ahcorde (changes requested):

image_publisher/src/image_publisher.cpp
Kotochleb: It doesn't, and I forgot to account for breaking changes. I changed it now so when FOV is 0 or less, it disables the whole feature and uses 1.0 for focal length instead. The default vale is now zero as well, so now the whole thing is not breaking anything.

ahcorde (approved):

ahcorde:
https://github.com/Mergifyio backport jazzy humble iron

PR #999, opened 2024-06-07
Formatting calib code before refactoring

MRo47:
As discussed in #975 and #973 
doing the linting first.

using style from [here](https://github.com/ament/ament_lint/blob/rolling/ament_pycodestyle/ament_pycodestyle/configuration/ament_pycodestyle.ini)


mikeferguson (approved):
I concur that this is formatting only changes - will also let @ahcorde have a chance to check that the style selected is correct

ahcorde (approved):

MRo47:
thanks @mikeferguson @ahcorde

PR #1010, opened 2024-06-26
Removed deprecation warnings

ahcorde:
Removed deprecation warnings included recently in message_filters

Related https://build.ros2.org/job/Rpr__image_pipeline__ubuntu_noble_amd64/19/


depth_image_proc/src/disparity.cpp
```diff
@@ -116,7 +116,8 @@ DisparityNode::DisparityNode(const rclcpp::NodeOptions & options)
         std::string topic = node_base->resolve_topic_or_service_name("left/image_rect", false);
         image_transport::TransportHints hints(this);
         sub_depth_image_.subscribe(this, topic, hints.getTransport());
-        sub_info_.subscribe(this, "right/camera_info");
+        sub_info_.subscribe(this, "right/camera_info",
+          rclcpp::QoS(rclcpp::QoSInitialization::from_rmw(rmw_qos_profile_default)));
```
clalancette: I think we can simplify this to:

```suggestion
        sub_info_.subscribe(this, "right/camera_info", rclcpp::QoS(10));
```

(same goes for all of the other instances below)

image_view/include/image_view/disparity_view_node.hpp
```diff
@@ -30,10 +30,8 @@ class DisparityViewNode
 {
 public:
   explicit DisparityViewNode(const rclcpp::NodeOptions & options);
-  explicit DisparityViewNode(const DisparityViewNode &) = default;
-  explicit DisparityViewNode(DisparityViewNode &&) = default;
-  DisparityViewNode & operator=(const DisparityViewNode &) = default;
-  DisparityViewNode & operator=(DisparityViewNode &&) = default;
+  explicit DisparityViewNode(const DisparityViewNode &) = delete;
+  explicit DisparityViewNode(DisparityViewNode &&) = delete;
```
clalancette: Hm, this is changing the node from being copyable to not being copyable.  Can you explain more about why we should do this?

clalancette (changes requested):

image_view/include/image_view/disparity_view_node.hpp
ahcorde: removed, I was testing something

stereo_image_proc/src/stereo_image_proc/disparity_node.cpp
```diff
@@ -326,10 +326,12 @@ DisparityNode::DisparityNode(const rclcpp::NodeOptions & options)
 
         sub_l_image_.subscribe(
           this, left_topic, hints.getTransport(), sensor_data_qos, sub_opts);
-        sub_l_info_.subscribe(this, left_info_topic, sensor_data_qos, sub_opts);
+        sub_l_info_.subscribe(this, left_info_topic,
+          rclcpp::QoS(rclcpp::QoSInitialization::from_rmw(sensor_data_qos)), sub_opts);
```
clalancette: I think we should refactor this a bit.

Above on line 318, we can change sensor_data_qos to:

```
const SensorDataQoS sensor_data_qos;
```

Then in line 327-328, we can change the code to use `get_rmw_qos_profile`:

```
        sub_l_image_.subscribe(
          this, left_topic, hints.getTransport(), sensor_data_qos.get_rmw_qos_profile(), sub_opts);
```

And then finally, on this line, we can do:

```suggestion
        sub_l_info_.subscribe(this, left_info_topic, sensor_data_qos, sub_opts);
```

I think that is a lot more readable than using the `QoSInitialization::from_rmw` method.

stereo_image_proc/src/stereo_image_proc/disparity_node.cpp
```diff
@@ -326,10 +326,12 @@ DisparityNode::DisparityNode(const rclcpp::NodeOptions & options)
...
         sub_r_image_.subscribe(
           this, right_topic, hints.getTransport(), sensor_data_qos, sub_opts);
-        sub_r_info_.subscribe(this, right_info_topic, sensor_data_qos, sub_opts);
+        sub_r_info_.subscribe(this, right_info_topic,
+          rclcpp::QoS(rclcpp::QoSInitialization::from_rmw(sensor_data_qos)), sub_opts);
```
clalancette: Same here.

stereo_image_proc/src/stereo_image_proc/point_cloud_node.cpp
```diff
@@ -191,9 +191,12 @@ PointCloudNode::PointCloudNode(const rclcpp::NodeOptions & options)
...
           this, left_topic, hints.getTransport(), sensor_data_qos, sub_opts);
-        sub_l_info_.subscribe(this, left_info_topic, sensor_data_qos, sub_opts);
-        sub_r_info_.subscribe(this, right_topic, sensor_data_qos, sub_opts);
-        sub_disparity_.subscribe(this, disparity_topic, sensor_data_qos, sub_opts);
+        sub_l_info_.subscribe(this, left_info_topic,
+          rclcpp::QoS(rclcpp::QoSInitialization::from_rmw(sensor_data_qos)), sub_opts);
+        sub_r_info_.subscribe(this, right_topic,
+          rclcpp::QoS(rclcpp::QoSInitialization::from_rmw(sensor_data_qos)), sub_opts);
+        sub_disparity_.subscribe(this, disparity_topic,
+          rclcpp::QoS(rclcpp::QoSInitialization::from_rmw(sensor_data_qos)), sub_opts);
```
clalancette: And the same basic idea here.

clalancette (changes requested):

clalancette (approved):

PR #1020, opened 2024-08-06
switch non testing build to devel

mikeferguson:


mikeferguson:
@ahcorde this is now failing for the right reasons (hpp file not present) - but I know it's not the cleanest approach - is there a better docker container we should be using? Or maybe we just manually define the ROS_DISTRO variable here in the config.yaml? (I've not really worked with GitHub CI recently...)

EDIT: in theory, we can also remove a bunch of the apt configuration stuff in the build-rolling-testing job, since the docker is already using the ros2-testing debian repository

ahcorde (approved):

mikeferguson:
Re-running CI now that rolling sync is out - I expect all tests to pass (and then I'll merge this)

Issue #1030, opened 2024-09-04
image_proc's TrackMarkerNode requires OpenCV 4.5.1 or greater


tedsteiner:
The `TrackMarkerNode` [includes](https://github.com/ros-perception/image_pipeline/blob/dbf697de8ac6e3f06044bc920a36785b28cb1cf4/image_proc/src/track_marker.cpp#L43) the OpenCV header `<opencv2/core/quaternion.hpp>`. Currently the `image_proc` [CMakeLists.txt](https://github.com/ros-perception/image_pipeline/blob/dbf697de8ac6e3f06044bc920a36785b28cb1cf4/image_proc/CMakeLists.txt#L17) file specifies that OpenCV 3.2.0 is the minimum required version. However, [quaternion.hpp](https://github.com/opencv/opencv/blob/4.x/modules/core/include/opencv2/core/quaternion.hpp) wasn't added to OpenCV until version 4.5.1. The result is a compiler error on earlier versions of OpenCV, such as those provided by Ubuntu 20.04 (4.2.0).

At first I thought this was maybe just a header left unintentionally because it's a bit subtle, but here's where the OpenCV quaternion header is used:
```c++
      // Convert angle-axis to quaternion
      cv::Quatd q = cv::Quatd::createFromRvec(rvecs[0]);
      pose.pose.orientation.x = q.x;
      pose.pose.orientation.y = q.y;
      pose.pose.orientation.z = q.z;
      pose.pose.orientation.w = q.w;
      pub_->publish(pose);
```
Is there maybe some other code that already exists somewhere to do this conversion without using the `cv::Quatd` class?

This causes a build error when trying to build ROS 2 Jazzy from source with the perception variant on Ubuntu 20.04 with the Apt-provided version of libopencv-dev. The original merge request was here: https://github.com/ros-perception/image_pipeline/pull/930

mikeferguson:
A couple thoughts:

 * We should fix the opencv version requirement on the Jazzy/rolling branches to be correct - that is an oversight
 * Jazzy targets 24.04, 20.04 is pretty old for us to also target.
 * It looks like even 22.04 has opencv 4.5.4 - so that meets the minimum all the way back to Humble (which, didn't actually get this feature backported)

mikeferguson:
Note: this is not to say we wouldn't merge a PR that addressed this so it would run on 20.04 - it's just not something the maintainers are likely to find time to tackle

tedsteiner:
[...]

PR #1031, opened 2024-09-06
Use TF2 package for quaternion conversion

tedsteiner:
The OpenCV quaternion class was not added until OpenCV 4.5.1, so it's less widely available than the TF2 conversion. This change allows a source build of the ROS 2 "perception" variant on Ubuntu 20.04 without a custom source build of OpenCV.

Addresses issue https://github.com/ros-perception/image_pipeline/issues/1030

### DIFF
--- a/image_proc/package.xml
+++ b/image_proc/package.xml
@@ -30,6 +30,8 @@
   <depend>rclcpp_components</depend>
   <depend>rcutils</depend>
   <depend>sensor_msgs</depend>
+  <depend>tf2</depend>
+  <depend>tf2_geometry_msgs</depend>
   <depend>tracetools_image_pipeline</depend>
 
   <test_depend>ament_lint_auto</test_depend>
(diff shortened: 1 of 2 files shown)


tedsteiner:
Hoping for a backport to Jazzy, as well.

mikeferguson:
If I recall correctly, you're unable to test this, right? I can test on a robot early next week to make sure there is no change in behavior.

tedsteiner:
> If I recall correctly, you're unable to test this, right? I can test on a robot early next week to make sure there is no change in behavior.

Thanks! This isn't a code package we use so I'd really appreciate if there's a way for someone else to verify the change without too much effort.

My interpretation of the build issues is that they're CppLint issues that aren't directly related to this change. One of them is that the track_marker.hpp include should come first and the other is incorrectly thinking that the TF2 quaternion.h header is a C system library because it's not *.hpp. Just let me know if you'd like me to do anything about those.

tedsteiner:
Just wanted to check in on this to see if you've had a chance to test it yet, and if so, whether it worked correctly. Thanks!

mikeferguson:
I'm back from ROSCon now - will take a look at this during the week

mikeferguson (approved):
I've tested locally - it looks good - just need to get CI to pass

tedsteiner:
Thank you! And would it be possible to also get this backported to Jazzy? It would save me a lot of maintenance effort.

mikeferguson:

PR #1033, opened 2024-09-15
Fix #1032

### LINKED ISSUE #1032: Accessing non-existing attribute in SetCameraInfo response handling
When the `SetCameraInfo` service call fails, `cameracalibrator` prints:
```
Traceback (most recent call last):
  File "/opt/ros/jazzy/lib/python3.12/site-packages/camera_calibration/camera_calibrator.py", line 304, in on_mouse
    if self.do_upload():
       ^^^^^^^^^^^^^^^^
  File "/opt/ros/jazzy/lib/python3.12/site-packages/camera_calibration/camera_calibrator.py", line 217, in do_upload
    rv = self.check_set_camera_info(response)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/ros/jazzy/lib/python3.12/site-packages/camera_calibration/camera_calibrator.py", line 199, in check_set_camera_info
    print("Attempt to set camera info failed: " + response.result() if response.result() is not None else "Not available")
                                                                       ^^^^^^^^^^^^^^^
AttributeError: 'SetCameraInfo_Response' object has no attribute 'result'
```
It should print the message from the `SetCameraInfo` response though.

ugol-1:
Use `status_message` from `SetCameraInfo` response instead of some non-existing `result()`.
Fixes #1032

### DIFF
--- a/camera_calibration/src/camera_calibration/camera_calibrator.py
+++ b/camera_calibration/src/camera_calibration/camera_calibrator.py
@@ -214,14 +214,14 @@ class CalibrationNode(Node):
         for i in range(10):
             print("!" * 80)
         print()
-        print("Attempt to set camera info failed: " + response.result()
-              if response.result() is not None else "Not available")
+        print("Attempt to set camera info failed: " + response.status_message
+              if response.status_message is not None else "Not available")
         print()
         for i in range(10):
             print("!" * 80)
         print()
         self.get_logger().error('Unable to set camera info for calibration. Failure message: %s' %
-                                response.result() if response.result() is not None else "Not available")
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -1,3 +1,7 @@
 *pyc
 .vscode/
 */doc/generated
+
```
ahcorde: do you mind to remove these changes?

ahcorde (changes requested):

.gitignore
ugol-1: Removed

PR #1034, opened 2024-09-20
check_set_camera_info() code clean-up

ugol-1:
Removed the unnecessary `if response.status_message is not None` checks


mikeferguson:
Retriggering CI because it failed on the ROS build farm

mikeferguson (approved):

PR #1039, opened 2024-10-29
Deprecating tf2 C Headers

CursedRock17:
Related to this [pull request](https://github.com/ros2/geometry2/pull/720) in `geometry2` in which we deprecated the `.h` style headers in favor of `.hpp`.


image_rotate/src/image_rotate_node.cpp
```diff
@@ -48,8 +48,8 @@
 #include <vector>
 
 #include "cv_bridge/cv_bridge.hpp"
-#include "tf2/LinearMath/Vector3.h"
-#include "tf2/LinearMath/Quaternion.h"
+#include "tf2/LinearMath/Vector3.hpp"
+#include "tf2/LinearMath/Quaternion.hpp"
```
mikeferguson, line 52: we should really wrap this with an if/else so that the branch continues to compile on older systems (since I expect this change to tf2 will only be released into rolling and later at this point):

```cpp
#if __has_include(<tf2/convert.hpp>)
#include <tf2/convert.hpp>
#else
#include <tf2/convert.h>
#endif

mikeferguson (changes requested):
We also shouldn't merge this until the actual geometry2 PR has been accepted upstream

image_rotate/src/image_rotate_node.cpp
ahcorde: we merged the cpp headers without the deprecation in humble and jazzy, I think it's save to merge this one

CursedRock17:
With the recent backports merged, this PR should be good for review.

mikeferguson:
> With the recent backports merged, this PR should be good for review.

Changes in rolling are released in 0.40.0 - which is staged but not yet synced - CI won't pass until the next sync happens

ahcorde (approved):
rolling testing is green, good to go!

PR #1051, opened 2024-11-28
stereo_image_proc: disparity_node: Add parameter to control camera_info

quic-zhaoyuan:
From https://github.com/ros-perception/image_pipeline/pull/1048 migrate to Rolling. Add info in document rst file
**Issue background:**
Unable to distinguish or remap left and right camera_info when using image_transport with identical topic names

**Description:**
When using DisparityNode from the stereo_image_proc package and running with ROS 2 bag or other scenarios involving image_transport, there is an issue if the camera_info topics for the left and right cameras are automatically derived and end up having the same name.
image
This happens because image_transport::getCameraInfoTopic derives camera_info topic names from the image topics (left/image_rect and right/image_rect). 
This makes it impossible to use the node properly in these scenarios, especially when working with recorded data or custom remapping.

**Proposed Solution**

I propose adding a parameter, use_image_transport_camera_info (default: true), to control whether DisparityNode uses image_transport::getCameraInfoTopic for deriving camera_info topics.

    Default Behavior (backward compatible):
        When use_image_transport_camera_info is true, the node continues using image_transport::getCameraInfoTopic for camera_info resolution, maintaining existing functionality.

    Custom Behavior:
        When use_image_transport_camera_info is false, the node directly uses the camera_info topics specified via remapping (e.g., left/camera_info and right/camera_info), bypassing image_transport's derivation logic.

This solution allows users to explicitly remap the camera_info topics for both cameras, providing flexibility for scenarios where topic names are not unique or need customization.

### DIFF
--- a/stereo_image_proc/doc/components.rst
+++ b/stereo_image_proc/doc/components.rst
@@ -89,6 +89,9 @@ Parameters
    over the network than camera info and/or the delay from disparity processing
    is too long.
 
+*Common*
+ * **use_image_transport_camera_info** (bool, default: true): To control whether DisparityNode uses image_transport::getCameraInfoTopic for deriving camera_info topics. To set false, the node directly uses the camera_info topics specified via remapping (e.g., left/camera_info and right/camera_info), bypassing image_transport's derivation logic.
+
 stereo_image_proc::PointCloudNode
 ---------------------------------
 Combines a rectified color image and disparity image to produce a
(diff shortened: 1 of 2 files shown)


quic-zhaoyuan:
@ahcorde @mikeferguson please check

ahcorde (approved):

mikeferguson (approved):

Issue #1064, opened 2025-02-04
Camera, left_camera and right_camera parameters aren't used


PfeifferMicha:
I can no longer change where the camera_calibrator looks for the set_camera_info parameters, because paths are hard coded.

I might be missing something, but it looks like currently on `rolling`, `jazzy` and `humble` branches, the camera_calibrator ignores the camera, left_camera and right_camera parameters (again).

This **was** already reported in https://github.com/ros-perception/image_pipeline/issues/754 and fixed in:
https://github.com/ros-perception/image_pipeline/pull/755
It had also already been placed on rolling:
https://github.com/ros-perception/image_pipeline/pull/761

But somehow the tip of the all of current branches is missing this change, making it impossible to set the location of the set_camera_info services. They all contain hard-coded paths:
https://github.com/ros-perception/image_pipeline/blob/722ca08b98f37b7b148d429753da133ff1e2c7cf/camera_calibration/src/camera_calibration/camera_calibrator.py#L106

Or am I missing something and there's now a different way to remap the service locations?

mikeferguson:
It looks like https://github.com/ros-perception/image_pipeline/pull/792 reverted the commit you mentioned, in favor of using remap rules rather than parameters (which is more ROS-y). You should be able to use a remap rather than a parameter - for instance, the linked PR has the following example:

Per our [documentation](https://docs.ros.org/en/rolling/p/camera_calibration/doc/tutorial_stereo.html), you should be able to remap the left/right cameras:

```
ros2 run camera_calibration cameracalibrator --approximate 0.1 --size 8x6 --square 0.108 right:=/stereo/right/image_raw left:=/stereo/left/image_raw right_camera:=/stereo/right left_camera:=/my_stereo/left
```

The first remap is for the image_raw/camera_info topics, and the second is for the camera info services. As noted in that tutorial, that command assumes your topic/service names are:

```
/stereo/left/camera_info
/stereo/left/image_raw
/stereo/left/set_camera_info (service)
/stereo/right/camera_info
/stereo/right/image_raw
/stereo/right/set_camera_info (service)
```

PfeifferMicha:
Ah, thanks!

I had already tried that, though. I just get:

```

Waiting for service camera/set_camera_info ...
Service camera/set_camera_info not found.
Waiting for service left_camera/set_camera_info ...
Service left_camera/set_camera_info not found.
Waiting for service right_camera/set_camera_info ...
Service right_camera/set_camera_info not found.
[...]

PR #1065, opened 2025-02-04
improve stereo calibration tutorial

mikeferguson:
* make topic names consistent by removing the `my_` prefix on some topics
* show the output of `ros2 service list` so that service names are more clear
* explain what left/right/left_camera/right_camera actually mean


mikeferguson:
Noticed a few things while answering #1064

ahcorde (approved):

ahcorde:

PR #1066, opened 2025-02-07
Check repeatedly

### LINKED ISSUE #1052: ConsumerThread terminates immediately
When running `cameracheck` from the `camera_calibration` package in ROS2 Jazzy, it subscribes to an image topic only once, performs the check, and then terminates. This behavior might be fine if the calibration board is always visible at the time of launch. However, it is quite inconvenient when the board needs to be moved into view after starting the node.
https://github.com/ros-perception/image_pipeline/blob/bb4ab63f19ec4d7df9cd80a0785ed2e4ffa144d1/camera_calibration/src/camera_calibration/camera_checker.py#L77-L82

Additionally, the results are output via `print()`, which may not be visible when launched through a launch file or similar methods. It would be better to use the ROS logger for consistent and accessible logging.

Tacha-S:
I made it so that it is checked multiple times instead of just once.

- fix #1052

### DIFF
--- a/camera_calibration/src/camera_calibration/camera_checker.py
+++ b/camera_calibration/src/camera_calibration/camera_checker.py
@@ -71,10 +71,10 @@ class ConsumerThread(threading.Thread):
 
     def run(self):
         while rclpy.ok():
-            m = self.queue.get()
             if self.queue.empty():
-                break
-        self.function(m)
+                continue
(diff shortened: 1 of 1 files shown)


ahcorde (approved):

ahcorde:
https://github.com/Mergifyio backport rolling

PR #1071, opened 2025-02-12
image_view：set CvtColorForDisplay encoding as bgr8

zycczy:
based on issue 1021, Fix the issue that the input image encoding as yuv, error to show

### DIFF
--- a/image_view/src/image_view_node.cpp
+++ b/image_view/src/image_view_node.cpp
@@ -214,11 +214,16 @@ void ImageViewNode::imageCb(const sensor_msgs::msg::Image::ConstSharedPtr & msg)
       }
     }
 
-    std::string encoding = "bgr8";
+    std::string encoding = msg->encoding.empty() ? "bgr8" : msg->encoding;
 
     // May want to view raw bayer data
     if (encoding.find("bayer") != std::string::npos) {
       encoding = "mono8";
+    }
+
+    // Add a special rule for YUV format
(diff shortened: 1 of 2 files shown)


zycczy:
Test result as below
Image

image_view/src/image_view_node.cpp
```diff
@@ -214,7 +214,7 @@ void ImageViewNode::imageCb(const sensor_msgs::msg::Image::ConstSharedPtr & msg)
       }
     }
 
-    std::string encoding = msg->encoding.empty() ? "bgr8" : msg->encoding;
+    std::string encoding = "bgr8";
```
mikeferguson: this isn't the right fix - it would likely break mono8 usage.

If a particular encoding doesn't work for display, we should add a special rule for it below (see line 220-222 where we override bayer to mono8 - you can add another rule there for your yuv format that doesn't currently work)

mikeferguson (changes requested):

zycczy:
Hi @mikeferguson 
I add a special rule for the YUV format in the same way as the existing rule for Bayer format. Here is the updated code:

This change ensures that if the input image encoding is YUV, it will be converted to BGR8 for display, while still handling other encodings appropriately.

mikeferguson (approved):

ahcorde (approved):

ahcorde:

PR #1077, opened 2025-02-18
fix depth_image_proc launch files

christian-rauch:
A couple of fixes to make the `depth_image_proc` launch files work again. The issues are mostly caused by API with launch and the RealSense node.

### DIFF
--- a/depth_image_proc/launch/convert_metric.launch.py
+++ b/depth_image_proc/launch/convert_metric.launch.py
@@ -45,7 +45,9 @@ def generate_launch_description():
     return LaunchDescription([
         # install realsense from https://github.com/intel/ros2_intel_realsense
         launch_ros.actions.Node(
-            package='realsense2_camera', executable='realsense2_camera_node',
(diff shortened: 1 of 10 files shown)


mikeferguson (approved):

depth_image_proc/launch/convert_metric.launch.py
```diff
@@ -45,7 +45,7 @@ def generate_launch_description():
     return LaunchDescription([
         # install realsense from https://github.com/intel/ros2_intel_realsense
         launch_ros.actions.Node(
-            package='realsense_ros2_camera', node_executable='realsense_ros2_camera',
+            package='realsense2_camera', executable="realsense2_camera_node",
```
mikeferguson: Looks like this is failing CI because of the double quotes - needs to be single quotes like it was

mikeferguson (changes requested):

mikeferguson (approved):

mikeferguson:

PR #1078, opened 2025-02-18
fix launch (jazzy)

christian-rauch:
Backport https://github.com/ros-perception/image_pipeline/pull/1077.


mikeferguson:
We can actually do an auto back port with mergify once the original PR is merged

PR #1081, opened 2025-02-21
[depth_image_proc] using sensor data qos profile everywhere

LorenzoFerriniCodes:
In `depth_image_proc`, some publishers are already set to use sensor data QoS; however, this is not in place for all of them, nor is for any subscriber.

This PR aims at changing the QoS profile for those publishers and subscribers that are still using the default one, in order to have all of them set to use a sensor data QoS.

### DIFF
--- a/depth_image_proc/src/convert_metric.cpp
+++ b/depth_image_proc/src/convert_metric.cpp
@@ -76,7 +76,7 @@ ConvertMetricNode::ConvertMetricNode(const rclcpp::NodeOptions & options)
   std::lock_guard<std::mutex> lock(connect_mutex_);
   // TODO(ros2) Implement when SubscriberStatusCallback is available
   // pub_depth_ = it_->advertise("image", 1, connect_cb, connect_cb);
-  pub_depth_ = image_transport::create_publisher(this, "image");
+  pub_depth_ = image_transport::create_publisher(this, "image", rmw_qos_profile_sensor_data);
 }
 
 // Handles (un)subscribing when clients (un)subscribe
@@ -92,7 +92,8 @@ void ConvertMetricNode::connectCb()
     sub_raw_ = image_transport::create_subscription(
       this, "image_raw",
       std::bind(&ConvertMetricNode::depthCb, this, std::placeholders::_1),
-      hints.getTransport());
+      hints.getTransport(),
+      rmw_qos_profile_sensor_data);
   }
 }
 
--- a/depth_image_proc/src/crop_foremost.cpp
+++ b/depth_image_proc/src/crop_foremost.cpp
@@ -81,7 +81,7 @@ CropForemostNode::CropForemostNode(const rclcpp::NodeOptions & options)
   std::lock_guard<std::mutex> lock(connect_mutex_);
   // TODO(ros2) Implement when SubscriberStatusCallback is available
   // pub_depth_ = it_->advertise("image", 1, connect_cb, connect_cb);
-  pub_depth_ = image_transport::create_publisher(this, "image");
+  pub_depth_ = image_transport::create_publisher(this, "image", rmw_qos_profile_sensor_data);
 }
 
 // Handles (un)subscribing when clients (un)subscribe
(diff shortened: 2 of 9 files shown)


depth_image_proc/src/convert_metric.cpp
```diff
@@ -76,7 +76,7 @@ ConvertMetricNode::ConvertMetricNode(const rclcpp::NodeOptions & options)
   std::lock_guard<std::mutex> lock(connect_mutex_);
   // TODO(ros2) Implement when SubscriberStatusCallback is available
   // pub_depth_ = it_->advertise("image", 1, connect_cb, connect_cb);
-  pub_depth_ = image_transport::create_publisher(this, "image");
+  pub_depth_ = image_transport::create_publisher(this, "image", rmw_qos_profile_sensor_data);
```
mikeferguson, line 79: This line should be reverted - the strategy throughout image_pipeline is supposed to be:
 * All publishers should be SystemDefaultQOS with allowable overrides
 * All subscribers should be SensorDataQOS with allowable overrides

This follows REP-2003 and deals with the fact that a DefaultQOS publisher will publish best effort when a SensorQOS subscribes to it - but is entirely incompatible with a DefaultQOS subscriber.

depth_image_proc/src/crop_foremost.cpp
```diff
@@ -81,7 +81,7 @@ CropForemostNode::CropForemostNode(const rclcpp::NodeOptions & options)
   std::lock_guard<std::mutex> lock(connect_mutex_);
   // TODO(ros2) Implement when SubscriberStatusCallback is available
   // pub_depth_ = it_->advertise("image", 1, connect_cb, connect_cb);
-  pub_depth_ = image_transport::create_publisher(this, "image");
+  pub_depth_ = image_transport::create_publisher(this, "image", rmw_qos_profile_sensor_data);
```
mikeferguson, line 84: also revert

depth_image_proc/src/register.cpp
```diff
@@ -132,7 +132,9 @@ RegisterNode::RegisterNode(const rclcpp::NodeOptions & options)
   // pub_registered_ = it_depth_reg.advertiseCamera("image_rect", 1,
   //                                               image_connect_cb, image_connect_cb,
   //                                               info_connect_cb, info_connect_cb);
-  pub_registered_ = image_transport::create_camera_publisher(this, "depth_registered/image_rect");
+  pub_registered_ = image_transport::create_camera_publisher(
+    this, "depth_registered/image_rect",
+    rmw_qos_profile_sensor_data);
```
mikeferguson, line 137: revert

depth_image_proc/src/register.cpp
```diff
@@ -192,6 +196,7 @@ void RegisterNode::imageCb(
   registered_msg->width = resolution.width;
   // step and data set in convert(), depend on depth data type
 
+
```
mikeferguson, line 199: we can revert this extra newline

mikeferguson (changes requested):

mikeferguson:
Wait - I just noticed this was targeted at humble (as I was trying to figure out how these updates got missed when I did a big pass over QoS updates). I'm not sure we actually want to change the default behavior in Humble - as that is a stable release, as Humble as been out nearly 3 years, people likely expect it's behavior not to change.

LorenzoFerriniCodes:
> Wait - I just noticed this was targeted at humble (as I was trying to figure out how these updates got missed when I did a big pass over QoS updates). I'm not sure we actually want to change the default behavior in Humble - as that is a stable release, as Humble as been out nearly 3 years, people likely expect it's behavior not to change.

I understand. I also had this doubt while creating the PR, however I also thought that setting the subscribers to use a data sensor qos profile would have been no issue, as it is anyway compatible with the default qos profile that people might have been using for their camera publishers so far in humble

mikeferguson:
> I understand. I also had this doubt while creating the PR, however I also thought that setting the subscribers to use a data sensor qos profile would have been no issue, as it is anyway compatible with the default qos profile that people might have been using for their camera publishers so far in humble

Yeah, I'm gonna leave this open and see if there is any other community feedback

PR #979, opened 2024-05-17
fix: cv2.aruco.interpolateCornersCharuco is deprecated

tfoldi:
There has been API Changes in the newer releases of opencv2 (from 4.8.0). The PR addresses this by supporting both the old and new APIs. 

updated Syntax
```
charucodetector = cv2.aruco.CharucoDetector(board)
charuco_corners, charuco_ids, marker_corners, marker_ids = charucodetector.detectBoard(image)
```
before 4.8.0
```
marker_corners, marker_ids, rejectedImgPoints = cv2.aruco.detectMarkers( image, dictionary)
retval, charuco_corners, charuco_ids = cv2.aruco.interpolateCornersCharuco( marker_corners, marker_ids, image, board)
```

See the changed examples in the main opencv2 repo:

https://github.com/opencv/opencv/blob/f9a59f2592993d3dcc080e495f4f5e02dd8ec7ef/samples/python/calibrate.py#L110

PR #1087, opened 2025-04-16
Replace OpenCV version string comparison with semver.

fgrcar:
When using the ChArUco markers with `cameracalibrator`, I found out that there's an issue with the last versions of OpenCV due to the fact that versions are compared as strings and the thing goes wrong for versions 4.10 and above e.g.:
```
>>> '4.7.0' >= '4.8.0'
False
>>> '4.8.0' >= '4.8.0'
True
>>> '4.9.0' >= '4.8.0'
True
>>> '4.10.0' >= '4.8.0'
False
>>> '4.11.0' >= '4.8.0'
False
```

I fixed the behavior by using `semver` library which is already used in the package e.g.:
```
>>> VersionInfo.parse('4.7.0') >= VersionInfo.parse('4.8.0')
False
>>> VersionInfo.parse('4.8.0') >= VersionInfo.parse('4.8.0')
True
>>> VersionInfo.parse('4.9.0') >= VersionInfo.parse('4.8.0')
True
>>> VersionInfo.parse('4.10.0') >= VersionInfo.parse('4.8.0')
True
>>> VersionInfo.parse('4.11.0') >= VersionInfo.parse('4.8.0')
True
```

Hope this fix helps anyone else too despite the fact that documentation/tutorials are skipping to mention the ChArUcos as a possible pattern (like [here](https://docs.ros.org/en/rolling/p/camera_calibration/doc/index.html)). In my experience, ChArUcos give better results than classic chessboards.

Also a question: does anyone mind if I cherry-pick [this fix here](https://github.com/ros-perception/image_pipeline/pull/979/files) along with mine above also to the humble?

### DIFF
--- a/camera_calibration/src/camera_calibration/calibrator.py
+++ b/camera_calibration/src/camera_calibration/calibrator.py
@@ -39,6 +39,7 @@ import numpy.linalg
 import sensor_msgs.msg
 import tarfile
 from enum import Enum
+from semver import VersionInfo
 
 # Supported camera models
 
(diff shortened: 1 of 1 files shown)


mikeferguson (approved):
Just waiting on CI

ahcorde (approved):

ahcorde:

PR #1099, opened 2025-05-28
Support image_transport lifecycle

ahcorde:
Related with https://github.com/ros-perception/image_common/pull/352


mikeferguson:
I presume we need to branch when we do this?

Yadunund:
I've opened https://github.com/ros-perception/image_pipeline/pull/1114 which is needed to unblock Rolling sync. It incorporates changes from this PR and other fixes to avoid deprecation warnings. @mikeferguson could you take a look?

ahcorde:
testing ci is passing! merging!

PR #1106, opened 2025-07-07
Replace rmq_qos_profile_t with rclcpp::QoS

ahcorde:
Related with https://github.com/ros-perception/image_common/pull/364

### DIFF
--- a/depth_image_proc/src/convert_metric.cpp
+++ b/depth_image_proc/src/convert_metric.cpp
@@ -85,7 +85,7 @@ ConvertMetricNode::ConvertMetricNode(const rclcpp::NodeOptions & options)
         // Get transport hints
         image_transport::TransportHints hints(this);
         // Create subscriber with QoS matched to subscribed topic publisher
-        auto qos_profile = image_proc::getTopicQosProfile(this, topic);
+        rclcpp::QoS qos_profile = image_proc::getQosProfile(this, topic);
         sub_raw_ = image_transport::create_subscription(
           this, topic,
           std::bind(&ConvertMetricNode::depthCb, this, std::placeholders::_1),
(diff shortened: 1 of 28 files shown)


image_proc/include/image_proc/utils.hpp
```diff
@@ -65,6 +65,31 @@ rmw_qos_profile_t getTopicQosProfile(rclcpp::Node * node, const std::string & to
   }
 }
 
+inline
+rclcpp::QoS getQosProfile(rclcpp::Node * node, const std::string & topic)
```
mikeferguson, line 70: should the old getTopicQosProfile be marked deprecated? (or maybe even just removed...?)

ahcorde:
can you take a look @asymingt ?

mikeferguson (approved):
Looks good once CI passes (I presume we need to wait for the underlying changes to sync)

depth_image_proc/src/convert_metric.cpp
```diff
@@ -85,7 +85,7 @@ ConvertMetricNode::ConvertMetricNode(const rclcpp::NodeOptions & options)
         // Get transport hints
         image_transport::TransportHints hints(this);
         // Create subscriber with QoS matched to subscribed topic publisher
-        auto qos_profile = image_proc::getTopicQosProfile(this, topic);
+        rclcpp::QoS qos_profile = image_proc::getQosProfile(this, topic);
```
asymingt, line 88: nit: There seems to be an inconsistency in this PR between whether we want to go `auto qos_profile = ...` or `rclcpp::QoS qos_profile = ...`. It doesn't faze me which one we choose, but I think we should at least be internally consistent.

asymingt (approved):
LGTM, just one nit.

ahcorde:
testing is passing!

PR #1125, opened 2026-01-14
Cleanup bsd 3 clause license usage

ahcorde:
Related with https://github.com/ament/ament_lint/pull/209

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,3 +1,3 @@
 Any contribution that you make to this repository will
-be under the BSD license 3.0, as dictated by that
+be under the BSD license, as dictated by that
 [license](https://opensource.org/licenses/BSD-3-Clause).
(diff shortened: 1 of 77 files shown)


LICENSE
```diff
@@ -1,34 +1,31 @@
-Portions of the code are licensed under the BSD License 2.0:
+Portions of the code are licensed under the BSD License 3.0:
```
tfoote: There's several variations of the BSD license. It's the 3 clause bsd license,  not version 3.  This should probably just be "BSD License"
```suggestion
Portions of the code are licensed under the BSD License:
```

CONTRIBUTING.md
```diff
@@ -1,3 +1,3 @@
 Any contribution that you make to this repository will
-be under the BSD license 2.0, as dictated by that
+be under the BSD license 3.0, as dictated by that
```
tfoote: Remove the whole 3.0 it's not a version. 
```suggestion
be under the BSD license, as dictated by that
```

tfoote (changes requested):
Two small tweaks on the verbiage but otherwise lgtm. It's too bad we don't have a little bit more flexibiltiy on unwrapping the content. But this is a good step forward.

tfoote (approved):

ahcorde:
https://github.com/Mergifyio backport kilted jazzy humble

PR #1128, opened 2026-03-16
fix(camera_calibration): fisheye stereo calibration fixes

diegomarzaa:
cv2.fisheye.stereoCalibrate is unstable in OpenCV 4.5.x with big FOV cameras and rolling shutter.

### DIFF
--- a/camera_calibration/src/camera_calibration/calibrator.py
+++ b/camera_calibration/src/camera_calibration/calibrator.py
@@ -1177,23 +1177,33 @@ class StereoCalibrator(Calibrator):
             if LooseVersion(cv2.__version__).version[0] == 2:
                 print("ERROR: You need OpenCV >3 to use fisheye camera model")
                 sys.exit()
-            else:
-                # WARNING: cv2.fisheye.stereoCalibrate wants float64 points
-                lipts64 = numpy.asarray(lipts, dtype=numpy.float64)
-                lipts = lipts64
-                ripts64 = numpy.asarray(ripts, dtype=numpy.float64)
-                ripts = ripts64
-                opts64 = numpy.asarray(opts, dtype=numpy.float64)
-                opts = opts64
 
-                cv2.fisheye.stereoCalibrate(opts, lipts, ripts,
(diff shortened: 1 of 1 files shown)


fallenmi (changes requested):
The new pinhole `stereoRectify()` path breaks the documented nonzero `alpha` range for valid wide-FOV stereo rigs. `StereoCalibrator.set_alpha()` says that `alpha` ranges from 0 (all calibrated pixels valid) to 1 (all original pixels visible), but this path can return a negative projection focal length and then passes that projection directly to `fisheye.initUndistortRectifyMap()`.

On the target Ubuntu 22.04 / OpenCV 4.5.4 stack, I reproduced this on exact head `615de852bed07b9b4046527b06e1d895d7bd2e4c` with a 1280x720 fisheye pair, 340 px focal length, and a 30-degree baseline yaw. At `alpha=0.5`, the submitted `cv2.stereoRectify(..., D=0)` call returns `P[0,0] = -84.011635`, and the resulting left rectification map has zero valid source samples. Running the previous `cv2.fisheye.stereoRectify(..., balance=0.5)` path with the same intrinsics, distortion, and recovered R/T returns a positive focal length of `116.252142` and a map with 37.637% valid samples. The exact result reproduced 10/10. A broader deterministic matrix found 60 submitted-path configurations with negative projection focal lengths.

This is isolated to rectification at nonzero alpha: a 432-case calibration matrix passes on both base and head, and the head's default `alpha=0` rectification checks pass. Please keep the fisheye-specific projection handling (or derive and validate a positive fisheye new-camera matrix before building the maps), and add a wide-FOV, nonparallel stereo regression that calls `set_alpha(0.5)` and asserts positive projections plus nonempty remap coverage.

AI disclosure: I used OpenAI Codex to inspect the exact calibration/rectification paths, construct and run the exact base/head/current-merge numerical oracles on OpenCV 4.5.4, and draft this review. I verified the inputs, projection matrices, remap coverage, exact commits, public policy/interactions, and conclusion.

Issue #1133, opened 2026-04-22
depth_image_proc: heap-buffer-overflow write in PointCloudXyzNode / convertDepth<unsigned short>() with oversized depth Image dimensions


BreakHell-zhiyuan:
**Describe the bug**
depth_image_proc::PointCloudXyzNode crashes when it receives a depth Image + CameraInfo pair carrying oversized width and height.
PointCloudXyzNode::depthCb(...) copies attacker-controlled image dimensions into the output PointCloud2, prepares point cloud storage, and then calls convertDepth<unsigned short>(). convertDepth() iterates over cloud_msg->height * cloud_msg->width and writes XYZ values through PointCloud2Iterator<float> objects.
With oversized forged dimensions, the allocated PointCloud2 backing buffer becomes inconsistent with the number of writes later performed. As a result, convertDepth<unsigned short>() writes past the end of the heap buffer and triggers a heap-buffer-overflow. ASan reports a WRITE of size 4 in convertDepth<unsigned short>(), and the allocation trace shows that the backing region allocated from PointCloud2Modifier::setPointCloud2FieldsByString(...) is only 1048576 bytes.

**Environment**
Ubuntu 22.04
ROS 2 Humble
package: depth_image_proc

**Steps to reproduce**
1.Start a component container
2.Load the node
3.Publish a malformed depth Image and matching CameraInfo with oversized dimensions, In my reproduction, I used a Python publisher that sends:
- CameraInfo.width = 65537
- CameraInfo.height = 65536
- Image.width = 65537
- Image.height = 65536
- Image.encoding = "16UC1"
- Image.step = 65537 * 2
- Image.data = 5MB of non-zero uint16 values

**Observed behavior**
the component container exits and ASan reports:
- heap-buffer-overflow
- WRITE of size 4
- crash in depth_image_proc::convertDepth<unsigned short>()

**Root cause**

- This is a buffer allocation / usage size mismatch.
- depthCb(...) trusts unvalidated width / height from the incoming depth image.
- These dimensions are copied into cloud_msg.
- PointCloud2Modifier::setPointCloud2FieldsByString(...) prepares the PointCloud2 backing buffer.
- convertDepth() then uses cloud_msg->height and cloud_msg->width as loop bounds and writes XYZ values through iter_x, iter_y, and iter_z.
[...]

PR #1136, opened 2026-04-22
heap-buffer-overflow write in PointCloudXyzNode / convertDepth<unsigned short>() with oversized depth Image dimensions

ahcorde:
Related with https://github.com/ros-perception/image_pipeline/issues/1133

### DIFF
--- a/depth_image_proc/src/point_cloud_xyz.cpp
+++ b/depth_image_proc/src/point_cloud_xyz.cpp
@@ -102,6 +102,46 @@ void PointCloudXyzNode::depthCb(
   const Image::ConstSharedPtr & depth_msg,
   const CameraInfo::ConstSharedPtr & info_msg)
 {
+  // Reject unreasonable or inconsistent dimensions before allocating the
+  // output cloud. Without this guard, oversized width/height can overflow
+  // internal size computations and produce a PointCloud2 buffer that is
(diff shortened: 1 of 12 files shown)


depth_image_proc/src/point_cloud_xyz.cpp
```diff
@@ -102,6 +102,46 @@ void PointCloudXyzNode::depthCb(
   const Image::ConstSharedPtr & depth_msg,
   const CameraInfo::ConstSharedPtr & info_msg)
 {
+  // Reject unreasonable or inconsistent dimensions before allocating the
+  // output cloud. Without this guard, oversized width/height can overflow
+  // internal size computations and produce a PointCloud2 buffer that is
+  // smaller than the iterator loop in convertDepth() will write.
+  // 100 MP is well above any realistic sensor; tweak if needed.
+  constexpr uint64_t kMaxPixels = 100ULL * 1000ULL * 1000ULL;
```
mikeferguson: does it make sense to just make this a parameter? I imagine that there is probably somebody out there who is using some stupid size point cloud, who is going to exceed this and rather than needing another PR, could just change the parameter

mikeferguson (approved):

ahcorde:
https://github.com/Mergifyio backport kilted jazzy

PR #1170, opened 2026-06-01
Use lambdas, replace NULL with nullptr and std::format

ahcorde:
- Use lambdas instead of std::bind
 - replace NULL with nullptr
 - use std::format


asymingt (commented):
In general LGTM. The only issue I see is that you have changed the expected parameter format from`"frame%04i.jpg"` to `"frame{:04}.jpg"`. anybody using the existing node might be surprised by this change. I would consider is a breaking change that would warrant a major version bump. Is there a way we can signal this to users?

ahcorde:
Related to `frame{:04}.jpg`. This is used by `std::format` to replace the name. the `{}` are no included in the filename.

PR #1177, opened 2026-06-02
Added flake8 and pep257 to camera calibration

ahcorde:


asymingt (approved):
Looks like a straight reformatting change plus linter tests. LGTM.

